PR #4, opened 2017-04-10
Minor updates to ExprTk

ArashPartow:
- Updated multi-sub expression operator to return final sub-expression type.
- Updates to exprtk_disable_return_statement macro for disabling return statements and associated exceptions at the source code level.


sjunges:
Hi,

thanks for updating exprtk!
We will pull exprtk shortly, as soon as we have some spare resources. 

-- Sebastian

PR #34, opened 2018-11-05
Fix broken model building when reordering is active

kleinj:
This PR fixes an issue with model building, when `--dynreordering` is active.

Consider the following PRISM model (designed to induce variable reordering).
```
dtmc

module m
   s0 : [0..10] init 0;
   s1 : [0..10] init 0;
   s2 : [0..10] init 0;
   s3 : [0..10] init 0;
   s4 : [0..20] init 0;
   tick : [0..1] init 0;

   [] tick=0 & s0<10 -> 1:(tick'=1)&(s0'=s0+1);
   [] tick=0 & s1<10 -> 1:(tick'=1)&(s1'=s1+1);
   [] tick=0 & s2<10 -> 1:(tick'=1)&(s2'=s2+1);
   [] tick=0 & s3<10 -> 1:(tick'=1)&(s3'=s3+1);
   [] tick=1 -> 1:(s4'=mod(s0+s1+s2+s3, 20))&(tick'=0);
endmodule
```

When running with reordering
```
$ bin/storm --prism test-reorder.prism --prop 'P=?[ F s4=6 ]' --engine dd --ddlib cudd --dynreorder
...
Model type:     DTMC (symbolic)
States:         466 (196 nodes)
Transitions:    797 (848 nodes)
Reward Models:  none
Variables:      rows: 6 meta variables (22 DD variables), columns: 6 meta variables (22 DD variables)
Labels:         2
   * deadlock -> 47 state(s) (41 nodes)
   * init -> 48 state(s) (30 nodes)

Model checking property "1": P=? [F (s4 = 6)] ...
Result (for initial states): [0, 1] (range)
Time for model checking: 0.005s.
```
notice the additional, unexpected initial states, leading to a result range of `[0,1]` instead of the normal result `1` (that's also the result without reordering).

This is due to [getBddEncodingLessOrEqualThanRec() in InternalCuddDdManager.cpp](https://github.com/moves-rwth/storm/blob/6fcebed10eafa1c15a3b12f16949d164b36a3d56/src/storm/storage/dd/cudd/InternalCuddDdManager.cpp#L64), which is not compatible with reordering. This leads to the `getRange()` calls for the variables to return bogus results.

This PR contains
 * infrastructure for inhibiting dynamic reordering in certain scopes
 * a variant of `getBddEncodingLessOrEqualThan` that is able to handle reordered DDs

I did some testing and looked at some DOTs for the generated DDs, but having a second pair of eyes on the changes would be good. Especially as there are no test cases for reordering, right?

Design decision: The code now always does some preprocessing to recover the "DD variable index/level to bit in the encoding" before the recursion, even though that is straightforward in the case where no reordering has occurred. From running the tests, it didn't look like it introduces significant overhead, but it could be the case that it's worthwhile to have a special case for that?

Output with PR applied:

```
$ bin/storm --prism test-reorder.prism --prop 'P=?[ F s4=6 ]' --engine dd --ddlib cudd --dynreorder
...
Model type:     DTMC (symbolic)
States:         419 (176 nodes)
Transitions:    797 (848 nodes)
Reward Models:  none
Variables:      rows: 6 meta variables (22 DD variables), columns: 6 meta variables (22 DD variables)
Labels:         2
   * deadlock -> 0 state(s) (1 nodes)
   * init -> 1 state(s) (23 nodes)

Model checking property "1": P=? [F (s4 = 6)] ...
Result (for initial states): 1
Time for model checking: 0.004s.
```

### DIFF
--- a/src/storm/storage/dd/DdManager.cpp
+++ b/src/storm/storage/dd/DdManager.cpp
@@ -431,7 +431,12 @@ namespace storm {
         void DdManager<LibraryType>::triggerReordering() {
             internalDdManager.triggerReordering();
         }
-        
+
+        template<DdType LibraryType>
+        std::unique_ptr<typename DdManager<LibraryType>::DynamicReorderingInhibitor> DdManager<LibraryType>::getDynamicReorderingInhibitor() {
+             return internalDdManager.getDynamicReorderingInhibitor();
+        }
+      
         template<DdType LibraryType>
         std::set<storm::expressions::Variable> DdManager<LibraryType>::getAllMetaVariables() const {
             std::set<storm::expressions::Variable> result;
--- a/src/storm/storage/dd/DdManager.h
+++ b/src/storm/storage/dd/DdManager.h
@@ -30,7 +30,9 @@ namespace storm {
 
             template<DdType LibraryTypePrime, typename ValueType>
             friend class AddIterator;
-            
+
+            typedef typename InternalDdManager<LibraryType>::DynamicReorderingInhibitor DynamicReorderingInhibitor;
+
             /*!
              * Creates an empty manager without any meta variables.
              */
@@ -271,6 +273,12 @@ namespace storm {
             void triggerReordering();
             
             /*!
+             * Requests an object that inhibits dynamic reordering as long as it is
+             * in scope.
+             */
+            std::unique_ptr<DynamicReorderingInhibitor> getDynamicReorderingInhibitor();
+            
+            /*!
              * Retrieves the meta variable with the given name if it exists.
              *
              * @param variable The expression variable associated with the meta variable.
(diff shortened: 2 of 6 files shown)


kleinj:
PR is broken, the logic for handling the case `thenNode=elseNode` is too restrictive. Will rework and post a new version using `Cudd_bddIte` to properly handle all the corner cases later.

sjunges:
Hi Joachim,

have you ever fixed this issue? I saw some recent paper from Dresden that mentioned this topic. 

Otherwise I will close this pull-request.

kleinj:
Hi Sebastian,

this is a good question. I'll have to check back to my discussions with Chris/Clemens and see what the status was. I'll get back to you.

sjunges:
Closed due to inactivity and probably Coe being outdated now. Added an issue to keep track of the problem.

PR #100, opened 2020-12-17
Implement manually triggered github actions CI

dbasgoeze:
It can be schedueled later after testing that the workflow works with the secrets, etc.

I can also change the name of course.


volkm:
Looks good!

PR #110, opened 2021-04-02
Make RelevantEvents independent of std::vector

dbasgoeze:
Instead use a flexible iterator based api that is independent of a container.

Also make RelevantEvents more easily mutable to add new events.

### DIFF
--- a/src/storm-dft/utility/RelevantEvents.h
+++ b/src/storm-dft/utility/RelevantEvents.h
@@ -152,6 +152,22 @@ namespace storm {
                 }
             }
 
+            /*!
+             * Merge the given RelevantEvents with *this
+             *
+             * @return A reference to *this, allowing chaining i.e. e.merge(a).merge(b)
+             */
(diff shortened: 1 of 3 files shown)


src/storm-dft/utility/RelevantEvents.h
```diff
@@ -15,58 +16,76 @@ namespace storm {
         public:
 
             /*!
-             * Create relevant events from given event names.
+             * Constructs empty RelevantEvents object
+             */
+            RelevantEvents() = default;
+
+            /*!
+             * Create relevant events from given event names in an initilizer list.
```
volkm, line 24:
```suggestion
             * Create relevant events from given event names in an initializer list.
```

volkm (approved):
Looks good! Thanks for improving the RelevantEvents.

src/storm-dft/utility/RelevantEvents.h
dbasgoeze: Fixed

PR #111, opened 2021-04-06
Derivative merge

glatteis:
This pull request merges the `derivative` module in `storm-pars` into `master` (now "rebased"). And trying out the GitHub merge! I cleaned this up and I think it's ready to merge now.

## Features (C++):
- A `DerivativeEvaluationHelper` class which can quickly compute derivatives of reachability probabilities or expected rewards of pMCs at given instantiations
- A `GradientDescentInstantiationSearcher` class which can find feasible instantiations and optima of pMCs 

## Features (Command Line)
- Flag `--compute-derivative` to compute the derivative at an instantiation using the `DerivativeEvaluationHelper`
- Flags `--find-extremum` and `--find-feasible` for doing exactly that using the `GradientDescentInstantiationSearcher`
- A number of options for the `GradientDescentInstantiationSearcher`


.gitignore
```diff
@@ -52,3 +52,4 @@ resources/3rdparty/cudd-3.0.0/Makefile.in
 resources/3rdparty/cudd-3.0.0/aclocal.m4
 # Travis helpers
 travis/mtime_cache/cache.json
+compile_commands.json
```
tquatmann, line 55:
```suggestion
```

glatteis:
The `DerivativeEvaluationHelperTest` seems to fail on travis, but works on my computer. I will look into why that is!

glatteis:
I know the test fails because the reference polynomial yields the wrong result, not the `DerivativeEvaluationHelper. I think this is because `carl::VariablePool::getInstance().findVariableWithName("p")` yields a different `p`than the `p` the dtmc on travis. It doesn't happen on my computer though, so I have to re-run this test again :-(

tquatmann:
> I know the test fails because the reference polynomial yields the wrong result, not the `DerivativeEvaluationHelper`. I think this is because `carl::VariablePool::getInstance().findVariableWithName("p")` yields a different `p`than the `p` the dtmc on travis. It doesn't happen on my computer though, so I have to re-run this test again :-(

Could it be that your tests should clear off the VariablePool? Like [this](https://github.com/moves-rwth/storm/blob/master/src/test/storm-pars/modelchecker/SparseDtmcParameterLiftingTest.cpp#L59).

glatteis:
> > I know the test fails because the reference polynomial yields the wrong result, not the `DerivativeEvaluationHelper`. I think this is because `carl::VariablePool::getInstance().findVariableWithName("p")` yields a different `p`than the `p` the dtmc on travis. It doesn't happen on my computer though, so I have to re-run this test again :-(
> 
> Could it be that your tests should clear off the VariablePool? Like [this](https://github.com/moves-rwth/storm/blob/master/src/test/storm-pars/modelchecker/SparseDtmcParameterLiftingTest.cpp#L59).

Yeah, that's probably it. I will do that, thanks! But I think with the workaround that is there now it should also work.

tquatmann:
In any case, I think cleaning the variable pool is a good idea to ensure that tests that are executed afterwards can start with a clean environment. Also note that the `TearDown()` thing I linked above is also called even when there is e.g. an uncatched exception in your test code.

glatteis:
> In any case, I think cleaning the variable pool is a good idea to ensure that tests that are executed afterwards can start with a clean environment. Also note that the `TearDown()` thing I linked above is also called even when there is e.g. an uncatched exception in your test code.

That makes sense! I'm also making it a cool `TYPED_TEST` with typed environments like the test you linked while I'm at it.

glatteis:
Let's hope this works! I also removed the persisting of the `Environment` inside of my classes, as I saw it's probably good practice in storm to pass it for every call to a method. A caveat here is that I definitely need the `Environment` inside of my constructors, so I need to trust that the caller uses the same environments in the constructor and the method calls.

glatteis:
It failed now because of the docker rate limit.

glatteis:
Will close and re-open the PR to restart the CI.

glatteis:
I guess the compilation fails with gcc now? It works with clang.

glatteis:
Failed because of the rate limit again - can something be done about that?

glatteis:
Again, the docker rate limit. I give up for now :D

sjunges:
Generally looks good. Some thoughts (happy to discuss)

The 'derivative helper' is not really a helper, it is like an instantiation checker? This code may also be useful in other combinations rather than just in the search. Likewise, finding feasible points is not necessarily restricted to gradient search.

Can we split the settings to reflect
-  that we have three parts here: a general part were contrary to region checking or solution function computation, you add feasibility and optimal feasibility as the task.
- have the feasibility options to select a method (just gradient descent for now) 
- have settings to fix gradient descent parameters as you do
- have settings for the analysis of the gradients. 

Regarding the derivative helper: 
- why do we need the formula if we already have a single target file? Also, if our properties must be a single property, why not just pass that?
- Do we need both mode and reward name? Or can we just say reward name and if the name is none, then it is a probability?

More comments may follow. Feel free to disagree with me :-)

glatteis:
Hi, thanks for the feedback! I will be looking at it in detail in two or three weeks, once I am back at the i2 :)

glatteis:
Another good change would be, I think, moving the optimization direction out of the formula, as the `Pmax`, `Pmin`, ... actually have a meaning for MDPs.

glatteis:
My new contract starts in 3 days, I can talk with Jip then and I will probably start finishing this :)

glatteis:
> Can we split the settings to reflect
> 
>     * that we have three parts here: a general part were contrary to region checking or solution function computation, you add feasibility and optimal feasibility as the task.
> 
>     * have the feasibility options to select a method (just gradient descent for now)
> 
>     * have settings to fix gradient descent parameters as you do
> 
>     * have settings for the analysis of the gradients.
> 

Can we brainstorm how the commands would look then?
This would still have feasible/optimum (I would call it one-shot because it is that) on the same level as checking an instatiation:

```--find-feasible --search-method gd``` and ```--one-shot --search-method gd``` and ```--check-instantiation ...```

So we can break this out and apply some "make invalid states unrepesentable" as I want to move the bounds out of the formulas:

```--find-instantiation --below 2.5``` / ```--find-instantiation --above 0.1``` /  ```--find-instantiation --one-shot max``` and ```---check-instantiation ...```

or

```--find-below 2.5``` / ```--find-above 0.1``` /  ```--one-shot max``` / ```--one-shot min``` and ```---check-instantiation ...```

or

```--find-instantiation below 2.5``` / ```--find-instantiation one-shot max``` and ```---check-instantiation ...```

(the last is my favorite)

> Regarding the derivative helper:
> 
>     * why do we need the formula if we already have a single target file? Also, if our properties must be a single property, why not just pass that?

I think I did this because multiple other places in Storm do the same. Maybe I should just get rid of the properties in general and use the target label (I need to change this anyway if I want to get rid of Pmax/Pmin and Pmax>=0.5 ...) But doing this would fix the API so that only derivatives of reachability can be checked. I think the best solution here is to expect a quantitative `CheckTask` for the `DerivativeInstantiationModelChecker` and maybe a `Formula` or also a `CheckTask` for the `GradientDescentInstantiationSearcher` and add explicit switches for max/min and bounds. 

>     * Do we need both mode and reward name? Or can we just say reward name and if the name is none, then it is a probability?

Yes, but I like this even less than the current situation. Getting rid of the enum makes invalid states unrepresentable, but also makes reward checking pretty implicit. Maybe there's a good middle ground?

glatteis:
Another TODO thing: Add an option to permanently fix some of the DTMC's parameters. Make it possible to not specify parameters that are optimized out in the model checker's CLI.

sjunges:
Regarding the API: (Making this independent comments to allow threads).

- Your variants are one option, we could also just allow for a "Task" struct which you can initialize accordingly (including max or min, reward and if so, which, etc). Then you pass this task object. This is in line with what we do for the model checkers and I think it is rather clean as we can even add a isValid and isSupported type of infrastructure.

glatteis:
> Regarding the API: (Making this independent comments to allow threads).
> 
>     * Your variants are one option, we could also just allow for a "Task" struct which you can initialize accordingly (including max or min, reward and if so, which, etc). Then you pass this task object. This is in line with what we do for the model checkers and I think it is rather clean as we can even add a isValid and isSupported type of infrastructure.

That's a good idea.

glatteis:
> Regarding the API: (Making this independent comments to allow threads).
> 
>     * Your variants are one option, we could also just allow for a "Task" struct which you can initialize accordingly (including max or min, reward and if so, which, etc). Then you pass this task object. This is in line with what we do for the model checkers and I think it is rather clean as we can even add a isValid and isSupported type of infrastructure.

I suppose I will make a `DerivativeCheckTask` that extends `CheckTask` with the additional info of the parameter and optionally the original model checking results.

glatteis:
Refactor started in a different branch: https://github.com/glatteis/storm/tree/derivative-refactor

glatteis:
I created the `DerivativeCheckTask`: https://github.com/glatteis/storm/blob/derivative-refactor/src/storm-pars/derivative/DerivativeCheckTask.h

I (almost successfully) changed `SparseDerivativeInstantiationModelChecker` to use it. Now the methods are analogous to the other model checkers.

sjunges:
Hi Linus, 

Regarding the command line, can we maybe meet next week to discuss this 'in person' via zoom? 

Sebastian

glatteis:
Oof, I accidentally merged some master changes into the last commit.

glatteis:
Now it looks good. Will see if it works in this state and I haven't made any mistakes

glatteis:
Looks good :)

glatteis:
I still seem to have a bug with reward models and CheckTasks :-(

sjunges:
@jipspel @glatteis What is the status of this pull request?

glatteis:
I think it can be squashed and merged.

PR #112, opened 2021-04-12
Github Actions: use current ref instead of hardcoded master

dbasgoeze:
Fixes pull requests building the master branch

### DIFF
--- a/.github/workflows/buildtest.yml
+++ b/.github/workflows/buildtest.yml
@@ -14,7 +14,7 @@ env:
   CARL_BRANCH: "master14"
   CARL_GIT_URL: "https://github.com/smtrat/carl.git"
   STORM_GIT_URL: "${{ github.server_url }}/${{ github.repository }}.git"
-  STORM_BRANCH: "master"
+  STORM_BRANCH: "${{ github.head_ref || github.sha }}"
   # github runners currently have two cores
   NR_JOBS: "2"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/buildtest.yml
```diff
@@ -119,7 +122,10 @@ jobs:
         # Build & TEST & DEPLOY STORM
         #####
       - name: Git clone
-        run: sudo docker exec storm git clone --branch $STORM_BRANCH $STORM_GIT_URL /opt/storm
+        run: |
+        # git clone cannot clone individual commits based on a sha and some other refs
+        # this workaround fixes this and fetches only one commit
+          sudo docker exec storm bash -c "mkdir /opt/storm; cd /opt/storm; git init && git remote add origin ${STORM_GIT_URL} && git fetch --depth 1 origin ${STORM_BRANCH} && git checkout FETCH_HEAD"
```
volkm, line 128: We should clone the complete repo here.
The versioning uses git tags which are missing if only one commit is pulled. While this still works in Storm without problems, stormpy will encounter issues with determining the right version. As we are also deploying this part to Docker, it should contain the complete repo.

volkm (changes requested):

.github/workflows/buildtest.yml
dbasgoeze: If have fixed it so that only when we deploy will the whole history be cloned.

.github/workflows/buildtest.yml
```diff
@@ -118,8 +121,17 @@ jobs:
...
         # Build & TEST & DEPLOY STORM
         #####
-      - name: Git clone
-        run: sudo docker exec storm git clone --branch $STORM_BRANCH $STORM_GIT_URL /opt/storm
+      - name: Git shallow clone
+        # Only clone shallow if not using master on original repo (and not for pull requests or forks)
+        if: ${{ !(github.repository_owner == 'moves-rwth' && github.ref == 'refs/heads/master') }}
+        run: |
+          # git clone cannot clone individual commits based on a sha and some other refs
+          sudo docker exec storm bash -c "mkdir /opt/storm; cd /opt/storm; git init && git remote add origin ${STORM_GIT_URL} && git fetch origin ${STORM_BRANCH} && git checkout FETCH_HEAD"
```
volkm, line 129: Is ``--depth 1`` missing here?

dbasgoeze: yes, I must have removed it before splitting Git clone into shallow and deep clone

volkm (approved):

PR #114, opened 2021-04-19
Support for Prism Programs with unbounded integers

tquatmann:
This PR adds support for PRISM programs that use unbounded integer variables declared via `x : int`.
See also the [PRISM documentation](https://www.prismmodelchecker.org/manual/ThePRISMLanguage/ModulesAndVariables) and the [hill-toggle](https://qcomp.org/benchmarks/#hill-toggle) and [p53](https://qcomp.org/benchmarks/#p53) benchmarks at QComp.

The explicit model builder (sparse engine) supports building such models assuming that they are finite. The `--int-bits <NUMBER>` setting can be used to specify how many bits shall be reserved for these variables.

The symbolic model builder will error for these models.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+
 Changelog
 ==============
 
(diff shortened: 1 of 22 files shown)


sjunges:
Looks good to me.

I am not 100% what this does to the exploration engine. Could that be a problem?

tquatmann:
Good point! I just checked: the exploration engine relies on the NextStateGenerator-interface which works as before. Of course, this might or might not terminate for infinite models.

PR #116, opened 2021-05-05
Remove effectless instantiation after specialization in ConstantsComparator.cpp

dbasgoeze:
Currently, we instantiate the ConstantComparator with ClnRationalNumber and GmpRationalNumber if we have them. Curiously one of them can be equal to storm::RationalNumber for which we have an explicit specialization.

This pull request proposes to just explicitly instantiate  the ConstantComparator for types that are not storm::RationalNumber.

I suspect that there is a bug hidden in there currently as if one has STORM_HAVE_CLN and also not STORM_USE_CLN_EA then there is an instantiation of ConstantsComparator<ClnRationalNumber> which does not use the explicit specialization as one would expect for a rational.

### DIFF
--- a/src/storm/utility/ConstantsComparator.cpp
+++ b/src/storm/utility/ConstantsComparator.cpp
@@ -14,17 +14,17 @@ namespace storm {
         bool ConstantsComparator<ValueType>::isOne(ValueType const& value) const {
             return storm::utility::isOne(value);
         }
-        
+
         template<typename ValueType>
         bool ConstantsComparator<ValueType>::isZero(ValueType const& value) const {
             return storm::utility::isZero(value);
         }
-        
+
         template<typename ValueType>
         bool ConstantsComparator<ValueType>::isEqual(ValueType const& value1, ValueType const& value2) const {
             return value1 == value2;
         }
-        
+
         template<typename ValueType>
         bool ConstantsComparator<ValueType>::isConstant(ValueType const& value) const {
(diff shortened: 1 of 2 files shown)


src/storm/utility/ConstantsComparator.cpp
```diff
@@ -159,11 +159,11 @@ namespace storm {
         template class ConstantsComparator<storm::storage::sparse::state_type>;
 
 #ifdef STORM_HAVE_CARL
-#if defined(STORM_HAVE_CLN)
+#if defined(STORM_HAVE_CLN) && !defined(STORM_USE_CLN_EA)
```
tquatmann, line 162: I'm slightly confused. Why do we instantiate the comparator for CLN only if we are _not_ using CLN (for exact arithmethic)?

dbasgoeze: I don't know, but it is used in other compilation units i.e. gives linkage errors if not instantiated. As I said I suspect a bug there, as this instantiation would use the template and not an explicit specialisation.

dbasgoeze: This commit should not change the current behavior but I do think the current behavior is buggy...

tquatmann:
I made a new [PR](https://github.com/moves-rwth/storm/pull/117) that  addresses the issue for this PR but also other issues with the ConstantsComparator.

PR #117, opened 2021-05-10
Revised ConstantsComparator

tquatmann:
**ConstantsComparator**: The old code used `abs(x-y)/abs(x+y)<= epsilon` when comparing two numbers for equality (modulo relative precision). This is weird when x and y have different signs and potentially even causes a div by 0 whenever `x=-y`. Moreover the templating was awkward, causing a lot of code duplications and a Warning with clang.

### DIFF
--- a/src/storm/builder/ExplicitModelBuilder.h
+++ b/src/storm/builder/ExplicitModelBuilder.h
@@ -32,9 +32,6 @@
 #include "storm/generator/VariableInformation.h"
 
 namespace storm {
-    namespace utility {
-        template<typename ValueType> class ConstantsComparator;
-    }
     
     namespace builder {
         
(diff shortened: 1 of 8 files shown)


sjunges:
I completely agree with the constantcomperator rewrite. LGTM

The forward declarations to constant comparator do allow to avoid dependence of header files on RationalNumberAdapter.h. 
I am open to a discussion, but I think that reducing header-in-header inclusions is a good thing, as we should aim to reduce storms compile times....

Happy to hear your thoughts.

tquatmann:
Yes, I also hesitated a bit about removing the forward declarations. Unfortunately---with the new code---there is no way to do a forward declaration like this

```C++
template<typename ValueType> class ConstantsComparator;
```

essentially because the dummy template parameter `Enable` is missing. We can also not add `Enable` to the forward declaration because then the forward-declaring class does not know what to put there.

`Enable` is a dummy parameter to allow SFINAE. I think with `constexpr if` there potentially are ways to write the `ConstantsComparator` without ugly code duplications and without an extra template parameter, but that is something for the future (i.e. c++17).

sjunges:
Okay, yes, lets not wait for going to c++17 before merging :-).

PR #123, opened 2021-06-09
Fixed incorrect type of modulo expression with integer inputs

johannesalehmann:
This pull request fixes the type of modulo expressions if both parameters are integers. Previously, the output of `operator%(...)` was always a rational expression, while it should be an integer expression in this case. This caused an error when reading prism files using the modulo operator with integer variables.

I also changed modulo(...), which already returned the correct type, but confusingly used the `minimumMaximum(...)` function for this.


tquatmann:
Looks very good! Thanks for also adding test cases.

PR #135, opened 2021-07-26
Fix FormulaParser Debug 

PrangerStefan:
With the following

``` cpp
// Enable the following lines to print debug output for most the rules.
debug(start);
debug(constantDefinition);
debug(stateFormula);
...
```
we are able to print debug outputs for formula parsing. This lacked the two `ostream` operators added with this PR.

### DIFF
--- a/src/storm-parsers/parser/ConstantDataType.cpp
+++ b/src/storm-parsers/parser/ConstantDataType.cpp
@@ -1,0 +1,14 @@
+#include "ConstantDataType.h"
+
+namespace storm {
+    namespace parser {
+        std::ostream& operator<<(std::ostream& out, ConstantDataType const& constantDataType) {
+          switch(constantDataType) {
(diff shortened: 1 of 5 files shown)


PrangerStefan:
I just realised that this does not cover all the rules defined in the grammar, will extend this asap (`untilFormula` is missing for example).

tquatmann:
Thanks for providing those. 
There has been a major rewrite of the formula parser, see [here](https://github.com/moves-rwth/storm/pull/137).
I would prefer to have those changes merged in first and go back to this PR afterward.

PrangerStefan:
rebased on to master (I don't know why the force-push does not show up here...)

sjunges:
@tquatmann are you taking a look at this one?

tquatmann:
Thanks for providing those! Looks good :)

PR #141, opened 2021-08-24
Added model export in .json.

tquatmann:
New option `--exportbuild` exports the state space of the model in various formats: `.ddrd`, `.drn`, `.dot`, and a new `.json` format.
The json export includes transition probabilities, labels, rewards, exit rates, as well as (if included during model building) state valuations, choice labels, and choice origins.
For example, `storm -qvbs stream 0 ""  --exportjson test.json --buildstateval --buildchoiceorig --buildfull --build-all-labels --exact` yields

```json
[
    {
        "c": [
            {
                "id": 0,
                "origin": {
                    "action-label": "buffer",
                    "transitions": [
                        {
                            "action": "buffer",
                            "automaton": "streamingclient",
                            "destinations": [
                                {
                                    "assignments": [
                                        {
                                            "comment": "s <- 1",
                                            "ref": "s",
                                            "value": 1
                                        }
                                    ],
                                    "location": "l"
                                }
                            ],
                            "guard": {
                                "comment": "(((s = 0) & (n < 10)) & (k = n))",
                                "exp": {
                                    "left": {
                                        "left": {
                                            "left": "s",
                                            "op": "=",
                                            "right": 0
                                        },
                                        "op": "∧",
                                        "right": {
                                            "left": "n",
                                            "op": "<",
                                            "right": 10
                                        }
                                    },
                                    "op": "∧",
                                    "right": {
                                        "left": "k",
                                        "op": "=",
                                        "right": "n"
                                    }
                                }
                            },
                            "location": "l"
                        }
                    ]
                },
                "succ": [
                    {
                        "id": 1,
                        "prob": "1"
                    }
                ]
            }
        ],
        "id": 0,
        "lab": [
            "init"
        ],
        "s": {
            "_loc_streamingclient": 0,
            "buffering": 0,
            "done": false,
            "k": 0,
            "n": 0,
            "numrestarts": 0,
            "running": false,
            "s": 0,
            "underrun": false
        }
    },
    {
        "c": [
            {
                "id": 1,
                "origin": {
```

### DIFF
--- a/src/storm/models/sparse/Model.cpp
+++ b/src/storm/models/sparse/Model.cpp
@@ -473,7 +473,10 @@ namespace storm {
                     storm::json<JsonValueType> stateRewardsJson;
                     for (auto const& rm : rewardModels) {
                         if (rm.second.hasStateRewards()) {
-                            stateRewardsJson[rm.first] = storm::utility::to_string(rm.second.getStateReward(state));
+                            auto r = rm.second.getStateReward(state);
+                            if (!storm::utility::isZero(r)) {
+                                stateRewardsJson[rm.first] = storm::utility::to_string(r);
+                            }
                         }
                     }
                     if (!stateRewardsJson.empty()) {
@@ -502,18 +505,21 @@ namespace storm {
                         if (hasChoiceLabeling()) {
                             auto choiceLabels = getChoiceLabeling().getLabelsOfChoice(choiceIndex);
                             if (!choiceLabels.empty()) {
-                                choiceJson["labels"] = choiceLabels;
+                                choiceJson["lab"] = choiceLabels;
                             }
                         }
-                        choiceJson["index"] = choiceIndex;
+                        choiceJson["id"] = choiceIndex;
                         storm::json<JsonValueType> choiceRewardsJson;
                         for (auto const& rm : rewardModels) {
                             if (rm.second.hasStateActionRewards()) {
-                                choiceRewardsJson[rm.first] = storm::utility::to_string(rm.second.getStateActionReward(choiceIndex));
+                                auto r = rm.second.getStateActionReward(choiceIndex);
+                                if (!storm::utility::isZero(r)) {
+                                    choiceRewardsJson[rm.first] = storm::utility::to_string(r);
+                                }
                             }
                         }
                         if (!choiceRewardsJson.empty()) {
-                            choiceRewardsJson["rewards"] = std::move(choiceRewardsJson);
+                            choiceRewardsJson["rew"] = std::move(choiceRewardsJson);
                         }
                         storm::json<JsonValueType> successors;
                         for (auto const& entry : transitionMatrix.getRow(choiceIndex)) {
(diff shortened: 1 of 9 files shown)


sjunges:
I think that exportjson is not a good name for the option; after all JANI among others is also in JSON. Can we do something more specific?

Regarding the export itself, what are s and c? are that states and commands? can we write them out?

I think we should also add a warning if people use this with a large system because it will result in a very large file.

src/storm/models/sparse/Model.cpp
```diff
@@ -453,7 +455,98 @@ namespace storm {
...
+                storm::json<JsonValueType> output;
+                for (uint64_t state = 0; state < getNumberOfStates(); ++state) {
+                    storm::json<JsonValueType> stateChoicesJson;
+                    stateChoicesJson["id"] = state;
+                    if (hasStateValuations()) {
+                        stateChoicesJson["s"] = getStateValuations().template toJson<JsonValueType>(state);
+                    }
+                    auto labels = getLabelsOfState(state);
+                    if (!labels.empty()) {
+                        stateChoicesJson["lab"] = labels;
```
sjunges, line 471: I would not make lab optional, but rather put in an empty array if there are no labels.

sjunges:
Final remark, we are not that strong on checking whether options together make sense, but it may make sense to add in the finalise part a check that this is not combined with dd or sth like that? I don't think we need to overdo it, but it may reduce a number of emails :-)

src/storm/models/sparse/Model.cpp
tquatmann: Agreed.

tquatmann:
> I think that exportjson is not a good name for the option; after all JANI among others is also in JSON. Can we do something more specific?

I was trying to be in line with the already existing `--exportdot` option but I totally see your point. I'd suggest some option that can deal with multiple file formats (e.g. mcsta has something like this as well). Maybe:

`--exportbuild <file> [<format>]`: Writes the built model to the given file in the provided format,  where `<format>` shall be any of `[auto, dot, drdd, drn, json]` .
[`auto` is the default and does an automatic detection based on the  file extension]

> Regarding the export itself, what are s and c? are that states and commands? can we write them out?

`s` is state and `c` is choice. I tried to have this at least a bit compatible with the .json export of schedulers. I'm not sure if we should change that.

> I think we should also add a warning if people use this with a large system because it will result in a very large file.

Yes, that is a good idea. Will do.

tquatmann:
> Final remark, we are not that strong on checking whether options together make sense, but it may make sense to add in the finalise part a check that this is not combined with dd or sth like that? I don't think we need to overdo it, but it may reduce a number of emails :-)

Yes, such a check makes sense. I'd rather have this in the CLI  code because it is hard to keep track of such compatibility issues across settings modules. [This](https://github.com/moves-rwth/storm/blob/155757b8d7c543ea45b19ec0438ce1bca7790789/src/storm-cli-utilities/model-handling.h#L575) is probably a good place to check.

Also, we could consider moving this kind of functionality to a separate binary (either `storm-conv` or a new one) to clean up the interface. But I think this is a bit out of scope for this PR.

sjunges:
I like `--exportbuild <file> [<format>]`. The warning you added is good.  I agree that checking combinations of settings should be part of the cli code, and I would suggest we create an issue for that for future reference.  Thanks for fixing the misleading function name :-)

tquatmann:
With the new `ModelExportFormat` enum, it was quite natural to check the compatibility between model representation and export format within the CLI since it is now just the default case of a switch statement.

sjunges (approved):

PR #143, opened 2021-09-29
Labels from program in simulator

sjunges:
Nothing wild, I added code such that one can obtain the current labels from the program into the simulator (no special labels). 
For that, I extended the prism generator such that one can evaluate boolean expressions for the currently loaded state. 

P.S. I don't know why the previous pull request code by Stefan reappears...


tquatmann (approved):
I'm a bit worried that the commits from the other PR appear here. Did you maybe do a rebase instead of a merge? 
Apart from that, this is looking good.

PR #144, opened 2021-09-29
remove cudd config.h.in and ignore that file

sjunges:
File is auto-generated during make (optimally, that file should be made in a different location, but it is auto tools and I am not gonna mess with that).


sjunges:
Before merging this, we first have to resolve the other pull request as I forgot to base this on the correct branch :/

PR #145, opened 2021-09-29
Export AIGs from BDDs via Storm

sjunges:
Code generates an AIG for the transition graph of an MDP, extracted from the DD representation.


.github/workflows/buildtest.yml
```diff
@@ -5,7 +5,7 @@ name: Build Test
 on:
   schedule:
       # run daily
-      - cron: '0 6 * * *'
+      #- cron: '0 6 * * *'
```
sjunges, line 8: This change has to be undone.

PR #146, opened 2021-09-30
remove cudd config.h.in that is regenerated during make

sjunges:


tquatmann (approved):
Can be merged once the build test is ready.

PR #150, opened 2021-10-20
Bug in Exploration-Engine

steffi4321:
We found a bug in the exploration engine that led to a Segmentation-Fault while checking the added example model. We tracked it down to be an incorrect assignment in the method `collapseMec`. The provided change should fix this issue.


tquatmann:
> I added the test. It runs through with the fix, and not without the fix. However, the exploration engine is based on random decisions, so the problem may not occur in every run.

Great! Thank you very much!

PR #159, opened 2021-10-29
Fixes

tquatmann:
This PR fixes two issues I noticed when running the QComp 2020 benchmarks.
It also fixes a couple of warnings that I haven't noticed before.

### DIFF
--- a/resources/examples/testfiles/ma/ftwc.jani
+++ b/resources/examples/testfiles/ma/ftwc.jani
@@ -1,0 +1,1534 @@
+﻿{
+	"jani-version": 1,
+	"name": "ftwc",
+	"type": "ma",
+	"features": [
+		"derived-operators",
+		"arrays"
+	],
+	"actions": [
+		{
+			"name": "startRepairBackbone?"
+		},
+		{
+			"name": "finishRepairBackbone?"
+		},
+		{
+			"name": "startRepairSwitch?"
+		},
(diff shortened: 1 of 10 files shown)


sjunges:
What is the reason to change the assertion for a developer bug to an exception for a user? I don't mind much, but would like to understand the reasoning such that I conform to the same reasoning :-)

tquatmann:
> What is the reason to change the assertion for a developer bug to an exception for a user? I don't mind much, but would like to understand the reasoning such that I conform to the same reasoning :-)

There is a switch statement that covers all items of an enum and there is a return statement in each case. GCC isn't sufficiently happy about that and prints a  `control reaches end of non-void function` warning. An `assert` doesn't silence that warning (at least not in release build) but throwing an exception does.  I agree that this would be a developer error and not a user error, though.

sjunges:
I see, thanks for the clarification.

ArashPartow:
The main reason for the stack limitation in ExprTk:

A user of the library had setup a site where users could provide expressions to be evaluated. They found that a user had provided a malicious expression that would cause a stackoverflow during parsing and possible root escalation - as there was a string component in the expression that had x86-64 instructions, luckily ASLR saved the day.

tquatmann:
Wow, thanks for providing those insights! That sounds like a perfectly valid reason.

PR #160, opened 2021-11-02
Revert "Update Eigen to 3.4.0"

tquatmann:
Reverts moves-rwth/storm#153

There seem to be some issues with gcc11 and the `GMRES` method in Eigen 3.4.0. The old Eigen version worked fine in this regard but had issues with `-DSTORM_USE_CLN_EA=On`.

However, since the issues with the new version affect the *default* CMAKE configuration, I think it's best to go back to the old version for now.

I'll add an issue addressing this for future reference.


sjunges:
LGTM, thanks of taking care of this

PR #161, opened 2021-11-02
Fix init state enumeration

tquatmann:
Revises enumeration of initial states in Prism and Jani NextStateGenerators.

FYI: The brute-force enumeration of all initial states is triggered whenever every (valid) variable valuation is initial, i.e., if no (non-transient) variable has an initial value and all initial restrictions are `true`.

Apparently, this enumeration was quite broken.

This PR fixes issues with models that contain

- boolean variables, 
- integer variables, where the lower bound is not zero, and
- initial locations other than 0.

There was also an issue with converting `CompressedState`s with negative integer assignments to a string.

### DIFF
--- a/src/storm/generator/JaniNextStateGenerator.cpp
+++ b/src/storm/generator/JaniNextStateGenerator.cpp
@@ -267,61 +267,57 @@ namespace storm {
                 
                 STORM_LOG_DEBUG("Enumerated " << initialStateIndices.size() << " initial states using SMT solving.");
             } else {
+                
+                // Create vectors holding all possible values
+                std::vector<std::vector<uint64_t>> allValues;
+                for (auto const& aRef : this->parallelAutomata) {
+                    auto const& aInitLocs = aRef.get().getInitialLocationIndices();
+                    allValues.template emplace_back(aInitLocs.begin(), aInitLocs.end());
+                }
+                uint64_t locEndIndex = allValues.size();
+                for (auto const& intVar : this->variableInformation.integerVariables) {
+                    STORM_LOG_ASSERT(intVar.lowerBound <= intVar.upperBound, "Expecting variable with non-empty set of possible values.");
+                    // The value of integer variables is shifted so that 0 is always the smallest possible value
+                    allValues.push_back(storm::utility::vector::buildVectorForRange<uint64_t>(static_cast<uint64_t>(0), intVar.upperBound + 1 - intVar.lowerBound));
+                }
+                uint64_t intEndIndex = allValues.size();
+                // For boolean variables we consider the values 0 and 1.
+                allValues.resize(allValues.size() + this->variableInformation.booleanVariables.size(), std::vector<uint64_t>({static_cast<uint64_t>(0), static_cast<uint64_t>(1)}));
+                
+                std::vector<std::vector<uint64_t>::const_iterator> its;
+                std::vector<std::vector<uint64_t>::const_iterator> ites;
+                for (auto const& valVec : allValues) {
+                    its.push_back(valVec.cbegin());
+                    ites.push_back(valVec.cend());
+                }
+                
+                // Now create an initial state for each combination of values
                 CompressedState initialState(this->variableInformation.getTotalBitOffset(true));
-                
-                std::vector<int_fast64_t> currentIntegerValues;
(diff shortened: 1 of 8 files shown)


src/storm/generator/JaniNextStateGenerator.cpp
```diff
@@ -267,61 +267,57 @@ namespace storm {
...
+                std::vector<std::vector<uint64_t>> allValues;
+                for (auto const& aRef : this->parallelAutomata) {
+                    auto const& aInitLocs = aRef.get().getInitialLocationIndices();
+                    allValues.template emplace_back(aInitLocs.begin(), aInitLocs.end());
+                }
+                uint64_t locEndIndex = allValues.size();
+                for (auto const& intVar : this->variableInformation.integerVariables) {
+                    STORM_LOG_ASSERT(intVar.lowerBound <= intVar.upperBound, "Expecting variable with non-empty set of possible values.");
+                    // The value of integer variables is shifted so that 0 is always the smallest possible value
+                    allValues.push_back(storm::utility::vector::buildVectorForRange<uint64_t>(static_cast<uint64_t>(0), intVar.upperBound + 1 - intVar.lowerBound));
```
AlexBork, line 281: Could it make sense to only store the bounds and generate the range vector when it is needed? Storing the vectors directly may lead to large memory requirements.

src/storm/generator/PrismNextStateGenerator.cpp
```diff
@@ -159,58 +160,47 @@ namespace storm {
...
-                    STORM_LOG_THROW(variable.lowerBound <= variable.upperBound, storm::exceptions::InvalidArgumentException, "Expecting variable with non-empty set of possible values.");
-                    currentIntegerValues.emplace_back(0);
-                    initialState.setFromInt(variable.bitOffset, variable.bitWidth, 0);
+                
+                // Create vectors holding all possible values
+                std::vector<std::vector<uint64_t>> allValues;
+                for (auto const& intVar : this->variableInformation.integerVariables) {
+                    STORM_LOG_ASSERT(intVar.lowerBound <= intVar.upperBound, "Expecting variable with non-empty set of possible values.");
+                    // The value of integer variables is shifted so that 0 is always the smallest possible value
+                    allValues.push_back(storm::utility::vector::buildVectorForRange<uint64_t>(static_cast<uint64_t>(0), intVar.upperBound + 1 - intVar.lowerBound));
```
AlexBork, line 169: Similar to the JANI case: Could it make sense to only store the bounds? in particular, if I see it correctly here all lower range bounds are 0, so it would suffice to store the upper bound.

AlexBork (approved):
Overall, I think it's okay to merge. The code is cleanly written. My comments concern a possible optimisation so I will wait for your opinion before merging.

src/storm/generator/JaniNextStateGenerator.cpp
tquatmann: Yes, this would make sense. However, for locations, we do not have a consecutive list of ints and `storm::utility::combinatorics::forEach` requires that all iterators have the same type. 
There is probably some way around this (maybe using [boost range](https://www.boost.org/doc/libs/1_76_0/libs/range/doc/html/index.html) ) but to me, this seems to be not worth the effort. In particular, note that we'll have at least `allValues[0].size() * ... * allValues[n-1].size()` many states anyway which will most likely be the bottleneck.

src/storm/generator/PrismNextStateGenerator.cpp
tquatmann: Here I agree that something like [counting_range](https://www.boost.org/doc/libs/1_76_0/libs/range/doc/html/range/reference/ranges/counting_range.html) would be better. However, I still think this is not very concerning because actually exploring those states is much much more costly

src/storm/generator/JaniNextStateGenerator.cpp
AlexBork: It might still cut the peak memory requirement by almost half as the vectors are kept in memory. Anyways, I understand that in practice this is unlikely to be a problem and probably not worth the effort.

src/storm/generator/PrismNextStateGenerator.cpp
AlexBork: Yep, I agree.

PR #165, opened 2021-11-09
fixed #164

### LINKED ISSUE #164: Storm-POMDP: MakePOMDPCanonic has a broken error message
Dear all,

the following error message is not helpful because somehow the output is broken. 
```

ERROR (MakePOMDPCanonic.cpp:199): Observation 1 sometimes provides multiple actions, but in state 6 provides only one action PG???~?`P??P??8$+	 %+	? +	?.
libc++abi: terminating with uncaught exception of type storm::exceptions::AmbiguousModelException: AmbiguousModelException: Observation 1 sometimes provides multiple actions, but in state 6 provides only one action PG

```

In particular, there is no action PG and the action PG is always different. The error message would also profit from showing more info about the state.

To reproduce 
```
storm-pomdp --prism maze-bug.prism.txt -const CMAX=2,THRESHOLD=1.0  -ec --prop "Rmax=? [F goal]"
```
[maze-bug.prism.txt](https://github.com/moves-rwth/storm/files/7505764/maze-bug.prism.txt)

sjunges:
Fixes a broken error message

### DIFF
--- a/src/storm-pomdp/transformer/MakePOMDPCanonic.cpp
+++ b/src/storm-pomdp/transformer/MakePOMDPCanonic.cpp
@@ -194,7 +194,12 @@ namespace storm {
                         }
                         std::string actionval= "";
                         if (pomdp.hasChoiceLabeling()) {
-                            actionval = *pomdp.getChoiceLabeling().getLabelsOfChoice(rowIndexFrom).begin();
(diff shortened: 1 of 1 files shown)


AlexBork (approved):
This definitely improves the error message in case no label is provided. I'll merge it.

PR #170, opened 2021-11-16
Computation of the expected number of times each state in a DTMC/CTMC is visited (sparse engine)

tquatmann:
This PR adds the functionality to compute the expected times each state of a DTMC or CTMC is visited. 
In other words, the resulting value at state `s` coincides with the total reward `R{"rew_s"}=? [C]`, where the reward structure `rew_s`  only assigns reward 1 to state `s`. If `s` does not lie on a cycle, this value actually also coincides with the probability to reach `s`.

The underlying idea is to solve the infamous forward equation system (`x * P + init = x`) for each non-bottom SCC with probability matrix `P`.

States in (reachable) BSCCs always have value `inf`.

Other related stuff in here:

- Improved the `--exportresult` output to also include state labels.
- Changed the JSON output for weird floats such as `inf` or `nan`. Before, the output just said `null` in those cases but I think a string literal like `"inf"` is better for our purposes.
- Replaced an old implementation of the expected visiting times with the new one. The old implementation was used to compute the reachability of each BSCC when computing the (very related) steady state distributions via `--steadystate`.

Here is an example output

```console
storm-debug --prism resources/examples/testfiles/dtmc/die.pm --expvisittimes --exportresult res.json --buildstateval --build-all-labels
Storm 1.6.4 (dev)
DEBUG BUILD

Date: Tue Nov 16 12:52:27 2021
Command line arguments: --prism resources/examples/testfiles/dtmc/die.pm --expvisittimes --exportresult res.json --buildstateval --build-all-labels
Current working directory: /Users/tim/storm

Time for model input parsing: 0.047s.

Time for model construction: 0.130s.

Model type: 	DTMC (sparse)
States: 	13
Transitions: 	20
Reward Models:  none
State Labels: 	6 labels
   * deadlock -> 0 item(s)
   * two -> 1 item(s)
   * done -> 6 item(s)
   * three -> 1 item(s)
   * init -> 1 item(s)
   * one -> 1 item(s)
Choice Labels: 	none
Write to file res.json.
Result: [0.3333333333, inf] (range)
Time for model checking: 0.007s.
?expvisittimes ~/storm> head -n 70 res.json                                                                                                                    
[
    {
        "l": [
            "init"
        ],
        "s": {
            "d": 0,
            "s": 0
        },
        "v": 1
    },
    {
        "l": [],
        "s": {
            "d": 0,
            "s": 1
        },
        "v": 0.666666666666667
    },
    {
        "l": [],
        "s": {
            "d": 0,
            "s": 2
        },
        "v": 0.666666666666667
    },
    {
        "l": [],
        "s": {
            "d": 0,
            "s": 3
        },
        "v": 0.333333333333333
    },
    {
        "l": [],
        "s": {
            "d": 0,
            "s": 4
        },
        "v": 0.333333333333333
    },
    {
        "l": [],
        "s": {
            "d": 0,
            "s": 5
        },
        "v": 0.333333333333333
    },
    {
        "l": [],
        "s": {
            "d": 0,
            "s": 6
        },
        "v": 0.333333333333333
    },
    {
        "l": [
            "done",
            "one"
        ],
        "s": {
            "d": 1,
            "s": 7
        },
        "v": "inf"
    },
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Version 1.6.x
 - Added `--exportbuilt` option that exports the built model in various formats. Deprecates `--io:exportexplicit`, `--io:exportdd` and `--io:exportdot`
 - Added export of built model in .json. which can be used to debug and explore the model.
 - Added computation of steady state probabilities for DTMC/CTMC in the sparse engine. Use `--steadystate` in the command line interface.
+- Added computation of the expected number of times each state in a DTMC/CTMC is visited (sparse engine). Use `--expvisittimes` in the command line interface.
 - Implemented parsing and model building of Stochastic multiplayer games (SMGs) in the PRISM language. No model checking implemented (yet).
 - API: Simulation of prism-models 
 - API: Model-builder takes a callback function to prevent extension of particular actions, prism-to-explicit mapping can be exported
--- a/resources/3rdparty/modernjson/src/json.hpp
+++ b/resources/3rdparty/modernjson/src/json.hpp
@@ -1571,7 +1571,7 @@ Format](http://rfc7159.net/rfc7159)
                 m_type = value_t::string;
                 m_value = string_t("inf");
             }
-            else if (storm::utility::isInfinity(-val))
+            else if (storm::utility::isInfinity<number_float_t>(-val))
             {
                 m_type = value_t::string;
                 m_value = string_t("-inf");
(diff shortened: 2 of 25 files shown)


AlexBork (approved):
Overall a well-written PR. The functionality is a reasonable extension for Storm. There are some minor comments, mostly some formulations for messages I would change for clarification.

PR #171, opened 2021-11-18
Location Elimination

johannesalehmann:
This PR implements "location elimination", a technique that can be applied to Jani models to reduce the size of the resulting DTMCs. The technique is described in more detail [in this paper](https://arxiv.org/pdf/2011.00983.pdf).

To enable the functionality, I added the command line option `--location-elimination` to both storm and storm-conv. This option takes two optional parameters to control the elimination process.

Most of the functionality is in a series of actions:

- `UnfoldAction`: This takes a variable and "unfolds" it into the location space. For example, if a model has locations `l1` and `l2` and we unfold a variable `x` with domain {1, 2, 3}, the new model will have locations `l1_x_1`, `l1_x_2`, …, `l_2_x_3`.
- `EliminateAction`: This removes a location from the model. By adding new edges, we ensure that reachability probabilities are preserved.
- `EliminateAutomaticallyAction`: This uses a heuristic to determine which locations can be removed and repeatedly calls `EliminateAction`.
- `RebuildWithoutUnreachableAction`: This removes unreachable locations and unsatisfiable edges from the model.
- `AutomaticAction`: This alternates between unfolding and eliminating. It uses a heuristic to decide which variable to unfold next.

The `JaniLocalEliminator` contains code that is shared between the actions. When location elimination is enabled using the `--location-elimination` flag, a single `AutomaticAction` is executed, so the user does not have to provide guidance for the elimination process.


src/storm/storage/jani/localeliminator/JaniLocalEliminator.cpp
```diff
@@ -0,0 +1,542 @@
...
+                    }
+                }
+
+                for (auto &var : newModel.getGlobalVariables()){
+                    // Only make integer variables non-transient, as those are the only variables added during the
+                    // process and all others were present since the beginning.
+                    if (var.getType().isBoundedType() && var.getType().asBoundedType().isIntegerType()){
+                        var.setTransient(false);
+                    }
+                }
```
tquatmann, line 112: What if there were transient integer variables in the original model?

tquatmann (changes requested):
I haven't looked into all details (especially those in `storage/jani/localeliminator`) but the code is looking quite good and mostly has a clean structure. It is also well tested and does not interfere much with the existing code base which is much appreciated. Good Job! 👍 

General comments:
- I think it would be helpful to add a comment to each of the main classes so that it is more clear what they are doing (maybe just use the descriptions you gave for this PR). Also, putting a link to the research paper is probably helpful (e.g. in `JaniLocalEliminator.h`).
- Some functions are a bit long, which sometimes makes it a bit hard to understand what is going on. (There is no need to rewrite everything now, just keep that in mind)

src/storm/storage/jani/localeliminator/AutomaticAction.h
```diff
@@ -8,6 +8,11 @@
 
 using namespace boost;
```
tquatmann, line 9:
```suggestion
```

src/storm/storage/jani/localeliminator/JaniLocalEliminator.cpp
johannesalehmann: I've removed the useTransientVariables option (it's now always true), since it wasn't used anyway, so there is no longer any danger of preexisting transient variables being converted into non-transient variables.

src/storm/storage/jani/Variable.h
```diff
@@ -63,6 +63,7 @@ namespace storm {
              * Sets the initial expression for this variable.
              */
             void setInitExpression(storm::expressions::Expression const& initialExpression);
+            void setTransient(bool transient);
```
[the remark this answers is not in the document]
johannesalehmann: It turns out that the only usage of this was in the !useTransientVariables branch, which I have removed, so I could also remove setTransient(...) without any major changes in other parts of the code.

PR #172, opened 2021-11-23
Jani flatten function fix

tquatmann:
This PR fixes flattening of jani models that use functions in guard expressions.

Also improves catching and outputting errors related to flattening.


AlexBork (approved):
The changes are clear and make sense. The additional error/warning messages are comprehensible. Thus, I suggest that we merge.

PR #173, opened 2021-12-04
Fix DFTModularizer

dbasgoeze:
The commits in this PR also makes small housekeeping changes.


volkm (approved):
LGTM. Good idea to improve some of the functions regarding the top event.

PR #176, opened 2021-12-16
an access method for the underlying model

sjunges:


src/storm-pars/modelchecker/instantiation/SparseInstantiationModelChecker.cpp
```diff
@@ -31,6 +31,11 @@ namespace storm {
         bool SparseInstantiationModelChecker<SparseModelType, ConstantType>::getInstantiationsAreGraphPreserving() const {
             return instantiationsAreGraphPreserving;
         }
+
+        template <typename SparseModelType, typename ConstantType>
+        bool SparseInstantiationModelChecker<SparseModelType, ConstantType>::getOriginalModel() const {
```
volkm, line 36: Should return a SparseModelType

volkm (changes requested):
One fix required for the return type. Otherwise it looks good to me and can be merged after the fix.

src/storm-pars/modelchecker/instantiation/SparseInstantiationModelChecker.cpp
sjunges: I should not commit old versions. Thanks for spotting

volkm (approved):

PR #179, opened 2021-12-23
Fix initial bound for pla

jipspel:
Initial bound for pla was not working correctly, I think I fixed this before but somewhere in merge it got lost. 
I now fixed it with boost optional, another option would be to either use a boolean to check if it got initialized, or correctly initialize it at 0/1 for probabilities and 0 and infty for rewards.


tquatmann (approved):
Looks good to me! Thanks for the fix.

Issue #80, opened 2020-07-15
Boost: Old version of optional


tquatmann:
The [CMakeLists.txt](https://github.com/moves-rwth/storm/blob/1343ae5c19e731573568dd6a3cfcbd06b722f9b5/CMakeLists.txt#L236) still contains this:
```
# TODO: remove forcing the old version of optional as soon as the related Spirit bug is fixed:
# https://svn.boost.org/trac/boost/ticket/12349
# Fix thanks to: https://github.com/freeorion/freeorion/issues/777
add_definitions(-DBOOST_RESULT_OF_USE_DECLTYPE -DBOOST_OPTIONAL_CONFIG_USE_OLD_IMPLEMENTATION_OF_OPTIONAL)
```

I have looked this up and it seems that the issue is fixed since boost 1.67.0 (However, I haven't tested this, yet).
If we drop the fix, we no longer support Debian 9 (Boost 1.62.0) and Ubuntu 18.04 (Boost 1.65.1) which might still be in use somewhere.

I would suggest detecting the boost version and enable/disable the fix automatically.

PR #180, opened 2021-12-23
Updated ubuntu base system and debian version in buildtest

tquatmann:
Previously, ubuntu 20-10 was used which reached end of life.

We now also run our tests for debian 11, but I decided to no longer test for debian 9.

Note that the latter ships with GCC 6 which has quite limited c++17 support, so I think that we should eventually drop support for debian 9, anyway.

### DIFF
--- a/.github/workflows/buildtest.yml
+++ b/.github/workflows/buildtest.yml
@@ -61,7 +61,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        distro: ["ubuntu-18.04", "debian-11", "debian-10", "ubuntu-20.04"]
+        distro: ["debian-10", "debian-11", "ubuntu-18.04", "ubuntu-20.04"]
         debugOrRelease: ["release"]
     steps:
       - name: Setup cmake arguments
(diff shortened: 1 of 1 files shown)


.github/workflows/buildtest.yml
```diff
@@ -61,7 +61,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        distro: ["ubuntu-18.04", "debian-10", "debian-9", "ubuntu-20.04"]
+        distro: ["ubuntu-18.04", "debian-11", "debian-10", "ubuntu-20.04"]
```
sjunges, line 64: I found the ordering somewhat confusing ;-)

sjunges:
LGTM (minor comment, no need to do something with that)

.github/workflows/buildtest.yml
tquatmann: Yes, I fixed it :)

dbasgoeze:
Can we now remove forcing the old version of optional? Iirc these versions where the reasons why we did not do it already.

tquatmann:
> Can we now remove forcing the old version of optional? Iirc these versions where the reasons why we did not do it already.

As far as I can see, this also affects ubuntu 18-04. I'm fine with ending support for that as well, but let's roll out a new release first before we actively break things.

I have also created issue #80 for this.

PR #182, opened 2021-12-28
Stormpy update

sjunges:
Towards supporting an update of stormpy, these base classes are missing virtual destructors even though this is strongly recommended.


sjunges:
havent properly updated the code.

PR #188, opened 2022-02-04
Add Lexicograhpic Model Checking

steffi4321:
This is the implementation of the journal version of "Stochastic games with lexicographic reachability-safety objective". It contains the necessary files to solve a lexicographic LTL-objective on MDPs.


tquatmann:
Thanks for the contribution! I'll look into this at some point.

Please run [this](https://github.com/steffi4321/storm/actions/workflows/formatapply.yml) workflow to fix the formatting.

Also, it seems that this does not compile without spot. Please use the `STORM_HAVE_SPOT` macro like e.g. [here](https://github.com/moves-rwth/storm/blob/c136f66c1ffbef064966d2a9dd6b2b105d467077/src/storm/automata/LTL2DeterministicAutomaton.cpp#L24)

tquatmann:
Thanks for the nice contribution!

PR #189, opened 2022-02-04
Fix BitVector::operator= OOB

Marckvdv:
Currently, assigning a large BitVector to a small one causes the assignment operator to write out of bounds. This PR fixes that by making sure the `bitCount` is updated *after* we check if the two BitVectors have the same size. Note that `bucketCount()` is computed using `bitCount`. In other words, assuming buckets is not null, the condition of the if statement always false, which means that the size of the current `buckets` is never changed. This leads to OOB in the `std::copy_n` that follows. I added a simple test to check that it indeed works. The test that I added crashes without the fix.

### DIFF
--- a/src/storm/storage/BitVector.cpp
+++ b/src/storm/storage/BitVector.cpp
@@ -113,11 +113,11 @@ BitVector::BitVector(BitVector const& other) : bitCount(other.bitCount), buckets
 BitVector& BitVector::operator=(BitVector const& other) {
     // Only perform the assignment if the source and target are not identical.
     if (this != &other) {
-        bitCount = other.bitCount;
         if (buckets && bucketCount() != other.bucketCount()) {
             delete[] buckets;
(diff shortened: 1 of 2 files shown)


sjunges:
LGTM. I agree with the fix, but since we are looking at this code; Is it the best that even if the new bitvector is shorter, we delete and re-alloc everything? Can we dealloc part of an array?

Marckvdv:
Right, I suppose this would be a nice use case of `realloc` in C, which can (sometimes) avoid unnecessary allocation/deallocation.  The c++ equivalent would be to use `std::vector` instead and call `resize`, which might not be a bad idea. This would also save you the headache of manually managing the bitvector's memory. 

But this would require rewriting the bitvector class so it would be easier to stick to the current code. We could also only reallocate if the new bitvector is larger, which would be more performant at the cost of using some more memory.

PR #196, opened 2022-02-24
Issue 155 fix

### LINKED ISSUE #155: Change sparsepp to parallel hashmap. And use it.
Sparsepp triggers a warning:
```
/Users/tim/storm/build-debug/include/resources/3rdparty/sparsepp/sparsepp/spp.h:942:33: warning: definition of implicit copy constructor for 'Two_d_destructive_iterator<std::__1::pair<const DdNode *const, bool>, spp::sparsegroup<std::__1::pair<const DdNode *const, bool>, spp::libc_allocator<std::__1::pair<const DdNode *const, bool>>> *, std::__1::pair<const DdNode *const, bool> *, std::__1::input_iterator_tag, spp::libc_allocator<std::__1::pair<const DdNode *const, bool>>>' is deprecated because it has a user-declared copy assignment operator [-Wdeprecated-copy]
```

As suggested in the [sparsepp readme](https://github.com/greg7mdp/sparsepp) we should  move on to [parallel hashmap](https://github.com/greg7mdp/parallel-hashmap) as the overall performance appears to be better.

We should also consider using this more often, instead of the (often slower) STL maps.

SvStein:
Replaced sparsepp with phmap. Closes #155 .


tquatmann:
LGTM

Issue #194, opened 2022-02-23
Adding AddressSanitizer as a build configuration option


Marckvdv:
# Adding runtime/static analysis as a build configuration option
This issue proposes to add several build configuration options related to using tooling that can help find common bugs. Initially, these options would just be available for whoever wants to use them. Next, when they are deemed to be stable enough, they could be added as a requirement for passing the tests performed by the CI. At that point, perhaps these options should be made as the default for debug builds.

We could add individual options for each tool, for example `-DSTORM_COMPILE_WITH_X`, as well as an option that combines all checkers `-DSTORM_COMPILE_WITH_ALL_CHECKERS`, which would be used for CI.

For now, I will only look into adding a `-DSTORM_COMPILE_WITH_ADDRESS_SANITIZER` flag, as I have good experiences with AddressSanitizer:
* Really simple to set up. It is part of GCC and Clang for close to 10 years so any somewhat recent compiler should support it. Enabling it in debug mode is simply a matter of compiling with the `-fsanitize=address` flag.
* No false positives yet. If ASan reports something is wrong, it probably is.
* Already found several relatively hard to spot bugs, see #193 and #189
* By default it also performs checks for memory leaks. Although it would be nice to keep this turned on, for now it can be disabled by setting the environment variable `ASAN_OPTIONS` to `detect_leaks=0`.

tquatmann:
I'm pretty much in favor of adding such cmake options (and already thought about doing this as well).

I'd propose to call the flag `-DSTORM_COMPILE_WITH_ALL_SANITIZERS` as the word "checker" can be misleading in the context of model *checkers*.

Marckvdv:
Closing the issue as the PR is merged. If we at some point we want other tools integrated they can be added as separate issues.

PR #197, opened 2022-03-04
Add AddressSanitizer as build configuration option

Marckvdv:
Adds AddressSanitizer as build configuration option and enables it for the 'indepthTests' jobs inside the 'Build Test' workflow. See related issue #194. For now the memory leak detection is turned off while running the tests as the current code does have some memory leaks, but that is another issue.

You can see the new workflow passing at [my fork](https://github.com/Marckvdv/storm/actions/workflows/buildtest.yml).


tquatmann:
Looks good to me!

> For now the memory leak detection is turned off while running the tests as the current code does have some memory leaks, but that is another issue.

Could you add an issue for that?

PR #200, opened 2022-03-28
POMDP almost-sure minor fix

sjunges:
Minor fix of a bug reported by Marck (notice that the parentheses dont make sense before the fix).


tquatmann:
LGTM
can be merged if CI is without errors.

PR #201, opened 2022-04-03
Fix Gradient Descent bugs

glatteis:
I've fixed some bugs in Gradient Descent, e.g.
- used to fail when the initial state wasn't zero
- used to fail when the formula wasn't "at some point target" (so storm-pars.cpp preprocessing), now it supports arbitrary labels and until formulas
- used to fail when there are transitions that have higher degree than one
- applied code formatting (makes the diff quite unreadable though :( )

These are all things that never hold for pMCs coming out of storm-pomdp, so I didn't catch these bugs previously.

I added a test for the first bug (applying bisimulation in a test, which results in the initial state not being zero), but not for the other ones yet - that's something I still need to do.

### DIFF
--- a/src/storm-pars/derivative/GradientDescentConstraintMethod.h
+++ b/src/storm-pars/derivative/GradientDescentConstraintMethod.h
@@ -3,19 +3,19 @@
 #include <boost/optional.hpp>
 #include <string>
 namespace storm {
-	namespace derivative {
-		/**
-		 * GradientDescentConstraintMethod is the method for mitigating constraints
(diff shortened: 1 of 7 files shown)


sjunges:
Hi Linus,

Thanks for these bugfixes. In the future, I would recommend not to push code formatting in the same pull request; this makes this a hard to process PR.

As all changes are local to the GD code, I am fine with merging it. 

@jipspel Could you have a look before we merge it?

glatteis:
Yeah, sorry for the formatting :D I didn't think about it when I was writing the PR and afterwards I couldn't "unformat" it

PR #203, opened 2022-04-15
Fix LP-based MDP Model Checking

sjunges:
The LP model checker appearantly did not set the optimization direction.


tquatmann:
LGTM

Issue #69, opened 2020-04-14
Move From Nondeterministic to Deterministic Model


sjunges:
What is the  right way to force an MDP to being a DTMC, without much overhead? That is, if I have an MDP (e.g., after applyScheduler with a full scheduler) and I want to get an DTMC, how do I make the resulting MDP a DTMC? Is there any code doing that already?

This is related to https://github.com/moves-rwth/storm/issues/63 and  https://github.com/moves-rwth/stormpy/issues/14

tquatmann:
AFAIK there currently is a getter for each model component, that retrieves it by (non-const) reference. It should be possible to do
`std::make_shared<DTMC<VT>>(std::move(mdp.getTransitionMatrix()), std::move(mdp.getStateLabeling()), std::move(mdp.getRewardModels);`
(this disregards other components like choice labels)

Another more elegant way could be to write the inverse of [this guy](https://github.com/moves-rwth/storm/blob/1f6fc7e2737b085f3d235cb4c368958b243d16c0/src/storm/utility/builder.h).
Then you could 
* get the model components
* change the model type of the components
* and then execute buildModelFromComponents

I'm not sure, but you might also need `SparseMatrix.makeRowGroupingTrivial()`

PR #206, opened 2022-05-03
Induced scheduler

sjunges:
When applying a fully defined scheduler to an MDP, we obtain a DTMC; this allows applying DTMC-specific code on this model.

### DIFF
--- a/src/storm/transformer/SubsystemBuilder.cpp
+++ b/src/storm/transformer/SubsystemBuilder.cpp
@@ -134,6 +134,10 @@ SubsystemBuilderReturnType<ValueType, RewardModelType> internalBuildSubsystem(st
     } else {
         components.transitionMatrix = originalModel.getTransitionMatrix().getSubmatrix(false, keptActions, subsystemStates);
     }
+    if (options.makeRowGroupingTrivial) {
+        components.transitionMatrix.makeRowGroupingTrivial();
+    }
+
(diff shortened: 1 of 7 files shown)


sjunges:
Additionally added a flag that enforces that an MDP remains an MDP etc; as this is sometimes useful when writing concise code

src/storm/transformer/SubsystemBuilder.cpp
```diff
@@ -134,6 +134,10 @@ SubsystemBuilderReturnType<ValueType, RewardModelType> internalBuildSubsystem(st
     } else {
         components.transitionMatrix = originalModel.getTransitionMatrix().getSubmatrix(false, keptActions, subsystemStates);
     }
+    if (options.makeRowGroupingTrivial) {
```
tquatmann, line 137: Do we have to catch the case where the transitionMatrix is still supposed to be nondeterministic (i.e. not square)?

tquatmann (commented):
I think it would be good to catch that one edge case in the subsystem builder with e.g. an assertion. Other then that, this looks good to me 👍

src/storm/transformer/SubsystemBuilder.cpp
sjunges: Yes, indeed, will do!

sjunges:
This also relates to issue #69 (it doesnt solve it, but it alleviates some cases and now allows to use the subsystem builder appropriately).

PR #211, opened 2022-05-11
Code formatting for storm-dft

volkm:
Enable automatic code formatting for `storm-dft`, `storm-dft-cli` and `test/storm-dft`.

### DIFF
--- a/.clang-format-ignore
+++ b/.clang-format-ignore
@@ -1,9 +1,6 @@
-./src/storm-dft/*
-./src/storm-dft-cli/*
 ./src/storm-pars/*
 ./src/storm-pars-cli/*
 ./src/storm-pomdp/*
 ./src/storm-pomdp-cli/*
-./src/test/storm-dft/*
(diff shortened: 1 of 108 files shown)


sjunges:
LGTM

tquatmann:
For the historians among us: I recommend to add the code-formatting commit hash to the `.git-blame-ignore-revs` file.

If you use the github workflow to apply the actual code formatting, it will add the commit hash automatically and also uses a dedicated committer name. Maybe that's preferable :)

sjunges:
> For the historians among us: I recommend to add the code-formatting commit hash to the `.git-blame-ignore-revs` file.
> 
> If you use the github workflow to apply the actual code formatting, it will add the commit hash automatically and also uses a dedicated committer name. Maybe that's preferable :)

Ah that is how that works! Great!

volkm:
So the preferred way would be to only remove the storm-dft directories from `.clang-format-ignore` and let the github workflow take care of the rest?

PR #212, opened 2022-05-11
Added Dockerfile template

volkm:
The Dockerfile can be used as a template for custom libraries/scripts/artifacts/etc.


doc/scripts/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+WORKDIR /opt/storm/build
+# If needed, checkout a specific branch
+#RUN git checkout specific_branch
+# Configure Storm
+RUN cmake .. "-DCMAKE_BUILD_TYPE=Release -DSTORM_DEVELOPER=OFF -DSTORM_LOG_DISABLE_DEBUG=ON -DSTORM_PORTABLE=ON"
+# Build binaries of Storm using a single thread
+RUN make binaries -j 1
+ENV PATH="/opt/storm/build/bin:$PATH"
+
+
```
sjunges, line 32: I would remove the build additional libraries; generally I do that in a separate docker file etc. We can add a docker file to the starter-template gits that does that part?

sjunges:
Generally great; I just have two design decisions I do not fully understand. Happy to discuss offline and summarize our result here.

doc/scripts/Dockerfile
volkm: Sounds good. Then this docker file is just for building the base Storm version.

volkm:
Following our discussion, I restructured the Dockerfile and moved it to the root folder of Storm. The Dockerfile now enables the compilation of everything inside the current repo. Additional libraries must be specified in their own Dockerfile.
@sjunges Maybe check whether `.dockerignore` is correct.

sjunges:
LGTM!

PR #215, opened 2022-05-11
more robust parsing for DRN files

sjunges:
Made DRN parsing more robust with better and more error messages and being less reliant on tabs vs spaces.


src/storm-parsers/parser/DirectEncodingParser.cpp
```diff
@@ -348,14 +355,16 @@ std::shared_ptr<storm::storage::sparse::ModelComponents<ValueType, RewardModelTy
 
         } else {
             // New transition
+            line = boost::trim_left_copy(line);
```
volkm, line 358: Why not use `trim_left`?

volkm (approved):
I would move the trimming to the beginning.
Other than that, everything is fine.

PR #220, opened 2022-05-13
CI tests use Ubuntu 22.04

volkm:
- Updated Github workflows to use new Ubuntu 22.04 (LTS) instead of Ubuntu 21.10
- Also build Carl without carl-parser. The parser is explicitly installed in the stormpy CI where it is needed.


sjunges:
This looks good. Is it correct that for 22.10, we would need to consider whether we put 22.04 to the LTS versions we also test on (and/or whether we want to abandon support for 18.04 which I think is justified).

PR #222, opened 2022-05-18
Added dedicated namespace storm::dft

volkm:
Moved all sources inside the `storm-dft` library into the namespace `storm::dft` and moved some files according to their namespace.


volkm:
These changes touch all files within the storm-dft directory. Maybe we want to add these commits to `.git-blame-ignore-revs` as well?

tquatmann:
Yeah, adding to .git-blame-ignore-revs  makes sense to me.

volkm:
Should I then squash all commits and add a single commit or use all commit hashes to get a better distinction?

Issue #221, opened 2022-05-14
Prism Language issue with mod() function


stevencarrau:
This issue is probably an extension of the discussion for C and modulo (https://stackoverflow.com/questions/4003232/how-to-code-a-modulo-operator-in-c-c-obj-c-that-handles-negative-numbers).

The implementation of the built-in function for mod(i,n) does not output the correct values for i if i<0. For example , -1 % 4 should output 3, instead of -1 (currently implemented). PRISM programs that are compatible with the PRISM model checker don't build on storm due to this issue.

One example in a module for rotating a robot (4 directions):

module robot
    ar : [0..3] init 0;
    [left] true -> 1: (ar'=mod(ar-1,4));
    [right] true -> 1:  (ar'=mod(ar+1,4));
endmodule

Throws an issue when ar=0 -> ar'=-1.

sjunges:
Thanks for catching this Steve!

@tquatmann (and others) I always found mod on negative numbers confusing, but I feel like its a good idea to be aligned with PRISM here (and potentially warn users about the discussion). What do you think?

sjunges:
Summarizing our discussion: We want to investigate the overhead of a solution that transforms `mod(a,b)` into `ITE(a >= 0, mod(a,b), mod(a,b)+b)`; i.e., see whether this negatively affects some of the QVBS benchmarks that have a mod operation *without a negative number*. 

@Marckvdv would you look into this?

stevencarrau:
This is also the way I would implement it. One thing to flag is if a+b< 0 then it may still throw the same issue. But for the module I wrote above the mod(a,b)+b fix would be more than sufficient. I don't see many instances where this would occur though.

sjunges:
Hi,

yes, I assumed that b is still positive. Doing modulo with a negative number as second argument should imo just be forbidden :-). Would be good to check that we handle that situation though.

@stevencarrau it is great if you can use this workaround for now, but I think we want to resolve this issue for future instances as this is not the first time we encounter this discrepancy... ;-). I will flag the thread accordingly.

Marckvdv:
Sure, I can take a look. We'll have to make sure it works for the exact versions too then.

PR #225, opened 2022-05-24
Remainder -> Modulo in accordance with Prism

Marckvdv:
Regarding issue #221. Correctly implements the modulo operator for expressions in accordance with Prism, by introducing an ITE for dealing with negative numbers. I added a test to check the behavior.

### DIFF
--- a/src/storm/storage/expressions/Expression.cpp
+++ b/src/storm/storage/expressions/Expression.cpp
@@ -543,9 +543,7 @@ Expression sum(std::vector<storm::expressions::Expression> const& expressions) {
 }
 
 Expression modulo(Expression const& first, Expression const& second) {
-    return Expression(std::shared_ptr<BaseExpression>(new BinaryNumericalFunctionExpression(
-        first.getBaseExpression().getManager(), first.getType().modulo(second.getType()), first.getBaseExpressionPointer(), second.getBaseExpressionPointer(),
-        BinaryNumericalFunctionExpression::OperatorType::Modulo)));
+    return first % second;
 }
 
 Expression apply(std::vector<storm::expressions::Expression> const& expressions,
--- a/src/storm/storage/expressions/Expression.cpp
+++ b/src/storm/storage/expressions/Expression.cpp
@@ -300,9 +300,13 @@ Expression operator/(Expression const& first, Expression const& second) {
 
 Expression operator%(Expression const& first, Expression const& second) {
     assertSameManager(first.getBaseExpression(), second.getBaseExpression());
-    return Expression(std::shared_ptr<BaseExpression>(new BinaryNumericalFunctionExpression(
+    // First we compute the remainder
+    Expression remainder(std::shared_ptr<BaseExpression>(new BinaryNumericalFunctionExpression(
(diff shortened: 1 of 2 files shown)


src/storm/storage/expressions/Expression.cpp
```diff
@@ -300,9 +300,13 @@ Expression operator/(Expression const& first, Expression const& second) {
 
 Expression operator%(Expression const& first, Expression const& second) {
     assertSameManager(first.getBaseExpression(), second.getBaseExpression());
-    return Expression(std::shared_ptr<BaseExpression>(new BinaryNumericalFunctionExpression(
+    // First we compute the remainder
+    Expression remainder(std::shared_ptr<BaseExpression>(new BinaryNumericalFunctionExpression(
         first.getBaseExpression().getManager(), first.getType().modulo(second.getType()), first.getBaseExpressionPointer(), second.getBaseExpressionPointer(),
         BinaryNumericalFunctionExpression::OperatorType::Modulo)));
+
+    // Deal with negative `first` in accordance with Prism
```
tquatmann, line 308:
```suggestion
    // Deal with negative `first` in accordance with Prism and Jani
```
The Jani spec says:
> Unlike e.g. in most implementations of C, this operation computes the remainder of the Euclidean division: the result is always positive and x % 0 is division by zero.

As far as I can see, this is what we are doing now with this fix, right?

Marckvdv: Yes, except for the `x % 0` part. For normal integers that is going to give some divide/modulo zero error.

tquatmann:
I just realized that `storm/storage/expressions/Expression.cpp` also implements a 
```
Expression modulo(Expression const& first, Expression const& second);
```
I guess this fix also has to be applied there?

Marckvdv:
Looking at the code, the function body of `modulo` can simply be replaced by `return first % second;`.

Alternatively we can decide to put the actual code in `modulo` and replace the function body of `operator%` with `return modulo(first, second)`.

tquatmann:
I created a prism file with a *ridiculous* amount of (nested!) modulos in there to check how much this can affect performance.
The model checking time went from `5s` (without the fix) to `15s` (with the fix). My suggestion above did not yield any improvement on this particular model.

Modulo does not seem to occur very often and my example is very artificial, so I don't object merging this as a simple fix to #221. But IMO we should create an issue  for future reference.

Here's my prism file which I built with `-const N=4,Q=4`: [polling.txt](https://github.com/moves-rwth/storm/files/8770893/polling.txt)

tquatmann:
> Looking at the code, the function body of `modulo` can simply be replaced by `return first % second;`.
> 
> Alternatively we can decide to put the actual code in `modulo` and replace the function body of `operator%` with `return modulo(first, second)`.

Good point! I think it's fine either way.

Marckvdv:
Yeah it is too bad that there is no separation of modulo/remainder in ExprTk... Some alternatives could be:
* `(n + (i % n)) % n`: branchless but probably much slower
* `i - floor(i/n)*n`: implementing modulo manually

PR #229, opened 2022-06-10
Updated deprecated std::iterator

volkm:
`std::iterator` is deprecated since C++17. Updated it according to https://www.fluentcpp.com/2018/05/08/std-iterator-deprecated/


volkm:
Tests are successful.

PR #230, opened 2022-06-22
Select Gurobi Method

sjunges:


src/storm/solver/GurobiLpSolver.cpp
```diff
@@ -23,86 +23,111 @@
 namespace storm {
 namespace solver {
 
+GurobiEnvironment::~GurobiEnvironment() {
+#ifdef STORM_HAS_GUROBI
+    GRBfreeenv(env);
```
tquatmann, line 28:
```suggestion
    if(initialized) {
        GRBfreeenv(env);
    }
```

src/storm/solver/GurobiLpSolver.h
```diff
@@ -17,6 +18,25 @@ int __stdcall GRBislp(GRBenv**, const char*, const char*, const char*, const cha
 namespace storm {
 namespace solver {
 
+class GurobiEnvironment {
+   public:
+    GurobiEnvironment() = default;
```
tquatmann, line 23: One shouldn't be able to create a copy of this 
```suggestion
    GurobiEnvironment() = default;
    GurobiEnvironment(GurobiEnvironment const&) = delete;
    GurobiEnvironment& operator=(GurobiEnvironment const&) = delete;
```

tquatmann (commented):
LGTM! I just have a few minor suggestions

The (empty) file `resources/cmake/find_modules/FindCPLEX.cmake` is included for some reason.

PR #233, opened 2022-06-29
Removed forcing old version of optional (Fixes #80 Fixes #232)

### LINKED ISSUE #80: Boost: Old version of optional

### LINKED ISSUE #232: Version issue with `boost` (workaround)
**Short Bug Description**: On systems where a version of `boost` newer than 1.78.0 is installed but 1.78 is not installed, STORM does not properly link, even though an executable is created. On runtime, an error is emitted:
```
./storm: error while loading shared libraries: libboost_filesystem.so.1.78.0: cannot open shared object file: No such file or directory
```
**More Information**: It appears in CMakeLists.txt:269-272 the following code forces an old version of `boost::optional` to be used:
```cmake
# TODO: remove forcing the old version of optional as soon as the related Spirit bug is fixed:
# https://svn.boost.org/trac/boost/ticket/12349
# Fix thanks to: https://github.com/freeorion/freeorion/issues/777
add_definitions(-DBOOST_RESULT_OF_USE_DECLTYPE -DBOOST_OPTIONAL_CONFIG_USE_OLD_IMPLEMENTATION_OF_OPTIONAL)
```
However, I have used STORM (mostly through our project, STAMINA, which is built on top of STORM), using Boost: 1.79 with no issue. I would like to make it clear that **THIS HAS JUST BEEN MY EXPERIENCE. BOOST 1.79 MAY BREAK SOMETHING IN STORM FOR A FEATURE I DON'T GENERALLY USE**. Therefore I have used the following workaround on my system:

**Easy workaround**: Create symlinks that link libboost_*.so.1.* to libboost_*.so.1.78. For my system, which has 1.79 on it, the following script does it for me:

```bash
sudo ln -T /usr/lib/libboost_filesystem.so.1.79.0 /usr/lib/libboost_filesystem.so.1.78.0
sudo ln -T /usr/lib/libboost_system.so.1.79.0 /usr/lib/libboost_system.so.1.78.0
sudo ln -T /usr/lib/libboost_program_options.so.1.79.0 /usr/lib/libboost_program_options.so.1.78.0
sudo ln -T /usr/lib/libboost_regex.so.1.79.0 /usr/lib/libboost_regex.so.1.78.0
sudo ln -T /usr/lib/libboost_unit_test_framework.so.1.79.0 /usr/lib/libboost_unit_test_framework.so.1.78.0
```

**More Permanent Solution**: If the Spirit bug no longer exists in boost, or if it doesn't affect STORM, you could simply remove lines 269-272 in `CMakeLists.txt`, but this needs to be tested before done. I am posting this workaround in this bug report because it worked for me so others can try it and see if it works for them. Note that Boost 1.78 is not available in my `pacman` repositories.

**System Information**:
System type: Linux (Manjaro)
Kernel: 5.18.6-1-MANJARO \#1 SMP PREEMPT_DYNAMIC
Architecture type: x86_64 GNU/Linux
DE: Plasma 5.24.5 
Shell: zsh 5.9 (x86_64-pc-linux-gnu)
Boost Version: 1.79
`git` branch of STORM used: HEAD -> master (hash: a491f13e738ba4491fc26acca351e20bbe9506fc)

tquatmann:
This PR removes a workaround for an old bug in boost::spirit.

This fixes issue #80 and issue #232
According to  #80, this PR breaks support for ubuntu 18-04 (which is fine in my opinion)


volkm:
Looks good.
Maybe we should explicitly set a minimal Boost version or at least issue a warning if the Boost version is too old.

volkm:
From the tests it seems that Ubuntu 18.04 also works with this change. Maybe PR #234 is not necessary?

tquatmann:
Indeed, we should increase the minimal required  boost version. `resources/3rdparty/CMakeLists.txt` currently requires at least Boost 1.61.0 . 

Yes, as it seems to work with 18-04, I'll discard PR #234 and set the minimum version to 1.65.1

volkm:
Thanks for integrating the suggestion.

PR #234, opened 2022-06-29
CI: Disable 'build and test' workflow for ubuntu 18-04

tquatmann:
~PR #233~ PR #273 breaks support for ubuntu 18-04.
This PR removes 18-04 from our CI workflow.

**EDIT:** Apparently, #233 does not actually break ubuntu 18-04 builds. 
**EDIT2:** But #273 does!

### DIFF
--- a/.github/workflows/buildtest.yml
+++ b/.github/workflows/buildtest.yml
@@ -60,7 +60,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        distro: ["debian-10", "debian-11", "ubuntu-18.04", "ubuntu-20.04"]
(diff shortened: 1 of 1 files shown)


sjunges:
Given that we released 1.7 and that Ubuntu has had two LTS releases since 18.04, I dont see a big problem breaking default ubuntu 18.04 support.

volkm:
I am also fine with removing 18.04. We should also update the [documentation](https://www.stormchecker.org/documentation/obtain-storm/build.html#supported-operating-systems).

sjunges:
Notice that Stormpy also moved away from Ubuntu 18.04 (definitively with https://github.com/moves-rwth/stormpy/pull/94)

PR #235, opened 2022-07-05
DRN Parser: Fix off-by-one issue

tquatmann:
The current code throws a
```
ERROR (DirectEncodingParser.cpp:209): More actions detected than declared (in @nr_choices).
```
when parsing the very last choice, even if the `@nr_choices` hint is correct.
This PR fixes the issue.


volkm:
Good catch. This fix was overlooked in 653fe0b320f811274049c062c596077f0a53e297.

PR #245, opened 2022-07-25
multiobjective: Fixed properly dealing with unnamed reward models

### LINKED ISSUE #238: Misleading error message for multiobjective
When running storm with multi-objective, I believe that the following error message is unclear and potentially misleading. 

```
Storm 1.6.5 (dev)

Command line arguments: --prism examples/grid_complete.prism -const N=6 --prop 'multi(Rmin=? [F( "station") ],Rmin=? [F "station"])'

Time for model input parsing: 0.000s.

Time for model construction: 0.020s.

Model type: 	MDP (sparse)
States: 	36
Transitions: 	240
Choices: 	120
Reward Models:  xdominated
State Labels: 	3 labels
   * init -> 1 item(s)
   * deadlock -> 0 item(s)
   * station -> 1 item(s)
Choice Labels: 	none

Model checking property "1": multi(R[exp]min=? [F "station"], R[exp]min=? [F "station"]) ...
ERROR (SparseMultiObjectivePreprocessor.cpp:353): The reward model is not unique or the formula R[exp]min=? [F "station"] does not specify an existing reward model.
 WARN (model-handling.h:986): Cannot handle property: InvalidPropertyException: The reward model is not unique or the formula R[exp]min=? [F "station"] does not specify an existing reward model.
ERROR (model-handling.h:946): Property is unsupported by selected engine/settings.
```

However, changing the command line to

`Command line arguments: --prism examples/grid_complete.prism -const N=6 --prop 'multi(R{"xdominated"}min=? [F( "station") ],Rmin=? [F "station"])'`

does work and gives the (intended, trivial) result. 

The prism file is as follows:

```
mdp

const int N;

module main
   x : [1..N] init 1;
   y : [1..N] init 3;

   [north] x > 1 -> 0.9:(x'=x-1) + 0.1:(x'=x);
   [south] x < N -> 0.9:(x'=x+1) + 0.1:(x'=x);
   [west] y > 1 -> 0.8:(y'=y-1) + 0.2:(y'=y);
   [east] y < N -> 0.8:(y'=y+1) + 0.2:(y'=y);
endmodule

label "station" = x=3 & y=1;
label "castle" = x=N-1 & y=N-1;

rewards "xdominated"
    true : x;
endrewards
```

tquatmann:
fixes #238

### DIFF
--- a/src/storm/modelchecker/multiobjective/preprocessing/SparseMultiObjectivePreprocessor.cpp
+++ b/src/storm/modelchecker/multiobjective/preprocessing/SparseMultiObjectivePreprocessor.cpp
@@ -347,20 +347,30 @@ template<typename SparseModelType>
 void SparseMultiObjectivePreprocessor<SparseModelType>::preprocessRewardOperatorFormula(storm::logic::RewardOperatorFormula const& formula,
                                                                                         storm::logic::OperatorInformation const& opInfo,
                                                                                         PreprocessorData& data) {
-    STORM_LOG_THROW((formula.hasRewardModelName() && data.model->hasRewardModel(formula.getRewardModelName())) ||
-                        (!formula.hasRewardModelName() && data.model->hasUniqueRewardModel()),
-                    storm::exceptions::InvalidPropertyException,
-                    "The reward model is not unique or the formula " << formula << " does not specify an existing reward model.");
-
     std::string rewardModelName;
     if (formula.hasRewardModelName()) {
         rewardModelName = formula.getRewardModelName();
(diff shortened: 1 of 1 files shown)


sjunges:
LGTM, but it relies rather heavily on the string naming scheme. Shall we add an issue for flagging reward models as auxiliary or something?

tquatmann:
The `data.rewardModelNamePrefix` is computed such that it is not a prefix of any pre-existing reward model name. Therefore, the *current* code is safe in that regard.

I get your point that it is still a bit hacky, but at least the hack is local to the `SparseMultiObjectivePreprocessor`. I also don't really like an auxiliary flag for reward models, because its meaning is not very clear in the general context[^1].



[^1]: I also see this with the auxiliary expression variables that we already have

PR #247, opened 2022-07-25
prism-to-jani: fixed detection of variables to make global

### LINKED ISSUE #205: storm-conv has issue with single state variable
When converting a model from prism to jani, the storm-conv throws an error regarding a single state variable:

The model in specific is [Toggle_10_10.txt](https://github.com/moves-rwth/storm/files/8575559/Toggle_10_10.txt)
 and the command we used to run the converter is as follows:
`storm-conv --input:prism Toggle_10_10.txt -pc --output:tojani Toggle_10_10.jani`

The output of the command:
```storm-conv --input:prism Toggle_10_10.sm -pc --output:tojani Toggle_10_10.jani
Storm-conv 1.6.4 (dev)

Date: Wed Apr 27 12:29:18 2022
Command line arguments: '--input:prism' Toggle_10_10.sm -pc '--output:tojani' Toggle_10_10.jani
Current working directory: /home/pgiles/prismModels

Parsing PRISM input ...  WARN (Program.cpp:179): The input model is a CTMC, but uses probabilistic commands like they are used in PRISM. Consider rewriting the commands to use Markovian commands instead.
 WARN (Program.cpp:1355): The model uses synchronizing Markovian commands. This may lead to unexpected verification results, because of unclear semantics.
 WARN (Program.cpp:1355): The model uses synchronizing Markovian commands. This may lead to unexpected verification results, because of unclear semantics.
 WARN (Program.cpp:1355): The model uses synchronizing Markovian commands. This may lead to unexpected verification results, because of unclear semantics.
 done. (0.003s seconds).
Converting PRISM Program to JANI model ...  done. (0.000s seconds).
Exporting JANI model ... ERROR (JSONExporter.cpp:667): Expression variable '_A' not known in Jani data structures.
ERROR (storm-conv.cpp:386): An exception caused Storm-conv to terminate. The message of the exception is: InvalidJaniException: Expression variable '_A' not known in Jani data structures.
```

tquatmann:
fixes #205

The issue was that variable declarations were not considered when finding the set of modules that access a given variable.
As a consequence, if a variable `x` is declared in a module `A` but *only* used in a different module `B`, then we assumed that only a single module accesses `x` and hence `x` was not made a global variable.

Now we also consider variable declarations, so `x` is accessed by both modules `A` and `B`.

I did consider further optimisations (like moving `x` to the scope of `B`; or even making `x` a constant since it is never written to). However, I consider this a rare edge case so I did not want to make things unnecessarily complicated.

### DIFF
--- a/src/storm/storage/prism/ToJaniConverter.cpp
+++ b/src/storm/storage/prism/ToJaniConverter.cpp
@@ -174,7 +174,7 @@ storm::jani::Model ToJaniConverter::convert(storm::prism::Program const& program
         storm::prism::Module const& module = program.getModule(index);
 
         // Gather all variables occurring in this module
-        std::set<storm::expressions::Variable> variables;
+        std::set<storm::expressions::Variable> variables = module.getAllExpressionVariables();
         for (auto const& command : module.getCommands()) {
             command.getGuardExpression().getBaseExpression().gatherVariables(variables);
             for (auto const& update : command.getUpdates()) {


sjunges:
LGTM. I would also not optimize for this case.

PR #251, opened 2022-07-26
Updated clang-format-lint-action to version 0.14

volkm:
This allows to use clang-format version 14 which is also the current version in Homebrew.
Clang-format version 14 leads to small formatting differences compared to version 12 (notably for `StandardRewardModel`).


tquatmann:
LGTM!

Shall we merge this and apply the formatting action afterwards so that the tests do not fail?

volkm:
Yes, sounds good to me. I will merge. Afterwards the CI should take care of the formatting.

PR #260, opened 2022-07-28
Updated Dockerfile with Spot support

volkm:
- Use `storm-basesystem:latest` as common base now. This avoids having to update the Dockerfile every time a new Ubuntu version is released.
- Added missing flag for carl
- Build with Spot


sjunges:
LGTM; thanks!

PR #261, opened 2022-07-28
Printing finite-state schedulers requires model 

sjunges:
This prints an error if this is not the case, instead of segfaulting.


volkm:
LGTM

PR #271, opened 2022-08-02
do not print diagnostics, instead use logging

sjunges:


volkm:
There are actually a lot of other instances where we are using `STORM_PRINT` as well. Maybe we want to change the occurrences outside the cli files to `STORM_LOG_INFO` as well at some point.
For now, it looks good to me.

PR #273, opened 2022-08-03
Sylvan upgrade

tquatmann:
This PR upgrades the shipped sylvan version.

Unfortunately, there is a notable performance degradation (around 20-30% slowdown in the dd engine for me; even more if the number of workers is not limited).
 I prefer to have that investigated a bit more before merging this PR.

On the plus side, I was able to compile this branch natively on Apple's m1 (see issue #274)
 

*Thanks @dbasgoeze for doing most of the work!*


tquatmann:
This breaks support for ubuntu 18.04 due to the cmake version required by sylvan.
I reopened PR #234 for this.

volkm:
@trolando

trolando:
So far, I have found that parallel performance seriously degraded with commit 2d5d83 in Sylvan. This is a rewrite of the CMake files. What I have discovered so far is that if testing is enabled for Sylvan, the build script adds "coverage" to compiling which means adding `-fprofile-arcs -ftest-coverage` which is pretty pretty bad for performance.

trolando:
I made a small script to run a few performance tests:
```
#!/bin/bash
FILES="anderson.4.bdd bakery.4.bdd schedule_world.2.bdd lifts.6.bdd"
for f in $FILES; do
    # echo -ne "SAT $f: "
    echo $f
    examples/bddmc -s bfs -w 1 ../models/$f | grep Time
    examples/bddmc -s par -w 1 ../models/$f | grep Time
    examples/bddmc -s par -w 2 ../models/$f | grep Time
    examples/bddmc -s par -w 4 ../models/$f | grep Time
    examples/bddmc -s sat -w 1 ../models/$f | grep Time
done
```

Here are my results for Sylvan tag `v1.5.0` cherry-picking `a4fac` (to fix warnings on modern gcc):
```
anderson.4.bdd
[    0.17] BFS Time: 0.165092
[    0.17] PAR Time: 0.163825
[    0.10] PAR Time: 0.095206
[    0.06] PAR Time: 0.056248
[    0.13] SAT Time: 0.129423
bakery.4.bdd
[    0.57] BFS Time: 0.559362
[    0.58] PAR Time: 0.565135
[    0.33] PAR Time: 0.320133
[    0.19] PAR Time: 0.174657
[    0.38] SAT Time: 0.365768
schedule_world.2.bdd
[    3.53] BFS Time: 3.527073
[    3.53] PAR Time: 3.522686
[    1.85] PAR Time: 1.840404
[    1.02] PAR Time: 1.013537
[    0.15] SAT Time: 0.141081
lifts.6.bdd
[   11.47] BFS Time: 11.445683
[   11.89] PAR Time: 11.866713
[    6.04] PAR Time: 6.009242
[    3.29] PAR Time: 3.247432
[    4.30] SAT Time: 4.272830
```

Here are my results for the `HEAD` of the `master` branch in Sylvan, then removing the coverage stuff from CMake:
```
anderson.4.bdd
[    0.15] BFS Time: 0.145815
[    0.15] PAR Time: 0.147422
[    0.09] PAR Time: 0.085670
[    0.06] PAR Time: 0.051587
[    0.13] SAT Time: 0.118849
bakery.4.bdd
[    0.51] BFS Time: 0.502833
[    0.51] PAR Time: 0.495781
[    0.28] PAR Time: 0.267565
[    0.17] PAR Time: 0.151103
[    0.32] SAT Time: 0.307729
schedule_world.2.bdd
[    2.97] BFS Time: 2.962590
[    2.97] PAR Time: 2.968177
[    1.64] PAR Time: 1.634357
[    0.93] PAR Time: 0.918933
[    0.13] SAT Time: 0.127762
lifts.6.bdd
[    8.99] BFS Time: 8.967668
[    9.06] PAR Time: 9.036274
[    4.84] PAR Time: 4.808833
[    2.70] PAR Time: 2.660959
[    3.38] SAT Time: 3.351562
```

As you see, at least on my Desktop computer using the 6-core `Intel Core i5-8400` and compiling with `gcc 12.1.1` I get a pretty clear performance improvement running symbolic exploration of a few example models.

trolando:
I have now pushed Sylvan v1.7.0 on my github repository, I have not yet checked if this fixes the issue in Storm, in part because I cannot build Storm right now since `libeiger` is down for maintenance.

trolando:
@tquatmann I am not sure if the higher CMake version is still needed in Sylvan v1.7.0, as I have pruned the CMake script a bit. But I have not yet tried to lower the version yet.

trolando:
I still notice a clear problem in how Storm uses Sylvan, which is related to the changes in Lace.

**In the old way**, initializing Lace would create `n_workers-1` threads as Lace workers and initialize the current thread as a Lace worker. This is actually not very great for interoperability with other languages and with software where Lace can be initialized from a thread that is later killed.

**In the new way**, initializing Lace creates `n_workers` threads as Lace workers. The current thread is not a Lace thread. Sylvan expects that its methods are called from inside a Lace worker, especially methods like `bdd_refs_push` and `sylvan_makenode`. If run outside Lace threads, then any parallel garbage collection likely results in data corruption. Furthermore, calling Sylvan macros that use `RUN` from outside a Lace worker causes a context switch to a different thread.

**Recommendation**: after initializing Lace, I would recommend running the rest of the program as a task that is run by the Lace framework. But if Storm is itself also running several threads, then I recommend that any BDD-intensive parts are executed as tasks inside Lace. In any case, methods like `bdd_refs_push` and `sylvan_makenode` should never be called outside Lace and I think I should add Debug assertions for this, because it is very easy to make this mistake.

sjunges:
@trolando thanks for all the input! What are the consequences of running everything inside Lace?

trolando:
> thanks for all the input! What are the consequences of running everything inside Lace?

A few minor inconveniences:
- you can't disable Lace workers if you want to do that for some reason (but this could be added as a feature)
- you need to wrap your program in a `TASK` after initializing Lace, so you probably need to start Lace 

Some advantages:
- you can run Lace tasks without extra overhead of 1) creating a dummy task to hold all parameters 2) waiting for a worker to steal it 3) waiting for the task to be done...
- you don't need to create threads for multithreaded execution (with the disclaimer that you can only run as many tasks simultaneously as there are Lace workers)

sjunges:
Thinking about it, there are two concerns that I want to raise:

- python bindings
- other tools, potentially running multithreaded code, that use storm as a library

There are two parts to this:

- can they use code that is running sylvan and do they need to do extra work then (acceptable)
- can they use code that is independent of BDDs and do they need to do extra work then (rather not acceptable, I guess)

trolando:
So if Storm is used from Python bindings I think wrapping Storm in Lace is not sensible. In that case I recommand that whenever you're going to perform some expensive operation on Storm, wrap that in the Lace task. For example, you could have the interface such that any Storm functionality using BDD operations is wrapped in a Lace `TASK`.

I guess it only makes sense to wrap Storm entirely in Lace if it is run standalone, but if you run Storm as a library, I would not recommend that. With the 'old' Lace design, this would be problematic for the same reason by the way, which highlights that the new design is probably better?

trolando:
There is by the way no clean way of having a "single-threaded" Sylvan with the current design. For example, the `bdd_refs_push` methods use thread-local stacks and the garbage collector needs to keep track of each thread-local stack. This happens when Sylvan is initialized (it creates these stacks in every Lace worker thread). So in any case, you need to create a special Lace worker for running Sylvan functionality in. But you could use `lace_suspend` and `lace_resume` to reduce performance loss from the busy-wait loop, and I could imagine a feature that automatically uses `lace_resume` and `lace_suspend` with the `RUN` macro whenever Lace is suspended.

trolando:
By the way, I discovered CMake 3.14 has very nice support for dependencies now using FetchContent, so I played a bit with it and now I can just include Lace as a dependency for Sylvan. Seems to work pretty well! (This is pretty useful for me because it allows me to do the same for my library Oink which uses Lace as well, and then Knor uses both Sylvan and Oink and currently has two copies of Lace, rather than letting CMake handle the dependency nicely.)

sjunges:
I agree that it is a design improvement, I am just trying to figure out where to initialize Lace :-)

tquatmann:
Thanks a lot for all the helpful insights @trolando !
I think the todo list for this PR is

- [ ] update the sylvan version used in this PR
- [ ] when auto-detecting the number of lace threads to use, take `max(1,lace_get_pu_count() - 1)` so that there is an ample thread for storm. Issue a warning if the user sets a larger number of threads.
- [ ] check if this irons out the performance regressions

From there, we can improve our lace usage (and eventually get rid of the `lace_get_pu_count() - 1` above). I'm in favour of putting the DD-intensive parts into lace tasks and make sure that lace is suspended outside of those tasks.

tquatmann:
As this now requires CMake 3.14, debian-10 builds fail as well. 
I'm not sure how much of a problem that would be. I guess if we still want to support debian 10, we could easily revert [this](https://github.com/trolando/sylvan/commit/117984736675dcce77461754815d8f1e37068f15) change (which downloads lace instead of shipping it) in our sylvan version.
We then also have to revert the new lace stuff in the Storm cmake.

By the way: Debian 11 was released August 14th, 2021

trolando:
Somewhat related to this discussion is the design of Lace. Triggered by some comments of Jaco, I investigated whether I could actually significantly update the design of Lace and what this would cost in terms of performance.

Please see https://github.com/trolando/lace/pull/10.

The advantage would be code that is much more readable especially for novices. The disadvantage is a potential performance hit, although in the case of Sylvan I suspect that there is no performance hit as BDD operations are memory intensive.

tquatmann:
image

Here are some benchmarking results (old version on the x axis, new version on the y-axis).
There still seems to be a noticeable performance drop. Guess we have to optimize our lace usage next.

tquatmann:
Dear Diary,
today, I tested running the `DdJaniModelBuilder` as a LACE Task which (on the three instances I tested) resulted in a Speed-up of 20-50%.
This makes me confident that this is our fix: run all the BDD heavy code in LACE.
On the one side, we don't want to run all of Storm in LACE (due to python and other multi-threaded code that doesn't use lace). On the other side, I'm not in favour of doing this too fine grained (mostly because that would be a decent amount of work and I don't want to delay this longer than necessary).

The next step would be to think about the most convenient way to run templated C++ functions as LACE task. Then we need to apply this when invoking

- [ ] Symbolic model building (PRISM/JANI)
- [ ] Symbolic Bisimulation
- [ ] Game-Based Abstraction
- [ ] Model Checking with hybrid/dd engine (In the hybrid engine this currently means that the sparse parts are run in LACE which is not ideal; we have the same problem with the current (old) design)
- [ ] Storm-DFT when it is using BDDs for the static parts
- [ ] ???

Long-term, we should combine this with suspending/resuming LACE but this requires some care. Every code path that leads to BDD code then needs to awake LACE before. Here it is easy to miss some use cases, especially when Storm is used as a library.

tquatmann:
I came up with the following solution: We can now run code that uses many DD operations through an `execute` method inside the `DdManager`. Specifically, instead of code like this

```c++
bar = foo(arg1, arg2);
```
we can now pass things as a lambda function of type `void()` to the `DdManager::execute` like this

```c++
manager->execute([&]() { bar = foo(arg1,arg2); };
```

So far, I added this in the DD-based PRISM and Jani model builders as well as some critical API methods. Then, I redid the experiments from above and was happy to see that the comparison between the old sylvan and the new sylvan versions look much less scary:

image

tquatmann:
It remains to implement these for game based abstraction and the DFT stuff. I'll do that soon and then I think we can finally merge this.

There are likely entry points that I missed (especially through stormpy). Right now, if BDD code is not run through the `execute` method this only means that code will run *a bit* more slowly, but imo that's tolerable and can get fixed over time.

Long-term I'd like to suspend the LACE workers whenever we are outside of a `DdManager::execute` scope. Right now, the LACE workers do busy waiting all the time (whenever a DdManager exists). So non-DD code runs concurrently to waiting workers.

dbasgoeze:
> It remains to implement these for game based abstraction and the DFT stuff.

If you need help in any way with the DFT stuff just drop me an email or something. I am available and still interested to help.

tquatmann:
> If you need help in any way with the DFT stuff just drop me an email or something. I am available and still interested to help.

Thanks @dbasgoeze! Very much appreciated! Is there a "nice" entry point where we can run things as a LACE task?

dbasgoeze:
Hmmn, first of all the SFT analysis uses Sylvan more or less directly not through the manager class that sylvan proper uses but through SylvanBddManager that only uses an InternalDdManager to initialize Sylvan. So one would need to forward the execute method you wrote.

Iirc most BDD manipulating is started in the SFTBDDChecker class. Unfortunatly there is some work done in the constructor  so I don't know how great this will translate into the execute pattern. In general I think it would work to implement all public functions of that class as lace tasks.

Then there is SFTBDDPropertyFormulaAdapter which probably also needs to be handled separately.

---

BTW is there a good reason to use a std::function? My last info was that this is slow as std::function is statefull, wouldn't a templated callable be better here? I.e. something like

```
template<typename F>
void execute(F f) const;
```

That would need to be implemented inline but that would also mean it can be inlined in other CUs. Tho I don't know how much we expect execute function calls and if this is premature optimization...

tquatmann:
I think forwarding the execute method with the `SylvanBddManager` will be more or less straightforward. We're already doing this with the "normal" `DdManager` right now.

I had a quick look to the code and I found the `analyzeDFTBdd` method [here](https://github.com/moves-rwth/storm/blob/master/src/storm-dft/api/storm-dft.cpp). Couldn't we have one execute block here?
I understand that this is not the best solution and probably not very future proof. I'm also in favour of having a more fine-grained use of `execute`–also for the other model checking engines.
However, right now I think we should come up with something *acceptable* so that this already big PR can be merged. From there, we can separate concerns a bit and come up with more future-proof solutions for the various usages of DDs.

And regarding `std::function`s overhead: My hope is that it will never be significant, since there usually should be some non-trivial code inside `execute` that will make the overhead neglectable.
The reason I used `std::function` was that I had my doubts that I can use templated code when defining the LACE task. More specifically, I didn't know what else to put [here](https://github.com/moves-rwth/storm/blob/cb0db3c9c6ba30a4fc18e2104fe52a01c3a59036/src/storm/storage/dd/sylvan/InternalSylvanDdManager.cpp#L37):

```c++
VOID_TASK_2(execute_sylvan, std::function<void()> const*, f, std::exception_ptr*, e) { .. }
```

If you have a suggestion, I'm happy to get rid of `std::function` :)

dbasgoeze:
Ok, didn't know you wanted it that coarse. Yes, then the `analyzeDFTBdd` method should work fine for now.

Regarding `std::function` the common solutions for that would be to build your own container class that is faster than `std::function` by omitting the dynamic binding behavior of it, but I think that it is not worth the work here if profiling does not say otherwise.

.github/workflows/buildtest.yml
```diff
@@ -60,7 +60,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        distro: ["debian-10", "debian-11", "ubuntu-18.04", "ubuntu-20.04"]
+        distro: ["debian-10", "debian-11", "ubuntu-20.04"]
```
sjunges, line 63: I would commit this separately :-)

sjunges:
Many thanks for the hard work. I only left a minor request.

.github/workflows/buildtest.yml
sjunges: https://github.com/moves-rwth/storm/commit/6a26fb7115cfce8dd1ac3388987e6207b185bd9c

tquatmann:
The used LACE version was already outdated, so I've updated it and made it a bit more explicit in the documentation that LACE needs to be updated separately.
Things are still working fine for me, so IMO we can merge!

sjunges:
Yes, but we cannot rebase it due to conflicts?

Issue #276, opened 2022-08-10
Avoid std::cout and std::endl


volkm:
Currently, some parts of the code are still using `std::cout`. We should revise them to use our custom macros such as `STORM_LOG_INFO` or `STORM_PRINT_AND_LOG`.

In addition, we opt to use `\n` instead of `std::endl`, see PR #178.
The files `GradientDescentInstantiationSearcher.cpp` and `SparseDerivativeInstantiationModelChecker.cpp` should be adapted accordingly.

Shurtu-gal:
Hello,
I am Ashish Padhy (Sophomore at Bachelor's) and am kind of beginning my open source journey.
So can you please clarify how to choose the macro to replace `std::cout` .
Thanks for your help.

tquatmann:
PR #277 addressed the two files you mentioned. Someone should check for other misplaces occurences of `std::cout`

@Shurtu-gal the macros allow us to control the tool output on a global level. For example, we could easily implement a compile-time option that disables all output.

Legoeggolas:
If there are any other source files that require similar enhancements, I'd be very happy to help with them as well :D This issue helped me find a good example of macro use in a mature code base and I am grateful for that experience

Shurtu-gal:
@tquatmann Thanks for the heads up.
Would be happy to contribute to other issues.

PR #277, opened 2022-08-15
Replace std::cout with STORM_PRINT family of macros

Legoeggolas:
On #276 

- Revise GradientDescentInstantiationSearcher.cpp to use the macros
- Revise SparseDerivativeInstantiationModelChecker.cpp to use the macros
- Use "\n" instead of std::endl


tquatmann:
Thanks for this contribution! LGTM!

PR #278, opened 2022-08-23
Improved DFT modularization

volkm:
- added dedicated data structure for `DftModule`
- independent modules are computed with `DftModularizer`
- module finding algorithm was improved to better account for dependencies/restrictors
- improved model checking via modules which now first replaces each dynamic module with a BE and then builds the resulting static fault tree only once
- also added `DftValidator` to check whether a given Dft is well-formed and conventional


sjunges:
LGTM

Issue #279, opened 2022-08-24
Correct boost version to use


modass:
I am trying to compile storm from source and am using the mentioned CArL branch. The shipped boost version that comes with it seems to be 1.64, which is also detected by the storm build system and should be okay (the dependencies-list mentions > 1.61).

When compiling, I get a wall of information (the typical boost::spirit style of errors) on a spirit issue in the PrismParser, more specifically in SpiritParserDefinitions.h, which seems to boil down to: 

```
carl/cmake-build-release/resources/include/boost/spirit/home/qi/detail/alternative_function.hpp:30:46: error: no type named ‘types’ in ‘boost::spirit::qi::detail::find_substitute<boost::optional<boost::variant<storm::prism::Constant, storm::prism::Formula, storm::prism::Module, storm::prism::RewardModel, storm::prism::Label, storm::prism::ObservationLabel, storm::prism::Player> >, storm::prism::Formula>::variant_type’ {aka ‘class boost::optional<boost::variant<storm::prism::Constant, storm::prism::Formula, storm::prism::Module, storm::prism::RewardModel, storm::prism::Label, storm::prism::ObservationLabel, storm::prism::Player> >’}
   30 |         typedef typename variant_type::types types;
```

I am convinced this is not an error but rather an issue with the boost version - could you give me a hint on what versions you are using? Can I take a more recent version of boost, or will I run into issues there?

volkm:
I think you should indeed use a newer Boost version. CMake of Storm currently requires Boost 1.65 (see [here](https://github.com/moves-rwth/storm/blob/master/resources/3rdparty/CMakeLists.txt#L83)).
We will need to adapt the documentation accordingly.

I am currently using 1.79 without problems.

modass:
Amazing, thanks a lot!

PR #280, opened 2022-08-24
Updated documentation on Boost dependency

volkm:
Raised in #279.
I opted to remove the part on manually building Boost, because all relevant distros should provide a recent enough Boost version.


tquatmann:
LGTM

PR #283, opened 2022-09-01
Additional template instantiations for `storm::storage::BitVectorHashMap`

ifndefJOSH:
Hey,

So our project, STAMINA, requires an instantiation of a `BitVectorHashMap<uint8_t, Murmur3BitVectorHash<uint32_t>>` for its multithreading protocol. The linker won't allow me to instantiate it in STAMINA, so I must instantiate it in STORM itself. I was hoping that this small change could be added to upstream STORM code, and I have a couple of ideas for the things STORM could be able to use it for:

- If STORM ever implements multithreading in explicit state exploration (I'm not sure how well STAMINA's multithreading protocol would work outside STAMINA, but potentially), it could be used for that.
- Any sort of arbitrary "grouping" of states by state value--e.g., stages of "error state".
- When the state index is guaranteed to be lower than 256 (i.e., in small models)

The changes are all in `src/storm/storage/BitVectorHashMap.cpp` and adds only two lines of code:
```cpp
template class BitVectorHashMap<uint8_t, Murmur3BitVectorHash<uint64_t>>;
template class BitVectorHashMap<uint8_t, Murmur3BitVectorHash<uint32_t>>;
```
Thanks!


src/storm/storage/BitVectorHashMap.cpp
```diff
@@ -190,5 +190,14 @@ void BitVectorHashMap<ValueType, Hash>::remap(std::function<ValueType(ValueType
...
 template class BitVectorHashMap<uint64_t>;
 template class BitVectorHashMap<uint32_t>;
+// These instantiations allow you to "group" states in a BitVectorHashMap. I.e.,
+// if you want to look at a state and know what "group" it is in (with groups
+// controlled by an 8 bit group index) you can instantiate a BitVectorHashMap<uint8_t>
+// which looks up your state and returns a group inded
+//
+// For example, the STAMINA project (built on STORM) uses this for thread indecies
+// in its multithreading implementation.
+template class BitVectorHashMap<uint8_t, Murmur3BitVectorHash<uint32_t>;
```
volkm, line 200: I think a closing `>` is missing?

ifndefJOSH:
Oops--fixed in [5f8f169](https://github.com/moves-rwth/storm/pull/283/commits/5f8f169bf833b19d558d1932bd6474749bb5e6c0)

sjunges:
Hi Josh, 

I think we can support this. 

More generally, we are right now reducing compile times with a lots of templates being only instantiated with the types we need. That has some downsides for using Storm as an extensible library. 

I saw that C++20 has some nice additions in terms of modules that could allow us to expose more templates while keeping compilation times down, but going to C++20 has some consequences. 

- First, right now, we cannot compile storm with C++20 (see https://github.com/moves-rwth/storm/issues/286). 
- Furthermore, modules require GCC11. We try to keep compiler support for the last two versions of Ubuntu LTS as a general rule, thus, until 2024, we cannot enforce GCC11 (that does not mean that we could not add something optional that requires C++20)

PR #292, opened 2022-09-09
remove jit from current code base

sjunges:
Based on discussions, JIT is currently almost unused and not maintained and we do not see how this changes in the foreseeable future. 

If we want to revive it, we can take it from the git history.


tquatmann (approved):
LGTM

Issue #258, opened 2022-07-28
Keep float support?


sjunges:
Some classes such as Matrices and Schedulers still support floats. Is this something we want to keep? (I think the original motivation was GPUs, but I would like to clean the code and get rid of the floats as they are not used from the outside at all?

tquatmann:
I second this. As far as I know, there is no need to have `float`s anymore.

PR #293, opened 2022-09-09
always use doubles instead of floats

sjunges:
Fixes https://github.com/moves-rwth/storm/issues/258

### DIFF
--- a/src/storm/models/sparse/Model.cpp
+++ b/src/storm/models/sparse/Model.cpp
@@ -600,11 +600,6 @@ void Model<double, storm::models::sparse::StandardRewardModel<storm::Interval>>:
     STORM_LOG_THROW(false, storm::exceptions::NotImplementedException, "Json export not implemented for this model type.");
 }
 
-template<>
-void Model<float>::writeJsonToStream(std::ostream& outStream) const {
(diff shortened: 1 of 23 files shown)


sjunges:
@volkm I changed some checks from EQ_FLOAT to EQ_DOUBLE as we are dealing with doubles but these seem to raise errors because the reference values are given less precisely. 

I think I would revert those changes for now as the main point of this PR is a different one.

(@tquatmann  Generally, do we use some precision, should we use our own macro?)

volkm:
Thanks for starting to clean up.
I think other parts of the tests use different checks such as EXPECT_NEAR which also allow to give a precision.
You can revert the failing changes for now and I will create a separate PR using better comparisons.

Am 10. September 2022 11:15:10 MESZ schrieb Sebastian Junges ***@***.***>:
***@***.*** I changed some checks from EQ_FLOAT to EQ_DOUBLE as we are dealing with doubles but these seem to raise errors because the reference values are given less precisely. 
>
>I think I would revert those changes for now as the main point of this PR is a different one.
>
***@***.***  Generally, do we use some precision, should we use our own macro?)

volkm (approved):
Looks good

PR #299, opened 2022-09-16
Support 'relevant' in DFT Json format

volkm:
Relevant events in DFT can be imported/exported from Json format.


sjunges:
I would consider adding a debug output about the relevant events for consistency. 

Other than that: LGTM

PR #300, opened 2022-09-16
Added information for developers

volkm:
Started on creating some general information for developers. The idea is to (1) help new developers get faster into the code, and (2) make some development policies we follow explicit.


sjunges:
Nice. We should also mention this in the readme.

volkm:
I took the opportunity to also update the README a bit.

sjunges:
Looks very good to me! Many thanks

PR #304, opened 2022-09-29
Gurobi: Fixed Typo in preprocessor directive

tquatmann:
Small fix that prevents potential memory leaks.


sjunges:
LGTM, thanks!

PR #305, opened 2022-10-02
Storm-dft: treat single BEs as modules

volkm:
Fixed an issue with the DFT modularisation pointed out by @AhmadZafarDGB.


sjunges:
LGTM

PR #307, opened 2022-10-05
Fix Formula Parser

### LINKED ISSUE #195: Formula Parsing Bug
Somehow parsing `Pmax=? [ F (1=1)]` works but  `Pmax=? [ F ((1)=1)]` yields a 

```
WrongFormatException: Expected expression 1 to be of boolean type.
```

tquatmann:
## Fixed parsing expression formulas (fixes #195)

The issue was that if we see an opening bracket `(` we expected that there is a *formula* inside the brackets. This is not the case for formulas like `(1)=2` as 1 is not a formula.

## Fixed backtracking mode in expression parser

With backtracking enabled, the expression parser is not supposed to throw an exception  on a formula like `1 & true`.
However, when such an error is encountered the expression parser currently overwrites the result parsed so far with a dummy expression which results in a BAD_ACCESS signal. This commit fixes the issue.

### DIFF
--- a/src/storm-parsers/parser/ExpressionParser.cpp
+++ b/src/storm-parsers/parser/ExpressionParser.cpp
@@ -96,10 +96,11 @@ ExpressionParser::ExpressionParser(storm::expressions::ExpressionManager const& 
     roundExpression.name("round expression");
 
     if (allowBacktracking) {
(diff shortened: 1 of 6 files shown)


sjunges:
LGTM; but I am not a parser expert :-)

tquatmann:
I'll do some more tests on a larger set of formulas (i.e. QVBS) before we merge. The parser is very delicate ;)

tquatmann:
I've let this run on an extensive set of formulas (QVBS plus some others) without issues so I think this is ready for merge.

PR #310, opened 2022-10-21
JANI to JSON export: Fixed unary minus.

tquatmann:
The jani-to-json export produced unary minus expressions, which are not part of the [jani-spec](https://docs.google.com/document/d/1BDQIzPBtscxJFFlDUEPIo8ivKHgXT8_X6hz5quq7jK0/).
This PR changes the exports of expressions like `-x` to

```json
{ "left": 0, "op": "-",  "right": "x" }
```


sjunges:
LGTM.

(There was a time where we had a flag to support standard jani vs our own jani extensions but I dont think that difference should be made here)

PR #312, opened 2022-10-24
Direct elimination of zero entries in SparseMatrix transformations

### LINKED ISSUE #309: Unexpected behaviour when applying scheduler resulting from makeRowDirac
The `makeRowDirac` method of SparseMatrix  leads to unexpected behaviour when applying a scheduler on an MDP based on the transformed matrix while dropping unreachable states.
In particular, consider the following scenario: I have an MDP on which I computed a scheduler. In this MDP there is a choice (selected by the scheduler) with two successor states where I want to drop the transition to one of those states. For that I make the row into a Dirac distribution and generate a new MDP using the transformed matrix. When applying the scheduler on this new MDP and dropping unreachable states, the following error occurs:
`ERROR (SubsystemBuilder.cpp:109): Expected that in each state, at least one action is selected. Got a deadlock state instead. (violated at STATE)`
where `STATE` is the index of the state the row made Dirac belongs to.

After some testing, I think the error occurs as making the row Dirac does not delete the other entries in the row, but only sets them to zero. Thus, a state may become unreachable and should be dropped, but the matrix entry still exists. In the `SubsystemBuilder`, we iterate over all entries of the matrix and determine 'successors' using the column, so there is a mismatch as a state is unreachable, but still referenced in the matrix.

A possible fix for this issue can be achieved by changing the condition in https://github.com/moves-rwth/storm/blob/77ff8ea757e72cb2c3224b12d4fe5d842c849a82/src/storm/transformer/SubsystemBuilder.cpp#L93
to `!subsystemStates.get(entry.getColumn()) && entry.getValue() != storm::utility::zero<ValueType>()`, i.e. expicitly checking for 'zero entries'.
What do you think?

AlexBork:
This PR adds a Boolean flag to all transformations in SparseMatrix that may lead to zero entries. If the flag is set, entries that are zero are deleted from the sparse matrix representation instead of just being set to zero. As in many cases, the performance impact of this operation is not desirable, the default behavior of not dropping zero entries is preserved.
This also directly exposes the default behavior and helps to prevent implicit expectations about the form of the matrix after the transformation has been applied (see also #309).
This pull request resolves #309.

### DIFF
--- a/src/storm/storage/SparseMatrix.cpp
+++ b/src/storm/storage/SparseMatrix.cpp
@@ -863,29 +863,39 @@ storm::storage::BitVector SparseMatrix<ValueType>::getRowGroupFilter(storm::stor
 }
 
 template<typename ValueType>
-void SparseMatrix<ValueType>::makeRowsAbsorbing(storm::storage::BitVector const& rows) {
+void SparseMatrix<ValueType>::makeRowsAbsorbing(storm::storage::BitVector const& rows, bool dropZeroEntries) {
+    // First transform ALL rows without dropping zero entries, then drop zero entries once
+    // This prevents iteration over the whole matrix every time an entry is set to zero.
     for (auto row : rows) {
-        makeRowDirac(row, row);
+        makeRowDirac(row, row, false);
+    }
+    if (dropZeroEntries) {
+        this->dropZeroEntries();
     }
 }
 
 template<typename ValueType>
-void SparseMatrix<ValueType>::makeRowGroupsAbsorbing(storm::storage::BitVector const& rowGroupConstraint) {
+void SparseMatrix<ValueType>::makeRowGroupsAbsorbing(storm::storage::BitVector const& rowGroupConstraint, bool dropZeroEntries) {
+    // First transform ALL rows without dropping zero entries, then drop zero entries once.
+    // This prevents iteration over the whole matrix every time an entry is set to zero.
     if (!this->hasTrivialRowGrouping()) {
         for (auto rowGroup : rowGroupConstraint) {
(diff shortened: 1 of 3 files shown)


sjunges:
I would not mention the commit in .git-blame-ignore-revs and squash all commits instead of merging them when we integrate this PR.. sorry :-)

src/storm/storage/SparseMatrix.cpp
```diff
@@ -863,29 +863,35 @@ storm::storage::BitVector SparseMatrix<ValueType>::getRowGroupFilter(storm::stor
 }
 
 template<typename ValueType>
-void SparseMatrix<ValueType>::makeRowsAbsorbing(storm::storage::BitVector const& rows) {
+void SparseMatrix<ValueType>::makeRowsAbsorbing(storm::storage::BitVector const& rows, bool dropZeroEntries) {
     for (auto row : rows) {
-        makeRowDirac(row, row);
+        makeRowDirac(row, row, false);
+    }
+    if (dropZeroEntries) {
```
sjunges, line 870: Maybe add a comment why we first make all rows dirac and only then clear the zero entries

src/storm/storage/SparseMatrix.h
```diff
@@ -659,24 +659,30 @@ class SparseMatrix {
      * This function makes the given rows absorbing.
      *
      * @param rows A bit vector indicating which rows are to be made absorbing.
+     * @param dropZeroEntries if true, zero entries resulting from the transformation are dropped from the matrix after the transformation.
+     * This action may impact the performance significantly.
```
tquatmann, line 663: This is a bit vague, maybe something like
```suggestion
     * Dropping zero entries takes time linear in the number of matrix entries.
```

AlexBork:
I have added the suggestions and squashed the commits.

tquatmann:
Thanks a lot! LGTM

Issue #259, opened 2022-07-28
Consistency wrt int types


sjunges:
We still seem to use the fast64 style integers, which I think we abandoned quite a while ago. Would it be good to do a pass over the code and remove these? Would it also make sense to add these rules to the CI?

PR #314, opened 2022-11-03
SparseMatrix: ranges for RowGroupIndices and no uint_fast64_t

tquatmann:
This PR adds an overload of `SparseMatrix<T>::getRowGroupIndices` that returns an index range for a given row group.

Instead of

```c++
for (auto rowIndex = matrix.getRowGroupIndices()[group]; rowIndex < matrix.getRowGroupIndices()[group + 1]; ++rowIndex) {
    // stuff
}
```

we can now do this

```c++
for (auto rowIndex : matrix.getRowGroupIndices(group)) {
    // stuff
}
```

The PR also avoids using `uint_fast64_t` in the SparseMatrix file (see #259).


volkm:
LGTM

PR #316, opened 2022-11-17
Change `storm::storage::BitVector` to use `std::shared_ptr` for 64-bit buckets

ifndefJOSH:
This PR changes `uint64_t * buckets` in `storm::storage::BitVector` to use a `std::shared_ptr<uint64_t>`, which can be initialized for an array as follows:

```cpp
std::shared_ptr<uint64_t> newBuckets(new uint64_t[bucketSize], std::default_delete<uint64_t>());
buckets = newBuckets;
```
Additionally, pointer arithmetic can be used with the `T * std::shared_ptr<T> get()` function:

```
// Old
buckets + n;
// New
buckets.get() + n
```

The reason I'm proposing this change is because it is possible for two BitVectors to exist, which share the same buckets in memory, and for the destructor of one to get called, invalidating the memory of the other and causing a segmentation fault. To explain why, let me give some background; the `BitVector` has two copy constructors:

1. `BitVector(const BitVector & other)`, which creates a deep copy, i.e., copies the buckets.
2. `BitVector(BitVector && other)`, which creates a shallow copy, i.e., has the new BitVector share the same buckets as the old one.
If creating a copy of a `BitVector` the first copy constructor is invoked, but if creating a copy of a `BitVector &` or a `const BitVector &`, like the `NextStateGenerator`s produce, the second copy constructor is invoked. If the copy outlasts the original (and the original's destructor is called, deleting `buckets`), the copy is left with a pointer to invalid memory. This is not detectable in the program at runtime since the pointer is not nulled, just invalid. Consequently, this often causes a segfault.

This issue was discovered in STAMINA, since STAMINA maintains copies of state values periodically. Since we want both shallow and deep copy to exist, there should be a way if shallow copy is used, to make sure that multiple copies of the same BitVector will never have invalid memory if one of them is destroyed.

### DIFF
--- a/src/storm/storage/BitVector.cpp
+++ b/src/storm/storage/BitVector.cpp
@@ -222,7 +222,7 @@ void BitVector::resize(uint_fast64_t newLength, bool init) {
         }
 
         if (newBucketCount > this->bucketCount()) {
-			std::shared_ptr<uint64_t> newBuckets(new uint64_t[bucketCount], std::default_delete<uint64_t[]>());
+			std::shared_ptr<uint64_t> newBuckets(new uint64_t[newBucketCount], std::default_delete<uint64_t[]>());
             std::copy_n(buckets, this->bucketCount(), newBuckets);
             if (init) {
                 if (this->bucketCount() > 0) {
--- a/src/storm/storage/BitVector.cpp
+++ b/src/storm/storage/BitVector.cpp
@@ -77,18 +77,18 @@ BitVector::BitVector(uint_fast64_t length, bool init) : bitCount(length), bucket
 
     // Initialize the storage with the required values.
     if (init) {
-		std::shared_ptr<uint64_t> newBuckets(new uint64_t[bucketCount], std::default_delete<uint64_t[]>());
-		buckets = newBuckets;
+        std::shared_ptr<uint64_t> newBuckets(new uint64_t[bucketCount], std::default_delete<uint64_t[]>());
+        buckets = newBuckets;
         std::fill_n(buckets.get(), bucketCount, -1ull);
         truncateLastBucket();
     } else {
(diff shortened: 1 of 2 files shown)


ifndefJOSH:
Converting to draft as there appears to be some unknown behavior now in model checking

tquatmann:
Thanks for looking into this! I'm not sure if I correctly understand the motivation, though.

> If creating a copy of a BitVector the first copy constructor is invoked, but if creating a copy of a BitVector & or a const BitVector &, like the NextStateGenerators produce, the second copy constructor is invoked.

I don't think this is the case, or maybe I just misunderstand? In the snippet below, both invocations use the first one.

```c++
#include <iostream>

class BitVector {
    public:
    BitVector() = default;
    BitVector(BitVector const& other) {
        std::cout << "First\n";
    }
    
    BitVector(BitVector&& other) {
        std::cout << "Second\n";
    }

};

int main()
{
    BitVector b;
    BitVector const& b_const_ref = b;
    BitVector & b_ref = b;
    
    BitVector b1(b_const_ref); // prints "First"
    BitVector b2(b_ref); // prints "Second"

    return 0;
}
```

[you can run the code here](https://onlinegdb.com/sHwKjn9Rh)

The constructor `BitVector(BitVector&& other)` is not meant to be a *copy* of `other`, but a *move* (think of copying a file vs. moving a file on disk). It actually sets `others.buckets = nullptr`.
As far as I know, it *should* be impossible to have two different BitVectors that share the same `buckets`.

ifndefJOSH:
Ah @tquatmann, you are correct. I misunderstood the issue. There is however something that is causing a segfault in `BitVector::~BitVector()` when attempting to `delete[] buckets`. I don't think it occurs in a typical run of Storm, but it happens in STAMINA. I will look into it more and provide more information, and if I can figure one out, a fix.

As for this PR, if you guys ever decide want to migrate `BitVector` to `std::shared_ptr`, now you have a starting point haha!

Thanks for clearing up my confusion!

PR #318, opened 2022-11-23
Fix total reward computation

tquatmann:
We compute `Rmin=? [C]` formulas by reducing them to standard reachability rewards `Rmin=? [F "sink"]`. When minimizing, this transformation requires to eliminate end components. This PR fixes a bug in that transformation.


volkm:
LGTM.
One small thing: did you mean maximizing instead of minimizing?

Issue #274, opened 2022-08-03
Compiling on Apple Silicon


tquatmann:
I propose to document the current state of compiling on apple silicon in this issue.

Once PR #273 is merged, Storm compiles natively on apple M1. However, some tweaks are likely necessary. Here is what worked for me:

* Install dependencies (again) using the *arm* version of homebrew. Skip `cln` and `ginac` as these are not supported by homebrew
* Make sure that you use an arm compiled `cmake` binary (check with `where cmake` and `file path/to/cmake`)
* Build carls `master14` branch without cln and ginac (`make lib_carl` suffices)
* When configuring storm, set cmake option `-DSTORM_PORTABLE=ON`

> Hint: when building x86 versions, I now set `-DSTORM_FORCE_SHIPPED_CARL=ON` as we can't use the arm installation

## TODO list for official support:

- [ ] merge PR #273 
- [ ] make installation of carl and storm without cln and ginac more straightforward (ideally, these dependencies are disabled automatically if compiling on M1 is detected).
- [ ] I had problems with setting `-DSTORM_FORCE_SHIPPED_CARL=ON` when compiling for arm: carl did not compile due to some eigen related issue.
- [ ] Some includes are taken from the x86 installation. For example, I ran into a known issue related to boost that should have already been fixed in my boost installation. The "solution" was  to upgrade my x86 boost installation.
- [ ] clang option `-march=native` does not seem to work yet which is why `-DSTORM_PORTABLE` was required for me. Also, the jit engine doesn't work for that reason, which also lets the tests fail
- [ ] The Homebrew formula(s) for Storm and Carl need to be updated as well
- [ ] Adapt Website documentation

sjunges:
Hi Tim,

great progress. Could you add an item regarding stormpy on the TODO list. At least it would be good if we have clarity on whether this will work when writing the documentation. 

Do we already detect M1/ARM64 architectures in CMake?

sjunges:
Regarding CLN: https://www.ginac.de/pipermail/cln-list/2023-March/000803.html, there seems to be some progress

tquatmann:
I'm compiling on Apple Silicon just fine and I can no longer reproduce the issues I had before when using a shipped carl version. I guess that problem solved itself.

The test `SparseMaPcaaMultiObjectiveModelCheckerTest.serverRationalNumbers` fails due to an issue in Eigen `v3.3.9`. I worked around this in PR #349

sjunges:
[...]

PR #323, opened 2023-01-18
Detection of Apple Silicon

AlexBork:
This PR changes the CMake files such that it is detected if Storm is being built for Apple Silicon.
In more detail, the additions are:
- detection of Apple Silicon
- warning that Storm does not yet compile natively and that Rosetta 2 should be used
- disabling of CLN and GiNaC usage in both Storm and the shipped carl version
- disabling of '-march=native' compiler option for compilers that do not support it
- ~~enabling of '-mcpu=apple-m1' compiler option for AppleClang 13.0 and higher~~ (see comment below)

This is a step towards full native Apple Silicon support (see https://github.com/moves-rwth/storm/issues/274)


resources/3rdparty/CMakeLists.txt
```diff
@@ -322,11 +322,16 @@ else()
 endif()
 
 if (STORM_SHIPPED_CARL)
+    if(NOT APPLE_SILICON)
```
volkm, line 325: I find it a bit confusing that these variables are set in two places. Would it make sense to set it to `ON` by default and only disable it in the root `CMakeLists.txt`?

volkm:
Other than the small comment, it looks good to me.

CMakeLists.txt
```diff
@@ -182,6 +182,20 @@ else()
 ENDIF()
 message(STATUS "Storm - Detected operating system ${OPERATING_SYSTEM}.")
 
+# Detect Apple Silicon and adjust settings
+if(APPLE AND ${CMAKE_SYSTEM_PROCESSOR} MATCHES arm64)
+	message(STATUS "Storm - Detected that target system uses Apple Silicon.")
+	message(WARNING "Storm currently does not compile natively on Apple Silicon! Please consider building Storm using Rosetta 2. FOr more information visit https://www.stormchecker.org/documentation/obtain-storm/apple-silicon.html")
```
tquatmann, line 188:
```suggestion
	message(WARNING "Storm currently does not compile natively on Apple Silicon! Please consider building Storm using Rosetta 2. For more information visit https://www.stormchecker.org/documentation/obtain-storm/apple-silicon.html")
```

CMakeLists.txt
```diff
@@ -329,8 +343,12 @@ endif()
 
 # In release mode, we turn on even more optimizations if we do not have to provide a portable binary.
 if (NOT STORM_PORTABLE)
-	set (CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -march=native")
-endif ()
+	if(NOT APPLE_SILICON OR (STORM_COMPILER_CLANG AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 15.0))
+		set (CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -march=native")
+	elseif(STORM_COMPILER_APPLECLANG AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 13.0)
+		set (CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -mcpu=apple-m1")
```
tquatmann, line 349: Does this break if people have other ARM processors, like Apple M2?

AlexBork: I don't think it does for M2. Some posts from before the `-mcpu` definition for M1 existed already suggested using the A-series chip (iPhone/iPad) defintion, so using M1 here shouldn't break anything. Unfortunately, I don't have a machine to test it on :)
Other ARM processors might have a problem, depending on whether `-march=native` is supported for the platform. I would leave it like this for now, if a problem occurs, we know where to look for it.

resources/3rdparty/CMakeLists.txt
AlexBork: Yes, that makes sense. I will change that.

AlexBork:
After consulting with @tquatmann, I removed the `-mcpu=apple-m1` flags as we are not sure about its behavior, in particular with M2 CPUs.

tquatmann:
I'm happy with this now :) Thanks for the effort!

volkm:
I am also happy. Thanks for adapting the CMake files.

PR #329, opened 2023-02-15
Fixed transitionRewardMatrix transformation in SubsystemBuilder

AlexBork:
This PR contains a small fix for the SubsystemBuilder. Previously, if a non-deterministic model had a transition reward matrix and a subsystem was created where the rowgrouping is trivial, this was not carried over to the transition reward matrix, leading to failed assertions.


volkm:
LGTM.

PR #332, opened 2023-03-06
List of tools using Storm

volkm:
I started a list of tools using Storm as a backend.

If you prefer a different location for this list, let me know. One possibility could be to split the *publications* page into two pages: one for publications on Storm and new features, and one page on publications and tools using Storm as a backend.


sjunges:
What makes a tool a tool? I have some papers where I would say "use storm as a backend" is more adequate than "a tool" but they often come with a prototype that you could consider a tool...

volkm:
My current (informal) criteria for tool is something that lives longer than just one paper. The plan was to have a list with "tools" build upon Storm which is a bit more "hand-on" than a list of papers. But maybe we should just make sure that the list of papers also has the relevant tags for website and code.

tquatmann:
Sebastian has a good point, but I think we can decide what a tool is on a case-by-case basis if there is any doubt.
I'm in favour of merging this.

volkm:
I updated the list of tools, thanks for your feedback.

tquatmann (approved):

PR #334, opened 2023-03-08
Permissive schedulers in a separate lib

sjunges:
In order to reduce main linking time and isolate core features from additional features, 
I have moved the computation of permissive schedulers as presented in a TACAS 2016 paper into an own lib.


sjunges:
I will squash :-)

PR #339, opened 2023-03-23
more adequate includes and forward refs

sjunges:
Accellerates some parts of the compilation process, but mostly  is in preparation of some changes in carl


tquatmann:
LGTM!

PR #342, opened 2023-03-28
add STORM_COMPILE_WITH_COMPILATION_PROFILING

sjunges:
This adds an option to switch on profiling information that eases profiling the compilation process.
Right now, the only flag that is set is explained e.g. here:
 https://www.snsystems.com/technology/tech-blog/clang-time-trace-feature


volkm:
Good idea. LGTM

PR #343, opened 2023-03-28
various forward declarations

sjunges:
Various forward declarations to speed up compiling and to prevent recompilation. Also updates the carl-repo (but should not break support for old repo right now).

I have not yet updated the github actions because I want this to run through with the old version of carl.

PR #344, opened 2023-04-03
fix for mistake in #343

sjunges:
fix for #343 …only was enabled by recent update to carl 14.24


volkm:
If the tests run through, I think we can merge it. Looks good from my side.

sjunges:
Thanks @volkm  for helping out

PR #345, opened 2023-04-12
Storm-dft: check if element is relevant

volkm:
Fix: the `DFTJsonParser` set elements as relevant even if the entry was set to `false`


tquatmann (approved):
LGTM

PR #350, opened 2023-04-21
re-implementation of value iteration-based approaches

tquatmann:
A re-implementation of 

* Value Iteration (VI)
* Optimistic Value Iteration (OVI)
* Sound Value Iteration (SVI)
* Interval Iteration (II)
* Rational Search (RS)

This provides a uniform way to implement Bellman-like operators using the new `ValueIterationOperator` class. Thanks to C++ templating, different *backends* can be provided to the operator to implement the different kinds of value iteration.

The new code removes quite a few code duplications (e.g. there was both DTMC + MDP code for VI,II, and RS; the OVI+SVI helper classes both had code doing the job of the new `ValueIterationOperator`). The hope is that this is now easier to maintain and extend towards e.g. other VI variants.

The new implementation improves performance, in particular for SVI, OVI, and II. Here are a few plots comparing runtime between the new/old versions of VI/OVI

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ Version 1.7.x
 -------------
 
 ## Version 1.7.1
-- Revised implementation of value iteration and its variants
+- Revised implementation of value iteration algorithms and its variants, fixing a bug in the optimistic value iteration heuristic.
 - Experimental support for compiling on Apple Silicon
 - Added SoPlex as a possible LP solver
 - Upgraded shipped version of sylvan
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Version 1.7.x
 -------------
 
 ## Version 1.7.1
+- Revised implementation of value iteration and its variants
 - Experimental support for compiling on Apple Silicon
 - Added SoPlex as a possible LP solver
 - Upgraded shipped version of sylvan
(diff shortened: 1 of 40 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Version 1.7.x
 -------------
 
 ## Version 1.7.1
+- Revised implementation of value iteration and its variants
```
sjunges, line 12: includes a bugfix, right?

src/storm/settings/modules/MinMaxEquationSolverSettings.cpp
```diff
@@ -64,8 +64,8 @@ MinMaxEquationSolverSettings::MinMaxEquationSolverSettings() : ModuleSettings(mo
                                          .build())
                         .build());
 
-    this->addOption(storm::settings::OptionBuilder(moduleName, intervalIterationSymmetricUpdatesOptionName, false,
-                                                   "If set, interval iteration performs an update on both, lower and upper bound in each iteration")
+    this->addOption(storm::settings::OptionBuilder(moduleName, forceUniqueSolutionRequirementOptionName, false,
+                                                   "If set, a MinMax solver always requires a unique solution for its input.")
```
sjunges, line 68: I find this a very vague description for a user setting. Can we provide some additional context when a user should set this flag?

sjunges:
Other than my comment above, LGTM, many thanks for the nice updates!

src/storm/settings/modules/MinMaxEquationSolverSettings.cpp
tquatmann: True, what about this:
`Enforces end component collapsing for MinMax equation systems so that their solution becomes unique. May simplify solving but causes some overhead` ?

PR #367, opened 2023-05-15
Fixed compilation if Z3 is not present

volkm:
In 99% of cases, z3 should be present anyway, but at least it should compile without it as well.

### DIFF
--- a/src/storm/utility/solver.cpp
+++ b/src/storm/utility/solver.cpp
@@ -1,14 +1,14 @@
 #include "storm/utility/solver.h"
 
-#include "storm/solver/GlpkLpSolver.h"
-#include "storm/solver/GurobiLpSolver.h"
-#include "storm/solver/SoplexLpSolver.h"
-#include "storm/solver/Z3LpSolver.h"
-
+#include "storm/exceptions/InvalidOperationException.h"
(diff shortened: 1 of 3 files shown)


src/storm/utility/solver.cpp
```diff
@@ -1,14 +1,14 @@
 #include "storm/utility/solver.h"
 
-#include "storm/solver/GlpkLpSolver.h"
-#include "storm/solver/GurobiLpSolver.h"
-#include "storm/solver/SoplexLpSolver.h"
-#include "storm/solver/Z3LpSolver.h"
-
+#include "storm/exceptions/InvalidOperationException.h"
```
tquatmann, line 3: Why this? I can not find any usage of `InvalidOperationException`.

tquatmann (approved):
Except for my little comment, this looks good.

src/storm/utility/solver.cpp
volkm: We are using it here: https://github.com/moves-rwth/storm/blob/master/src/storm/utility/solver.cpp#L134

tquatmann: Sorry, cmd+f apparently didn't work somehow... Thanks!

PR #368, opened 2023-05-15
Added `#include <cstdint>` to some files

ifndefJOSH:
Added `#include <cstdint>` to the following files:

- src/storm/logic/RewardAccumulation.h
- src/storm/modelchecker/prctl/helper/rewardbounded/EpochManager.h
- src/storm/utility/NumberTraits.h
- src/storm/storage/jani/traverser/InformationCollector.h
- src/storm/utility/string.h
- src/storm-cli-utilities/cli.h

I am not sure why, but the omission of this include caused errors on the latest versions of both GCC and Clang on Artix Linux.

Tested with the following C++ compilers
gcc (GCC) 13.1.1 20230429
clang version 15.0.7

CMake version:
cmake version 3.26.3

PR #371, opened 2023-05-16
CI tests with different compilers

volkm:
Adds compiler tests to the CI: building Storm with GCC and Clang on Archlinux. Using Archlinux as OS allows to use the most recent compiler versions.

Note that the tests will currently fail. These compile issues should be fixed by https://github.com/moves-rwth/storm/pull/368.


tquatmann:
LGTM. Thanks a lot for doing this!

volkm:
Merging as PR #368 has been integrated.

PR #374, opened 2023-05-20
More robust cudd compilation

tquatmann:
* Makes sure that necessary autotools binaries are found and that their location is passed to `autoreconf` and `make`
* fixes these warnings
```console
configure.ac:132: warning: The macro `AC_TRY_RUN' is obsolete.
configure.ac:132: You should run autoupdate.
./lib/autoconf/general.m4:2997: AC_TRY_RUN is expanded from...
lib/m4sugar/m4sh.m4:692: _AS_IF_ELSE is expanded from...
lib/m4sugar/m4sh.m4:699: AS_IF is expanded from...
./lib/autoconf/general.m4:2249: AC_CACHE_VAL is expanded from...
configure.ac:132: the top level
```
(I just ran `autoupdate` as suggested)

### DIFF
--- a/resources/3rdparty/cudd-3.0.0/configure.ac
+++ b/resources/3rdparty/cudd-3.0.0/configure.ac
@@ -2,7 +2,7 @@
 # Process this file with autoconf to produce a configure script.
 
 AC_PREREQ([2.69])
-AC_INIT([cudd], [3.0.0], [Fabio@Colorado.EDU])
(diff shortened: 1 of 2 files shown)


tquatmann:
Looks like I have to do `autoupdate` with an older version to make sure that thinks are compatible. I'll look into this.

volkm:
LGTM

PR #375, opened 2023-05-22
Add Storm-POMDP to auto-formatting

AlexBork:
This adds Storm-POMDP as a target for formatting.


volkm:
LGTM. After merging, we should trigger the auto-format CI action.

PR #377, opened 2023-05-22
Fix missing declaration

AlexBork:
After applying the formatting, for some reason WinningRegion.h is missing the declaration of storm::expressions::Expression which prevents compilation.
This fixes the issue. Thanks to @volkm for bringing this to my attention.


volkm:
LGTM.

I think the issues was triggered because after the formatting `WinningRegion.cpp` is included `WinningRegion.h` before `Expression.h`.

volkm (approved):

volkm:
Thanks for the fixes!

PR #381, opened 2023-05-31
Added storm-pomdp changes to CHANGELOG.md

AlexBork:
see title


volkm:
Looks good. Thanks for adding it.

(You can skip the CI if you want, because the change does not affect the compilation.)

PR #382, opened 2023-05-31
Moved parsing of explicit model files to storm-parsers

tquatmann:
There were some `#include "storm-parsers/..."` in files located in `src/storm/` which is not valid since `storm-parsers` depends on the core `storm` library (and not the other way round). I removed spurious includes and moved API methods for building models from explicit files (e.g. from .drn)  from `storm` to `storm-parsers`.


volkm:
LGTM.

PR #392, opened 2023-06-19
Dockerfile in CI

volkm:
- Updated `Dockerfile` to accept arguments
- Major revision of CI which now builds upon the existing Dockerfile. This should simplify the CI workflow.
- Added two Dockerfiles for CI:
  * `Dockerfile.archlinux` for building on the latest Archlinux version. This allows to test with the latest compiler and library versions.
  * `Dockerfile.carl_implicit` for implicitly building Carl-storm from within Storm

### DIFF
--- a/.github/workflows/Dockerfile.archlinux
+++ b/.github/workflows/Dockerfile.archlinux
@@ -1,0 +1,53 @@
+# Base Dockerfile for using Storm on Archlinux
+##############################################
+# The Docker image can be built by executing:
+# docker build -t yourusername/storm .
+
+FROM archlinux:latest
+MAINTAINER Matthias Volk <m.volk@utwente.nl>
+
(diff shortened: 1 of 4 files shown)


sjunges:
This looks good. I do not fully comprehend what you mean by building carl implicitly? Is this the fact that you then  do or do not use an image in which carl is already built?

volkm:
The "normal" Dockerfile builds carl-storm explicitly, i.e. clone, configure and compile carl before doing the same for Storm. This is needed such that for example pycarl can properly find the carl directory.
The "implicit" Dockerfile does not clone carl, but lets Storm do this implicitly as part of building the Storm resources.

The "normal" Dockerfile could also be based upon an already existing carl-storm image, but we are not doing it at the moment.

Issue #386, opened 2023-06-14
Increase minimal CMake version


volkm:
Currently, we require CMake version 3.8. I propose to increase this to e.g. version 3.15 or higher.
- Sylvan already requires 3.13
- Workarounds such as https://github.com/moves-rwth/storm/commit/bf735f0b0041558317ab6fad378d3a9ca3e9ae90 could be removed
- Getting the Boost version [becomes easier](https://cmake.org/cmake/help/latest/policy/CMP0093.html#policy:CMP0093)
- stormpy already requires 3.14
- All supported Ubuntu versions ship with [CMake >= 3.16](https://packages.ubuntu.com/search?keywords=cmake), only Debian 10 "Buster" still has [CMake 3.13](https://packages.debian.org/search?keywords=cmake), but `buster-backports` allow getting a newer version

volkm:
Closed with #393

PR #393, opened 2023-06-20
Increased CMake version to 3.15

volkm:
Addresses #386


tquatmann:
LGTM2! 

Potentially in another PR, we should add `debian-12` to the  workflow, see [debian releases](https://www.debian.org/releases/).
Also, the [supported operated systems on the website](https://www.stormchecker.org/documentation/obtain-storm/build.html) should be bumped to debian-11

volkm:
PR https://github.com/moves-rwth/storm/pull/392 will add support for `debian-12`. I will also update the documentation.

PR #397, opened 2023-06-29
Website: Debian 10 is not supported anymore

volkm:
Updated documentation after https://github.com/moves-rwth/storm/pull/393


sjunges:
LGTM

PR #399, opened 2023-07-05
Website: updated information for AUR on Arch Linux

volkm:
- `stormchecker` package on AUR is working again
- Added information which method of Storm installation is supported per OS


sjunges:
LGTM, many thanks!

PR #401, opened 2023-07-06
Enable export to cmake package registry

volkm:
CMake 3.15 does not export to package registry by default (see [here](https://cmake.org/cmake/help/latest/policy/CMP0090.html#policy:CMP0090)). This change allows to explicitly enable the export again.


sjunges:
Everything but the indentation look good to me. Feel free to merge anyways :-)

volkm:
No, we have to have proper indentation ;).

PR #403, opened 2023-07-31
Improvements for CLI

volkm:
Changes:
- Moved `storm-config.h.in` to `src`. I find this better than keeping it in the root dir.
- Extended the information on linked libraries when using `--version`. For instance, information on Carl and Spot is also printed now.
- Unified handling of CLI through function `process(name, executableName, initSettingsFunc, processOptionsFunc, argc, argv)` in which `initSettingsFunc` initializes the settings object and `processOptionsFunc` is the main function processing the input. This unified function ensures consistent behaviour for the binaries and avoids code duplication.
- Consistent use of `STORM_3RDPARTY_SOURCE_DIR`.

Some notes:
- I did not test the output for Soplex and HyPro as I do not have them installed.
- Do we need a generic function for printing build information to a string? For instance for stormpy? 
- Binary `storm-conv` is currently the only binary not using the unified function `process(..)` because it comes with a check whether stdout is enabled. We might want to adapt it as well.
- Further unification could be done by having a generic `main` method as well which handles the outer `try catch` block (see [here](https://github.com/moves-rwth/storm/blob/master/src/storm/storm.cpp#L17)).


src/storm-cli-utilities/print.cpp
```diff
@@ -0,0 +1,208 @@
...
+#include "storm/utility/macros.h"
+
+#include <boost/algorithm/string/replace.hpp>
+#include <ctime>
+
+// Includes for the linked libraries and versions header.
+#include "storm/adapters/IntelTbbAdapter.h"
+
+#ifdef STORM_HAVE_CUDA
+#include <cuda.h>
```
sjunges, line 14: These guarded includes here seem to be the same as in the cli.cpp. Is that something that is worth refactoring?

src/storm-conv-cli/storm-conv.cpp
```diff
@@ -345,9 +346,8 @@ bool parseOptions(const int argc, const char* argv[]) {
     }
 
     if (general.isVersionSet()) {
-        storm::cli::printVersion("storm-conv");
+        storm::cli::printVersion();
```
sjunges, line 349: Why do we no longer need this argument?

CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ The releases of major and minor versions contain an overview of changes since th
 Version 1.8.x
 -------------
 
+## Version 1.8.2
+- Print all linked libraries when using `--version`.
+- Developer: Require at least CMake version 3.15.
+- Developer: Use Dockerfile in CI.
```
sjunges, line 14: For developers, it may be good to highlight you moved the storm-config.h.in ;-)

sjunges:
Hi, great improvements, only minor comments. LGTM.

tquatmann:
I very much like these changes. 

- The stdout option of `storm-conv` can be obsoleted.
- We can likely throw out the HyPro dependency, so there is no need to consider that

src/storm-cli-utilities/print.cpp
volkm: The includes in `cli.cpp` were obsolete and I removed them.

src/storm-conv-cli/storm-conv.cpp
volkm: The name of the binary is from now on only printed through `printHeader` which is the first thing to print for all executions.

CHANGELOG.md
volkm: Yes, good idea.

volkm:
I removed the option `--stdout`.
I can still not use the unified function `process(..)`, because storm-conv introduces its own set of common flags such as `--verbose` or `--trace`. I suggest to add `GeneralSettings`, `DebugSettings` and `ResourceSettings` in storm-conv. These three modules are required for all binaries when using the unified functions.

volkm:
Regarding HyPro, I opened a separate issue https://github.com/moves-rwth/storm/issues/418.

sjunges:
You save me one merge conflict if you do not change storm-pars-cli... :-)

volkm:
Leaving storm-pars-cli will introduce compile issues because other cli functions have changed. Merging now and dumping the merge conflict on Sebastian ;)

PR #405, opened 2023-08-18
Added support for logarithm expressions in PRISM and JANI.

### LINKED ISSUE #384: logarithm expressions are not expressions in PRISM files
# Problem
Although according to the [PRISM Manual](http://www.prismmodelchecker.org/manual/ThePRISMLanguage/Expressions) logarithm expressions are allowed in prism files, storm does not actually recognise log as an expression.

# Minimal Example
min_log.pm:
```
dtmc

module test
	c : [0..1] init 0;
	[] (c = ceil(log(1,2)) -> (c'=1);

endmodule
```

# Error
executing `storm --prism min_log.pm` results in
```
ERROR (SpiritErrorHandler.h:26): Parsing error at 5:19:  expecting <expression>, here:
		[] (c = floor(log(1,2)) -> (c'=1);
	                  ^

ERROR (storm.cpp:23): An exception caused Storm to terminate. The message of the exception is: WrongFormatException: Parsing error at 5:19:  expecting <expression>, here:
		[] (c = floor(log(1,2)) -> (c'=1);
```
# Suggestion to fix
According to @tquatmann this function is not implemented yet. Could you please implement it? :(

tquatmann:
fixes #384

### DIFF
--- a/src/storm/storage/expressions/BinaryNumericalFunctionExpression.cpp
+++ b/src/storm/storage/expressions/BinaryNumericalFunctionExpression.cpp
@@ -55,8 +55,24 @@ storm::expressions::OperatorType BinaryNumericalFunctionExpression::getOperator(
         case OperatorType::Modulo:
             result = storm::expressions::OperatorType::Modulo;
             break;
+        case OperatorType::Logarithm:
+            result = storm::expressions::OperatorType::Logarithm;
+            break;
     }
     return result;
+}
+
+template<typename V>
+V logHelper(V const& x, V const& b) {
+    // There is no std::log method for arbitrary bases.
+    // This catches common cases which should yield more accurate results
+    if (b == static_cast<V>(2)) {
+        return std::log2(x);
+    } else if (b == static_cast<V>(10)) {
+        return std::log10(x);
+    } else {
+        return std::log2(x) / std::log2(b);
+    }
 }
 
 int_fast64_t BinaryNumericalFunctionExpression::evaluateAsInt(Valuation const* valuation) const {
@@ -89,6 +105,9 @@ int_fast64_t BinaryNumericalFunctionExpression::evaluateAsInt(Valuation const* v
             break;
         case OperatorType::Modulo:
             result = firstOperandEvaluation % secondOperandEvaluation;
+            break;
+        case OperatorType::Logarithm:
+            result = logHelper(firstOperandEvaluation, secondOperandEvaluation);
             break;
     }
     return result;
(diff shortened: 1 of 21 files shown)


src/storm/storage/expressions/BinaryNumericalFunctionExpression.cpp
```diff
@@ -90,6 +106,9 @@ int_fast64_t BinaryNumericalFunctionExpression::evaluateAsInt(Valuation const* v
         case OperatorType::Modulo:
             result = firstOperandEvaluation % secondOperandEvaluation;
             break;
+        case OperatorType::Logarithm:
+            result = logHelper(firstOperandEvaluation, secondOperandEvaluation);
```
volkm, line 110: Should this throw an exception when the result is not an integer?

src/storm/storage/expressions/ToCppVisitor.cpp
```diff
@@ -151,6 +151,8 @@ boost::any ToCppVisitor::visit(BinaryNumericalFunctionExpression const& expressi
             expression.getSecondOperand()->accept(*this, data);
             stream << ")";
             break;
+        case BinaryNumericalFunctionExpression::OperatorType::Logarithm:
+            STORM_LOG_THROW(false, storm::exceptions::NotSupportedException, "Log expressions not implemented for C++ translation.");
```
volkm, line 155: Could we not use something like `std::log2(first) / std::log2(second)` here? But might also not be relevant anymore now that JIT support was dropped.

volkm (approved):
Looks good to me. Just two minor questions.

src/storm/storage/expressions/BinaryNumericalFunctionExpression.cpp
tquatmann: I think evaluateAsInt is just supposed to do integer arithmetic. For example we also do not throw for divide and power.

src/storm/storage/expressions/ToCppVisitor.cpp
tquatmann: Yeah I did not include this for now because we dropped jit and it didn't seem relevant. If it becomes relevant at some point, it's not too hard to add, I guess.
`std::log2` requires `#include <cmath>` and I don't know if this can be assumed.

volkm: I agree that we can always add it if needed.

src/storm/storage/expressions/BinaryNumericalFunctionExpression.cpp
volkm: Good to know. The [description of the function](https://github.com/moves-rwth/storm/blob/master/src/storm/storage/expressions/BaseExpression.h#L78) made it seem like it should throw an exception if the result is not integer (but casted to one).

volkm:
We can merge it.

src/storm/storage/expressions/BinaryNumericalFunctionExpression.cpp
tquatmann: Yeah this is not very consequent. Afaik, the description refers to the 'type' as a syntactical thing. Like evaluating `0.5` (a RationalLiteralExpression) as int should throw an error while evaluating `1/2` (the division between two IntegerLiteralExpressions) works.

PR #407, opened 2023-08-21
More efficient MEC and SCC Decompositions

tquatmann:
- option to cache data used during SCC decomposition algorithm, avoiding reallocations
- option to get an SCC decomposition in a more raw format
- rewrite of MEC decomposition algorithm using cache and raw MEC decomposition results

### DIFF
--- a/src/storm/storage/MaximalEndComponentDecomposition.h
+++ b/src/storm/storage/MaximalEndComponentDecomposition.h
@@ -38,6 +38,7 @@ class MaximalEndComponentDecomposition : public Decomposition<MaximalEndComponen
 
     /*
      * Creates an MEC decomposition of the given subsystem of given model (represented by a row-grouped matrix).
+     * If a state-action pair (aka choice) has a transition that leaves the subsystem, the entire state-action pair is ignored.
      *
      * @param transitionMatrix The transition relation of model to decompose into MECs.
      * @param backwardTransition The reversed transition relation.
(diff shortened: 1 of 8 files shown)


sjunges (commented):
So this does look good, without looking into the details of the algorithmic changes. Generallly, I guess it is good to provide a naive != if you provide a ==, but that is very minor.

src/storm/storage/MaximalEndComponentDecomposition.h
```diff
@@ -105,8 +108,8 @@ class MaximalEndComponentDecomposition : public Decomposition<MaximalEndComponen
      * @param choices The choices of the subsystem to decompose.
      */
     void performMaximalEndComponentDecomposition(storm::storage::SparseMatrix<ValueType> const& transitionMatrix,
-                                                 storm::storage::SparseMatrix<ValueType> backwardTransitions, storm::storage::BitVector const* states = nullptr,
-                                                 storm::storage::BitVector const* choices = nullptr);
+                                                 storm::storage::SparseMatrix<ValueType> const& backwardTransitions,
+                                                 storm::storage::BitVector const* states = nullptr, storm::storage::BitVector const* choices = nullptr);
```
sjunges, line 112: We hardly use raw pointers, in particular in the API. Is it necessary that these are pointers here? If so, I guess we should document ownership etc

tquatmann: This is now using storm::OptionalRef, see #414

PR #410, opened 2023-08-28
Add check to only compile clipping tests if Z3 is installed

### LINKED ISSUE #408: storm::test::z3AtLeastVersion not found on linux mint
I tried compiling storm on Linux Mint 20, but make encountered the following error:

/home/heemsjan/storm/src/test/storm-pomdp/modelchecker/BeliefExplorationPomdpModelCheckerTest.cpp: In member function ‘void {anonymous}::BeliefExplorationTest_refuel_Pmin_Clip_Test<gtest_TypeParam_>::TestBody()’:
/home/heemsjan/storm/src/test/storm-pomdp/modelchecker/BeliefExplorationPomdpModelCheckerTest.cpp:981:23: error: ‘z3AtLeastVersion’ is not a member of ‘storm::test’
  981 |     if (!storm::test::z3AtLeastVersion(4, 8, 5)) {
      |                       ^~~~~~~~~~~~~~~~

The error occurs at several other places in the same file.

It does compile a binary that seems to work, but no solver is compiled, so it can't check JANI files.

AlexBork:
Add check to only compile clipping tests if Z3 is installed.

Fixes #408

### DIFF
--- a/src/test/storm-pomdp/modelchecker/BeliefExplorationPomdpModelCheckerTest.cpp
+++ b/src/test/storm-pomdp/modelchecker/BeliefExplorationPomdpModelCheckerTest.cpp
@@ -329,24 +329,6 @@ TYPED_TEST(BeliefExplorationTest, simple_Pmax_SE) {
         << "] is not precise enough. If (only) this fails, the result bounds are still correct, but they might be unexpectedly imprecise.\n";
 }
 
-TYPED_TEST(BeliefExplorationTest, simple_Pmax_Clip) {
-    if (!storm::test::z3AtLeastVersion(4, 8, 5)) {
-        GTEST_SKIP() << "Test disabled since it triggers a bug in the installed version of z3.";
-    }
-    typedef typename TestFixture::ValueType ValueType;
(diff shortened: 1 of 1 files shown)


AlexBork:
I reordered the tests a bit so that only one check is needed. The delta looks more complicated as there a couple of similar lines that were interpreted to not be changed, but functionally nothing besides the check changed.

Witht he solution, I followed how this is handeled for example in src/test/storm/modelchecker/multiobjective/SparseMdpPcaaMultiObjectiveModelCheckerTest.cpp. Generally, I think it's a good idea to let storm::test::z3AtLeastVersion(x, y, z) return false if z3 is not installed as that would reflect the intuitive meaning of the function. As other instances where this is used should also be guarded using the preprocessor directive, changing it should (hopefully) not create any problems. Let me know if you want me to implement that.

sjunges:
Ok, thanks for the explanations. I see that skipping makes sense, and contrary to what I thought, the test-z3-version method is only part of the test-utilities. 

LGTM :-)

PR #413, opened 2023-09-13
Remove storm::utility::simplify instantiation for matrix entries. 

### LINKED ISSUE #411: Compile issue: ambigous partial ordering for simplify()
The following compile error occurs in the CI for Clang 16.0.6. It was also encountered for the upcoming Apple clang version 15.0.0.

`storm/src/storm/utility/constants.cpp:931:83: error: partial ordering for explicit instantiation of 'simplify' is ambiguous` (see [here](https://github.com/moves-rwth/storm/actions/runs/6021262873/job/16333845567#step:4:3434))

tquatmann:
Fixes #411.

I decided to drop the `simplify` instantiations that caused trouble when compiling with newer clang versions.
The removed functions felt a bit out-of-place anyway and were used exactly twice  in the entire code base.
 Fortunately, I could replace these two usages using `simplify` directly on the entry values


volkm (approved):
LGTM. Thanks for fixing it.

PR #414, opened 2023-09-20
Add helper class to store optional references.

tquatmann:
Adds `storm::OptionalRef`, a helper class that optionally holds a reference to an object.
 This mimics the interface of std::optional, except that an OptionalRef never takes ownership of an object.

A possible use case is the implementation of optional function arguments, where std::optional would trigger a deep copy of the object.  For example, `foo(storm::OptionalRef<T const> bar = storm::NullRef)` instead of `foo(T const& bar)`.
 
## Why?
This seems to be the best option for providing optional arguments:

- We use void `foo(std::optional<A const> const& a) {...}`  a lot, but it's imo very implicit that e.g.  in `foo(bar);`  a copy of `bar` is created when it is of type  `A`  or `A const&` .
- Sometimes, we use a raw pointer like `void foo(A const* a) {...}`  which avoids a copy but it's not very explicit about the fact that `a` is optional and it makes ownership unclear.
- `std::optional<A const&>` is not allowed by the CPP standard.
- We could use `void foo(std::optional<std::reference_wrapper<A>> const& a) { ... }`  but it becomes a bit annoying  since in the body of `foo` we'd have to deal with both, the optional and the reference_wrapper, e.g.  `A const& a_ref = a.value().get()`
- Function overloading (i.e. have both a  `foo()` and a `foo(A const& a)` is not always feasible (code duplications, dealing with multiple optional arguments, ...)

### DIFF
--- a/src/storm/utility/OptionalRef.h
+++ b/src/storm/utility/OptionalRef.h
@@ -1,0 +1,182 @@
+#pragma once
+
+#include <memory>
+#include <type_traits>
+
+#include "storm/utility/macros.h"
+
+namespace storm {
+
+/*!
+ * Auxiliary struct used to identify OptionalRefs that do not contain a reference.
+ */
+struct NullRefType {
+} constexpr NullRef;
+
+/*!
+ * Helper class that optionally holds a reference to an object of type T.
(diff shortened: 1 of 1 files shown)


volkm:
LGTM

PR #415, opened 2023-10-04
Fix for multi-objective model checking

tquatmann:
Fixes an issue where sub-optimal schedulers were generated.


sjunges:
LGTM :-)

Marckvdv:
Thanks, this seems to fix the problems I was having!

PR #417, opened 2023-10-18
Fixed issue with out-of-bounds flag

volkm:
Raised in https://github.com/moves-rwth/stormpy/issues/136.

The issue was that initially an empty `VariableInformation` was created and only later replaced by the complete one in the `NextStateGenerator`. As a result information about out-of-bounds values was not considered. I fixed it by adding a new method `initializeSpecialStates` which needs to be explicitly called after creating the VariableInformation.


sjunges:
LGTM, many thanks for fixing this

PR #421, opened 2023-10-31
CI: explicitly set STORM_VERSION_COMMITS_AHEAD

volkm:
Explicitly set `STORM_VERSION_COMMITS_AHEAD` in `version.cmake`. This ensures that the Storm version is correctly set in the Docker container even though no git information is available.

Also updated some Github actions to the latest version.


sjunges:
LGTM! Many thanks

PR #422, opened 2023-11-08
Fix for detecting terminal states

tquatmann:
Depending on the checked property, we might flag some states as terminal, indicating that it is not necessary to continue state-space exploration from them.

The previous code had a bug, where for a nested property like `P=? [ "a" U (P>0.9 [F "b"]) ]` we would stop exploration from `!a`-states, which might invalidate the inner property.

Now, terminal states are only added if both sides of the (Bounded)Until are atomic.


volkm:
LGTM

PR #360, opened 2023-04-28
storm-pars, towards new command line interface & API

sjunges:
- [x] First merge https://github.com/moves-rwth/storm/pull/359.
- [x] I have to update the website.  I will make a separate PR for this here: https://github.com/moves-rwth/storm/pull/426
- [x] I have to check stormpy: https://github.com/moves-rwth/stormpy/pull/138

PR #426, opened 2023-11-12
parameter doc update

sjunges:
Merge together with https://github.com/moves-rwth/storm/pull/360


_bibliography/references.bib
```diff
@@ -731,6 +763,19 @@ @article{JAHJKQV19
   website   = {https://moves-rwth.github.io/prophesy/}
 }
 
+@inproceedings{SJK19,
+  author       = {Jip Spel and
+                  Sebastian Junges and
+                  Joost{-}Pieter Katoen},
+  title        = {Are Parametric Markov Chains Monotonic?},
+  booktitle    = {{ATVA}},
+  series       = {Lecture Notes in Computer Science},
```
volkm, line 772: Very nitpicky, but "LNCS"?

documentation/usage/running-storm-on-parametric-models.md
```diff
@@ -10,108 +10,163 @@ categories: [Use Storm]
...
 
 ```console
-$ storm-pars --prism brp.pm --prop "P>0.5 [F s=5]" --region "0.1 <= pL <= 0.9, 0.1 <=pK <=0.9" --refine 0.01 10
+$ storm-pars  --mode feasibility --feasibility:method pla --prism brp.pm  --prop 'P=? [F s=5]' --direction min --region "0.1 <= pL <= 0.9, 0.1 <= pK <= 0.9" --guarantee 0.0001 abs
 ```
 
-We use the `--regions` option to set the parameter region.
-The arguments we give to `--refine` are the coverage threshold and the depth limit. The coverage threshold decides how precisely we want to partition our region. The value  0.01 means that Storm will continue to refine until only 1% of the given region remains ambiguous while the rest has been determined to be either safe or unsafe. The depth limit is an optional second argument. If the level at which regions are split reaches this bound, the region is not refined any further, regardless of the coverage achieved so far.
+In this case, we do not give a bound and specify with `--direction min` that we want to minimize.
+Additionally, we give a `--guarantee 0.0001 abs`, specifying that we want to be find a solution which is within 0.0001 of the optimal value, where the distance is measured in absolute terms.
```
volkm, line 67: want to find a solution

documentation/usage/running-storm-on-parametric-models.md
```diff
@@ -10,108 +10,163 @@ categories: [Use Storm]
...
 
-This produces the following output:
+The complete output to be expected is:
 
-{% include includes/show_output.html class="pla_parametric_models" path="parametric/brp_pla.out" %}
+{% include includes/show_output.html class="pla_parametric_models" path="parametric/brp_feasibility_pla_guarantee.out" %}
```
volkm, line 71: By giving all log output the  same CSS class, they are all shown/hidden if clicking on any of them. You can give them separate class names to only enable show/hide per log output.

documentation/usage/running-storm-on-parametric-models.md
```diff
@@ -10,108 +10,163 @@ categories: [Use Storm]
...
 
-We can see that for two parameters, Storm even visulalizes the acquired data to give an initial understanding of the region partition.
+In particular, the output says
+
+```console
+Result at initial state: 5282881788238101/9223372036854775808 ( approx. 0.0005727711912) at [pK=287/320, pL=287/320].
+```
+
+This output indicates not only how to instantiate pK and pL to obtain `P=? [F s=5] = approx. 0.0005727711912`, but also that this value is within 0.0001 from the absolute lower boudn (within the given region).
```
volkm, line 79: lower bound

documentation/usage/running-storm-on-parametric-models.md
```diff
@@ -10,108 +10,163 @@ categories: [Use Storm]
...
 
+In contrast, running
 ```console
-$ storm-pars --help mon
+$ storm-pars  --mode verification --prism brp.pm  --prop 'P<=0.99 [F s=5]' --region "0.1 <= pL <= 0.9, 0.1 <= pK <= 0.9"
 ```
+will yield output that says that the statement is *not* true:
+
+{% include includes/show_output.html class="pla_parametric_models" path="parametric/brp_verification_invalid.out" %}
+
```
volkm, line 120: Maybe say that feasibility mode can then be used to find a violating assignment. But I guess this should also be clear by now.

documentation/usage/running-storm-on-parametric-models.md
```diff
@@ -10,108 +10,163 @@ categories: [Use Storm]
...
 
-## Computing extrema
-Another feature Storm offers is the computations of extrema for a property within a given region.
+## Parameter Space Partitioning
 
-### Example 4 (Extremum computation for a bounded retransmission protocol)
-We reuse our BRP model from earlier to compute the maximum probability for a successful transmission that is possible in the given region.
+The verification result may be too coarse: Potentially, we do want to find subregions that satisfy the property.
+While we generally advocate to use our (Python) API for flexible support of such scenarios, the Storm command-line interface does yield some support:
```
volkm, line 125: Give link to stormpy?

documentation/usage/running-storm-on-parametric-models.md
```diff
@@ -10,108 +10,163 @@ categories: [Use Storm]
...
 
 ```console
-$ storm-pars --prism brp.pm --prop "P=? [F s=5]" --region "0.1 <= pL <= 0.9, 0.1 <= pK <= 0.9" --extremum max 0.01
+$ storm-pars  --mode partitioning --prism brp.pm  --prop 'P<=0.99 [F s=5]' --region "0.1 <= pL <= 0.9, 0.1 <= pK <= 0.9"
```
volkm, line 128: Can we add the covered area as a parameter as well to show how this can be manually set?

documentation/usage/running-storm-on-parametric-models.md
```diff
@@ -10,108 +10,163 @@ categories: [Use Storm]
...
+
+{% include includes/show_output.html class="pla_parametric_models" path="parametric/brp_partitioning.out" %}
 
-{% include includes/show_output.html class="extremum_parametric_models" path="parametric/brp_extremum.out" %}
+We can see that for two parameters, Storm even visulalizes the acquired data to give an initial understanding of the region partition.
+
+
+## Computing the exact solution function
+
+We can run storm to obtain closed-form solutions, i.e. a rational function that represents, e.g., the probability with which the given property is satisfied for certain parameter valuations.  A comprehensive overview on the theoretical backgrounds is given in {% cite JAHJKQV19 %}.
```
volkm, line 140: storm -> Storm

volkm:
I found it a bit hard to find the correct reference, because the list of references only has bullet points. We could change `bibliography_list_tag: ol` in `_config.yml` which would yield an ordered list. But this is a global option and then the Publications webpage does not look good.

_bibliography/references.bib
sjunges: Yes, although there were 13 occurrences of Lecture Notes... :-)

volkm: I have a script I can run over it ;)

documentation/usage/running-storm-on-parametric-models.md
sjunges: Maybe in the future, when we have a more stable python API for parametric models. Working on that :-)

PR #427, opened 2023-11-15
Fix not detecting terminal states when Boolean literals are present

tquatmann:
PR #422 introduced an error where for formulas like, e.g., `P=? [true U "a"]` the `"a"` was no longer detected as terminal label.


volkm:
Thanks for the fix. LGTM

PR #430, opened 2023-11-18
two operations on prism programs to support PAYNT

sjunges:
These two operations op prism programs would help development of PAYNT (@randriu) I believe they are sufficiently general to be included in the Storm code base.

### DIFF
--- a/src/storm/storage/prism/Module.cpp
+++ b/src/storm/storage/prism/Module.cpp
@@ -369,6 +369,21 @@ void Module::createMissingInitialValues() {
     }
 }
 
+void Module::removeVariableInitialization() {
+    for (auto& variable : booleanVariables) {
+        variable.setInitialValueExpression(expressions::Expression());
+        assert(!variable.hasInitialValue());
+    }
+    for (auto& variable : integerVariables) {
+        variable.setInitialValueExpression(expressions::Expression());
+        assert(!variable.hasInitialValue());
+    }
+    for (auto& variable : clockVariables) {
+        variable.setInitialValueExpression(expressions::Expression());
+        assert(!variable.hasInitialValue());
+    }
+}
+
 bool Module::hasInvariant() const {
     return this->invariant.isInitialized();
 }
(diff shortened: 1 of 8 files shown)


src/storm/storage/prism/Module.cpp
```diff
@@ -369,6 +369,21 @@ void Module::createMissingInitialValues() {
     }
 }
 
+void Module::removeVariableInitialization() {
+    for (auto& variable : booleanVariables) {
+        variable.setInitialValueExpression(expressions::Expression());
+        assert(!variable.hasInitialValue());
```
volkm, line 375: `STORM_LOG_ASSERT`

src/storm/storage/prism/Program.cpp
```diff
@@ -1174,6 +1174,47 @@ Program Program::labelUnlabelledCommands(std::map<uint64_t, std::string> const&
                    this->getObservationLabels(), this->getOptionalInitialConstruct(), this->getOptionalSystemCompositionConstruct(), prismCompatibility);
 }
 
+Program Program::replaceVariableInitializationByInitExpression() const {
+    std::vector<BooleanVariable> newBooleanVariables;
```
volkm, line 1178: Is this supposed to be an empty constructor? Then the subsequent lines do not have an effect.

src/storm/storage/prism/Program.h
```diff
@@ -696,6 +696,20 @@ class Program : public LocatedInformation {
...
+    /*!
+     * Replace the initialization in variables by an init-expression. This should not change the semantics of the program and can be a preprocessing step.
+     *
+     * @return
+     */
+    Program replaceVariableInitializationByInitExpression() const;
+
+    /**
+     * Substitutes a constant with a given name by a global variable that is bound between lowerBound and upperBound
+     * @return
```
tquatmann, line 708:
```suggestion
     * Substitutes the given constant by a fresh global variable that is bound between lowerBound and upperBound.
     * The provided lowerBound and upperBound are not considered if the constant is Boolean.
     * If the constant is defined by an expression, that expression is used as initial value for the created variable.
     * @note If the given constant appears at positions, where a variable would be invalid (e.g. variable bounds, 
         or the definition of another constant), the resulting program is malformed.
     * @return
```

tquatmann (approved):

sjunges:
Thanks, I will address these points before merging!

PR #432, opened 2023-11-18
move stuff that is only used by clis to cli utils

sjunges:


volkm:
In principle I agree with these changes, but the automatic selection might also be interesting for stormpy. Should we therefore maybe keep it in the main storm library?

sjunges:
That is a good point. I am also okay with leaving it in the main lib. For the QVBS part, I am more motivated because it depends on the settings object and I want to avoid those things in the main lib. The automaticsettings caught my eye because they are called "Settings" but maybe it is more like an AutoConfig or so :-)

sjunges:
I will close this and redo a QVBS move on the updated main branch -- there is a merge conflict right now anyways.

Issue #338, opened 2023-03-22
Move Gamebased-AR to own library


sjunges:
We discussed that as part of our efforts to speed up compilation of the main lib, we would like to move the gamebased-ar code to a separate library.

Issue #433, opened 2023-11-19
CI: Build against MathSat


sjunges:
As far as I can see, we do not download mathsat in our CI script. I think that MathSat 4 prevented us from doing that, but for MathSat 5, this should be possible. 

In particular, I see that we have some non-compiling tests when linking against mathsat 5. I will try to fix them.

PR #434, opened 2023-11-19
move gamebased-ar to own lib

sjunges:
Fixes https://github.com/moves-rwth/storm/issues/338
Also ensures compilation with MathSat again, see also https://github.com/moves-rwth/storm/issues/433

The 105 files touched are not soo bad ;-)

I noticed that there are a lot of thin check result classes that we may want to move back or replace at some point. No need to do that now.


volkm:
I would suggest to also add a namespace `storm::gamebased` or similar.

volkm:
I also saw some changes related to includes in `storm-pars` etc which seem not related to the changes in this PR. Are they needed to ensure compilation?

sjunges:
Regarding the discussion in #435: I cannot do the namespace refactoring today/tomorrow. We may want to consider merging this and i can handle an issue on renaming the namespace later. Or you'll have to wait a bit :-)

volkm:
I am fine with merging this PR and changing the namespace later on.
I guess the changes in storm-pars are needed here because some includes are missing after moving the files?

tquatmann:
I'm also fine with merging this now and change the namespace later.
BTW, my vote is for `storm::abstraction` or `storm::gbar` or `storm::gamebased_ar" or something, since `gamebased` alone is a bit too generic imo

PR #437, opened 2023-11-21
Reworked the BaierUpperRewardBoundsComputer.

tquatmann:
The algorithm is now more efficient (in practise) as it maintains a set of candidates. It might also give slightly tighter bound due to the backward processing of states.
Furthermore, we can now reuse information from context such as backward transitions and SCC decompositions.


sjunges:
LGTM, many thanks

PR #445, opened 2023-11-26
Fixed construction of dependencies in pseudo state for DFTs

volkm:
Fixed issue in symmetry reduction for DFTs where dependencies were not properly initialized.


sjunges:
LGTM

PR #451, opened 2023-12-03
precompiled headers

sjunges:
Adds precompiled headers, which leads to a significant improvement in speed. I currenlty only opted for some boost libraries and standard headers to keep the build as stable as possible.

Notice that this requires cmake 3.16. However, that is included in ubuntu 20.04 and debian oldstable. 

I did not modify storm-pgcl - I want to remove that anyway. 

I did not add an option to disable precompiled headers, this can be done via cmake_disable_precompiled_headers 

- [x] Merge https://github.com/moves-rwth/storm/pull/458 first


volkm:
I think `storm-gamebased-ar` is missing, leading to the failed test.

sjunges:
Well-spotted! I think https://github.com/moves-rwth/storm/pull/458 resolves this.

src/storm/CMakeLists.txt
```diff
@@ -56,6 +56,7 @@ add_dependencies(storm resources)
 target_link_libraries(storm PUBLIC ${STORM_DEP_TARGETS} ${STORM_DEP_IMP_TARGETS} ${STORM_LINK_LIBRARIES})
 list(APPEND STORM_TARGETS storm)
 set(STORM_TARGETS ${STORM_TARGETS} PARENT_SCOPE)
+target_precompile_headers(storm PRIVATE <any> <map> <vector> <optional> <iostream> <set> <fstream> <boost/container/flat_set.hpp> <boost/dynamic_bitset.hpp> <boost/range/irange.hpp>)
```
tquatmann, line 59: How is it decided what is added here? Is there some guideline? Maybe add a comment in the cmake file

```suggestion
target_precompile_headers(storm PRIVATE <any> <map> <vector> <optional> <iostream> <set> <fstream> <functional> <memory> <string> <type_traits> <unordered_map> <unordered_set> <boost/container/flat_set.hpp> <boost/dynamic_bitset.hpp> <boost/range/irange.hpp>)
```

sjunges: I am fixing the bug on arch and will also make sure that I discuss how this is decided.

volkm:
LGTM and the tests are also successful now.

PR #453, opened 2023-12-03
remove an instance of boost::any

sjunges:


volkm:
I can rerun the jobs for some reason. Let's see whether it works the second time.

volkm (approved):
LGTM

sjunges:
I do not understand even remotely what happens here :/

tquatmann:
Given that this happens in Ubuntu 20.04, this could be a bug with, e.g., older boost versions.

Given that `std::any` can be uninitialised, the type `boost::optional<std::any>` does not make a lot of sense in the first place. We could simplify this to just `std::any` and use the [has_value](https://en.cppreference.com/w/cpp/utility/any/has_value) function.

sjunges:
Agreed, thanks for the suggestion.

PR #456, opened 2023-12-03
remove storm-pgcl

sjunges:
As discussed in our developers meeting, we currently have no way to properly support PGCL.
We also see that the academic community has moved significantly further and there are plenty of useful (verification) libraries. We believe it is worthwhile to support the generation of jani-models from some probabilistic while language (potentially in combination with a jani-extension that makes it easier to do so).

If anyone in the future wants to properly support PGCL, we are open to add these functionalities again.


volkm:
LGTM. Thanks for removing it.
We should remember to update the CHANGELOG after all these changes.

PR #459, opened 2023-12-04
Use forward declarations of modernjson in more header files

volkm:
Use `JsonForward.h` instead of `JsonAdapter.h` for more files.

Just for reference: the following files still include `JsonAdapter.h`, mainly because they have a json object as member:
- storm-parsers/parser/JaniParser.h
- storm/storage/Qvbs.h
- storm/storage/jani/visitor/JSONExporter.h
- storm/api/export.h


sjunges:
LGTM

PR #462, opened 2023-12-07
Improved CI by using pre-installed dependencies

### LINKED ISSUE #442: Accelerate CI by precompiling spot
I checked the compile times today in github actions. It seems that we spent roughly 330 seconds building spot, which accounts to roughly 10% of the CompilerTests test times. I think that we can avoid those timings by preinstalling spot into the docker file?

This is of course not urgent in any way :-)

volkm:
Use the new docker image `storm-dependencies` which comes pre-installed with carl-storm, Mathsat, SoPlex and Spot.
Fixes #442 

Also fixed a compile issue for SoPlex triggered by these new tests.


sjunges:
This fixes https://github.com/moves-rwth/storm/issues/442 ?!

volkm:
Yes. I forgot to reference it.

PR #463, opened 2023-12-07
Fixes and improvements for DFT symmetries and DFT simulation

volkm:
- Fixed computation of symmetries in DFT for restrictors and dependencies
- Extended finding of symmetries to all BE types (Weibull, Erlang, ...)
- Added tests for symmetry computations
- Fixed issue where elements under a Mutex where incorrectly set to Don't Care
- Explicit handling of failing BE and triggered dependency in DFT simulator
- Improved computation of initial state in the presence of immediately failed BEs
- Fixed compiler warnings


src/storm-dft/storage/elements/DFTGate.h
```diff
@@ -66,7 +66,7 @@ class DFTGate : public DFTChildren<ValueType> {
             queues.checkRestrictionLater(restr);
         }
         state.setFailed(this->mId);
-        this->childrenDontCare(state, queues);
+        this->childrenDontCare(queues);
```
sjunges, line 69: I guess this is what you want, but I was surprised that you can mark things not just in the state that you are considering? As in, i dont understand what is happening here.

sjunges:
LGTM. I flagged one part of the code where I did not understand what is happening.

src/storm-dft/storage/elements/DFTGate.h
volkm: The line just adds the children to a queue for future consideration. Another function is then taking care of setting the state.
I agree that this is quite confusing and I think it can be improved. I would opt to make a separate PR for this though.

sjunges: Sure! It is actually better that it is now explicit that this code somehow does not require the state as an argument, but it does confuse me a lot :-)

volkm: Issue https://github.com/moves-rwth/storm/issues/470 keeps track of this part.

PR #465, opened 2023-12-09
remove one more header include of JsonAdapter

sjunges:
Three occurrences of the json adapter in a header remain (notice that this is bad for compile times).


volkm:
LGTM.
For the remaining headers we might have to change the members to shared pointers to allow for forward declarations.

sjunges:
Indeed!  I will make an updated remark in the compile times issue.

PR #466, opened 2023-12-09
various improvements on the inclusion structure

sjunges:
Mostly issues to reduce what code depends on settings and based on analysis of compilation times. 
Also added a couple of missing std's


volkm:
LGTM

PR #471, opened 2023-12-13
Removed documentation on pGCL as support was removed

volkm:
Updated documentation after https://github.com/moves-rwth/storm/pull/456


sjunges:
Great. I did not think about the website. Many thanks

PR #472, opened 2023-12-16
bugfix in overlapping guards analysis

sjunges:
TQ rightfully pointed out that this should not be part of the huge interval commit..

Regarding content: The conjunction is not defined on empty sets. This fixes the issue.


tquatmann (approved):
LGTM

PR #473, opened 2023-12-17
refactoring towards compile speed

sjunges:


volkm:
LGTM but there is an `#include <carl/formula/Formula.h>` missing, see CI.

sjunges:
Removing that include is (one of) the point(s) of this PR. So i have to think a bit more.

sjunges:
@linusheck Could you eventually look into this? THat would be great

PR #477, opened 2023-12-28
prism program ops improvements and potential to update initial states

sjunges:
- Initial construct does not need to be exposed to the outside world. Makes the API tiny bit easier.
- Comparing two constants is helpful 
- Erase & Remove idiom correctly applied
- Updating initial state expression is helpful for programmatically changing the initial state.


volkm (approved):
LGTM

PR #478, opened 2023-12-28
removed buggy transformer to eliminate constant rewards 

sjunges:
As discussed with TQ.  A fixed version  with tests can be added later if someone needs it.


volkm (approved):
LGTM

PR #484, opened 2024-01-02
Removed warnings in tests

volkm:
Tests should build without warnings in the CI.
- removed hard-coded path and formula in `RobustMdpPrctlModelCheckerTest`
- removed `env` parameter in POMDP modelchecker which is unused


sjunges:
LGTM, thanks for the fix.

PR #486, opened 2024-01-02
remove some expensive header includes

sjunges:


volkm (commented):
The changes look good to me, especially because it gradually forces us to make the includes more explicit. The CI seems to fail because some includes are still missing.

sjunges:
I hate that one has to write the includes manually, but it saves significantly on compile times.... there are just too many files that do not depend on rational numbers / rational functions and still include those. That is not only expensive for the headers, but also because it forces all kind of instantiations.

sjunges:
Before the logs disappear again: 

```
#16 104.5 In file included from /opt/storm/src/storm/modelchecker/AbstractModelChecker.cpp:1:
#16 104.5 In file included from /opt/storm/src/storm/modelchecker/AbstractModelChecker.h:7:
#16 104.5 /opt/storm/src/storm/modelchecker/CheckTask.h:229:69: error: implicit instantiation of undefined template '__gmp_expr<__mpq_struct[1], __mpq_struct[1]>'
#16 104.5   229 |         return storm::utility::convertNumber<ValueType>(bound.get().evaluateThresholdAsRational());
#16 104.5       |                                                                     ^
#16 104.5 /opt/storm/src/storm/adapters/RationalNumberForward.h:9:7: note: template is declared here
#16 104.5     9 | class __gmp_expr;
```

PR #488, opened 2024-01-03
fix warnings

sjunges:


volkm (approved):
Good catches w.r.t. `pragma` and thanks for disabling the Eigen warnings.

sjunges:
I saw that I introduced a new warning. Will delay merging this until I have time.

sjunges:
The problem is that old gcc version do not support -Wdeprecated-enum-enum-conversion while newers throw a warning. How shall i handle this? Accept the warning about pragmas in old gccs? Maybe we can detect GCC versions and disable warnings about unknown pragmas on old gccs? (but i really would love to focus on new gcc versions)

sjunges:
@tquatmann  decided that introducing the warning on Ubuntu 20.04 and Debian 11 is ok for now

PR #490, opened 2024-01-03
disable all kind of warnings when setting clang-allwarnings

sjunges:
I wanted to re-enable the allwarnings on my machine and that was such a mess that I disabled many of them for now. Hope it makes this useful again.


volkm (approved):
LGTM.
I can take a look at the GCC warnings in the near future.

Issue #250, opened 2022-07-26
Abstraction engine not considered in canHandle


volkm:
@Vescatur found that using the abstraction engine always prints an error. However, the model-checking query can still be successfully executed.
The problem is that `canHandle` in [utility/Engine.cpp](https://github.com/moves-rwth/storm/blob/master/src/storm/utility/Engine.cpp#L107) is missing the case for this engine.

An example call is the following:
`./bin/storm --jani qcomp/benchmarks/mdp/beb/beb.3-4.jani --janiproperty LineSeized --constants N=3 --core:engine abs`

tquatmann:
The exploration engine seems to be affected, as well.

PR #491, opened 2024-01-05
Fixed canHandle check for exploration and abstraction engine

tquatmann:
Fixes a current issue that when selecting `--engine expl` or `--engine abs` we always get output like this (no matter if the property is supported or not)

```
ERROR (Engine.cpp:160): The selected engine expl is not considered.
ERROR (Engine.cpp:163): The selected combination of engine (expl) and model type (dtmc) does not seem to be supported for this value type.
 WARN (storm.cpp:27): The model checking query does not seem to be supported for the selected engine. Storm will try to solve the query, but you will most likely get an error for at least one of the provided properties.
```


volkm:
Fixes https://github.com/moves-rwth/storm/issues/250

volkm (approved):
LGTM

PR #493, opened 2024-01-17
Improved expected visiting times (EVTs) and steady state distribution computations

tquatmann:
This PR incorporates the result of our TACAS'24 Paper "Accurately Computing Expected Visiting Times and Stationary Distributions in Markov Chains"  Preprint available [here](https://arxiv.org/abs/2401.10638).

`--sound` mode now works properly with correct accuracy using OVI and II for EVTs.
Steady-state distributions can now be computed using a reduction to EVT-based computations.

Using `--expvisittimes --prop 'Phi'` now restricts the result to states satisfying qualitative formula Phi. 
`--steadystate --prop 'Phi'` works similarly.

Co-authored-by: @m-hannah

### DIFF
--- a/src/storm-cli-utilities/model-handling.h
+++ b/src/storm-cli-utilities/model-handling.h
@@ -440,9 +440,8 @@ std::shared_ptr<storm::models::ModelBase> buildModelDd(SymbolicInput const& inpu
                                                              !buildSettings.isApplyNoMaximumProgressAssumptionSet());
 }
 
-template<typename ValueType>
-std::shared_ptr<storm::models::ModelBase> buildModelSparse(SymbolicInput const& input, storm::settings::modules::IOSettings const& ioSettings,
-                                                           storm::settings::modules::BuildSettings const& buildSettings) {
+storm::builder::BuilderOptions createBuildOptionsSparse(SymbolicInput const& input, storm::settings::modules::IOSettings const& ioSettings,
+                                                        storm::settings::modules::BuildSettings const& buildSettings) {
     storm::builder::BuilderOptions options(createFormulasToRespect(input.properties), input.model.get());
     options.setBuildChoiceLabels(options.isBuildChoiceLabelsSet() || buildSettings.isBuildChoiceLabelsSet());
     options.setBuildStateValuations(options.isBuildStateValuationsSet() || buildSettings.isBuildStateValuationsSet());
@@ -481,7 +480,11 @@ std::shared_ptr<storm::models::ModelBase> buildModelSparse(SymbolicInput const& 
     if (ioSettings.isComputeExpectedVisitingTimesSet() || ioSettings.isComputeSteadyStateDistributionSet()) {
         options.clearTerminalStates();
     }
+    return options;
+}
 
+template<typename ValueType>
+std::shared_ptr<storm::models::ModelBase> buildModelSparse(SymbolicInput const& input, storm::builder::BuilderOptions const& options) {
     return storm::api::buildSparseModel<ValueType>(input.model.get(), options);
 }
 
@@ -518,7 +521,8 @@ std::shared_ptr<storm::models::ModelBase> buildModel(SymbolicInput const& input,
         if (builderType == storm::builder::BuilderType::Dd) {
             result = buildModelDd<DdType, ValueType>(input);
         } else if (builderType == storm::builder::BuilderType::Explicit) {
-            result = buildModelSparse<ValueType>(input, ioSettings, buildSettings);
+            auto options = createBuildOptionsSparse(input, ioSettings, buildSettings);
+            result = buildModelSparse<ValueType>(input, options);
         }
     } else if (ioSettings.isExplicitSet() || ioSettings.isExplicitDRNSet() || ioSettings.isExplicitIMCASet()) {
         STORM_LOG_THROW(mpi.engine == storm::utility::Engine::Sparse, storm::exceptions::InvalidSettingsException,
@@ -962,6 +966,12 @@ struct PostprocessingIdentity {
     }
 };
 
+/*!
+ * Verifies the given formula plus a filter formula to identify relevant states and warns the user in case of issues
+ * @param formula the formula to check
+ * @param statesFilter a second formula that identifies the relevant states (needs to be qualitative)
+ * @param verificationCallback Function to perform the actual verification task for a given formula plus a filter formula to identify relevant states
+ */
 template<typename ValueType>
 std::unique_ptr<storm::modelchecker::CheckResult> verifyProperty(std::shared_ptr<storm::logic::Formula const> const& formula,
                                                                  std::shared_ptr<storm::logic::Formula const> const& statesFilter,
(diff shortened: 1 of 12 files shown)


sjunges:
the clang-debug run reveals some novel warnings introduced by this code :-) 

Could you have a look?

src/storm-cli-utilities/model-handling.h
```diff
@@ -441,7 +441,8 @@ std::shared_ptr<storm::models::ModelBase> buildModelDd(SymbolicInput const& inpu
 }
 
 template<typename ValueType>
-std::shared_ptr<storm::models::ModelBase> buildModelSparse(SymbolicInput const& input, storm::settings::modules::BuildSettings const& buildSettings) {
+std::shared_ptr<storm::models::ModelBase> buildModelSparse(SymbolicInput const& input, storm::settings::modules::IOSettings const& ioSettings,
```
sjunges, line 444: I understand this, but can we refactor this such that we do not pass more and more settings around. Can we do a createOptions first?

src/storm-cli-utilities/model-handling.h
```diff
@@ -947,48 +952,102 @@ void printResult(std::unique_ptr<storm::modelchecker::CheckResult> const& result
...
-                } else {
-                    ignored = true;
-                }
-            } else {
-                result = verificationCallback(property.getRawFormula(), property.getFilter().getStatesFormula());
-            }
-        } catch (storm::exceptions::BaseException const& ex) {
-            STORM_LOG_WARN("Cannot handle property: " << ex.what());
-        }
+        auto result = verifyProperty<ValueType>(property.getRawFormula(), property.getFilter().getStatesFormula(), verificationCallback);
```
sjunges, line 998: :-)

src/storm-cli-utilities/model-handling.h
```diff
@@ -1110,7 +1169,7 @@ void verifyWithSparseEngine(std::shared_ptr<storm::models::ModelBase> const& mod
         std::unique_ptr<storm::modelchecker::CheckResult> filter;
         if (filterForInitialStates) {
             filter = std::make_unique<storm::modelchecker::ExplicitQualitativeCheckResult>(sparseModel->getInitialStates());
-        } else {
+        } else if (!states->isTrueFormula()) {
             filter = storm::api::verifyWithSparseEngine<ValueType>(mpi.env, sparseModel, storm::api::createTask<ValueType>(states, false));
```
sjunges, line 1173: do we need a new else case?  looks like filter is otherwise unitialized? Or are those semantics ok?

src/storm-cli-utilities/model-handling.h
```diff
@@ -1201,7 +1248,7 @@ void verifyWithHybridEngine(std::shared_ptr<storm::models::ModelBase> const& mod
             if (filterForInitialStates) {
                 filter = std::make_unique<storm::modelchecker::SymbolicQualitativeCheckResult<DdType>>(symbolicModel->getReachableStates(),
                                                                                                        symbolicModel->getInitialStates());
-            } else {
+            } else if (!states->isTrueFormula()) {
                 filter = storm::api::verifyWithHybridEngine<DdType, ValueType>(mpi.env, symbolicModel, storm::api::createTask<ValueType>(states, false));
```
sjunges, line 1252: same comment as before

src/storm-cli-utilities/model-handling.h
```diff
@@ -1226,7 +1273,7 @@ void verifyWithDdEngine(std::shared_ptr<storm::models::ModelBase> const& model,
             if (filterForInitialStates) {
                 filter = std::make_unique<storm::modelchecker::SymbolicQualitativeCheckResult<DdType>>(symbolicModel->getReachableStates(),
                                                                                                        symbolicModel->getInitialStates());
-            } else {
+            } else if (!states->isTrueFormula()) {
```
sjunges, line 1276: and once more

src/storm/modelchecker/helper/indefinitehorizon/visitingtimes/SparseDeterministicVisitingTimesHelper.h
```diff
@@ -100,30 +116,53 @@ class SparseDeterministicVisitingTimesHelper : public SingleValueModelCheckerHel
...
+     * Solves the equation system for non-singleton (subs)sets of the chain's non-bottom states.
+     * @return for each state of the given set the expected number of times that state is visited.
      */
-    std::vector<ValueType> computeValueForNonTrivialScc(storm::Environment const& env, storm::storage::BitVector const& sccAsBitVector,
-                                                        std::vector<ValueType> const& stateValues) const;
+    std::vector<ValueType> computeValueForStateSet(storm::Environment const& env, storm::storage::BitVector const& stateSetAsBitVector,
+                                                   std::vector<ValueType> const& stateValues) const;
+
+    storm::storage::SparseMatrix<ValueType> const& transitionMatrix;
+    std::vector<ValueType> const* exitRates;
```
sjunges, line 157: are these const* intended?

src/storm/settings/modules/IOSettings.h
```diff
@@ -5,6 +5,7 @@
 
 #include "storm-config.h"
 #include "storm/builder/ExplorationOrder.h"
+#include "storm/environment/modelchecker/ModelCheckerEnvironment.h"
```
sjunges, line 8: i am not so happy with this include. Including the IOSettings now means including all environments. Maybe that is anyway already the case, so dont sweat it :-)

src/storm-cli-utilities/model-handling.h
tquatmann: The next lines say
```c++
        if (result && filter) {
            result->filter(filter->asQualitativeCheckResult());
        }
```
So if `filter` remains uninitialized, no filtering is applied which is exactly what we want if we filter for states satisfying `true`. Maybe I can add `// No need to apply filter for formula 'true'` to make this case more explicit

src/storm-cli-utilities/model-handling.h
tquatmann: I made a `createOptionsSparse` which makes it a bit more clear what's happening but does not really avoid passing around the settings. Any suggestions?

src/storm/modelchecker/helper/indefinitehorizon/visitingtimes/SparseDeterministicVisitingTimesHelper.h
tquatmann: Thanks, this is a case for `storm::OptionalRef`

tquatmann:
Thanks, great feedback!
I fixed this warning:
```
storm/src/storm/modelchecker/helper/indefinitehorizon/visitingtimes/SparseDeterministicVisitingTimesHelper.cpp:395:35: warning: comparison of integers of different signs: 'int' and 'uint_fast64_t' (aka 'unsigned long long') [-Wsign-compare]
                for (int i = 1; i < sccDecomposition->getMaxSccDepth(); i++) {
```

src/storm-cli-utilities/model-handling.h
sjunges: I think that is a significant improvement because it isolates the settings from the rest of the logic. I would also not mind if that createOptionsSparseFromSettings would just grab the settings from the manager, rather than them being passed around? (It is a function that explicitly depends on settings, so ...)

src/storm-cli-utilities/model-handling.h
sjunges: Hi, yeah this is the problem with only reviewing a code snippet :-) The comment does help though.

src/storm-cli-utilities/model-handling.h
tquatmann: Yes, 100% agree. I've updated this now.

sjunges:
LGTM

PR #494, opened 2024-01-25
Fixed DdPrismModelBuilder

tquatmann:
The bug could occur when the same action appeared in multiple modules and multiple commands.

The revised code is now similar to the analogous code in the [DdJaniModelBuilder](https://github.com/tquatmann/storm/blob/fb047739d71d38f4e8b9e008696e0d927e5f84bd/src/storm/builder/DdJaniModelBuilder.cpp#L666)

Thanks @volkm for reporting this

### DIFF
--- a/resources/examples/testfiles/mdp/sync.nm
+++ b/resources/examples/testfiles/mdp/sync.nm
@@ -1,0 +1,15 @@
+mdp
+
+module m1
+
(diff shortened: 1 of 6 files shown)


volkm (approved):
LGTM. Thanks for the quick fix!

PR #496, opened 2024-01-27
resolve unused parameter warnings in spotproduct

sjunges:


volkm:
The issue is really weird. It seems like the source code from the master is taken instead of the one from the PR. Otherwise I do not see where `(void) model` comes from. But on the other hand the function `ltl2daSpotProduct` is defined without `model` as argument, meaning that the PR version is taken.

Can you try merging the master and see whether it makes a difference?

volkm:
The merge did not help. I will investigate it further...

PR #497, opened 2024-01-28
Better automatic detection of available threads (fixes #495)

### LINKED ISSUE #495: Sylvan: Better detection of thread count
It seems that the automatic detection of the number of supported concurrent threads in Sylvan (or more specifically Lace) is not very accurate when it is restricted via, e.g., Docker or  Slurm (more specifically, I *think* these are all using cgroups). This leads to slow execution of DD code and often requires to manually set the thread count using `--sylvan:threads N`.

When googling this, [std::thread::hardware_concurency()](https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency) pops up, but this also seems to yield the number of  physical threads the hardware provides (and not the number of threads we can actually use). I haven't tested this, though.

tquatmann:
* consider limit imposed by SLURM
* consider limit imposed by cgroups (e.g. Docker)
* print how many threads are used by sylvan

### DIFF
--- a/src/storm/storage/dd/sylvan/InternalSylvanDdManager.h
+++ b/src/storm/storage/dd/sylvan/InternalSylvanDdManager.h
@@ -38,6 +38,8 @@ class InternalDdManager<DdType::Sylvan> {
      */
     ~InternalDdManager();
 
+    static std::size_t getNumberOfThreads(std::size_t numberFromSettings = 0);
+
     /*!
      * Retrieves a BDD representing the constant one function.
      *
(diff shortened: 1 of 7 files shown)


src/storm/storage/dd/sylvan/InternalSylvanDdManager.h
```diff
@@ -38,6 +38,8 @@ class InternalDdManager<DdType::Sylvan> {
      */
     ~InternalDdManager();
 
+    static std::size_t getNumberOfThreads(std::size_t numberFromSettings = 0);
+
```
volkm, line 42: Seems to be unused.

volkm (commented):
LGTM. Just one small comment.

PR #500, opened 2024-02-10
interval model checking fix for postprocessing results

sjunges:
For interval models, the model checking result actually contains values for every state. So we do not need to merge it with the qualitative state results.


volkm (approved):
LGTM. Thanks for the quick fix.

PR #501, opened 2024-02-13
replace print with logging when computing expected visits

randriu:
STORM_PRINT produces unwanted messages when computing expected visits multiple times, and (as far as I know) cannot be disabled.


tquatmann:
LGTM!

I fixed the automatic formatting (hope you don't mind).

randriu:
Thanks a lot!

PR #503, opened 2024-02-20
Fix jani functions in properties

tquatmann:
Previously, functions occurring in properties were not handled correctly. This was the case because all functions declared in the model were eliminated before simplifying the properties. See `preprocessSymbolicInput` in model-handling.h, where the SymbolicModelDescription is [preprocessed](https://github.com/moves-rwth/storm/blob/24cb52a9959cb11d55c525cf5dab31e94cfb6390/src/storm-cli-utilities/model-handling.h#L371) before [the Jani model is simplified](https://github.com/moves-rwth/storm/blob/24cb52a9959cb11d55c525cf5dab31e94cfb6390/src/storm-cli-utilities/model-handling.h#L401).

Since the `preprocess()` method was used in the `DdJaniModelBuilderTest` to (also) eliminate functions, I reworked this (and for consistency, the `ExplicitJaniModelBuilderTest` as well).

### DIFF
--- a/src/storm/storage/SymbolicModelDescription.cpp
+++ b/src/storm/storage/SymbolicModelDescription.cpp
@@ -167,7 +167,9 @@ SymbolicModelDescription SymbolicModelDescription::preprocess(std::string const&
 SymbolicModelDescription SymbolicModelDescription::preprocess(
     std::map<storm::expressions::Variable, storm::expressions::Expression> const& constantDefinitions) const {
     if (this->isJaniModel()) {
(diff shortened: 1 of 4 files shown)


sjunges:
Looks good. May be good to avoid the now even more appearant code duplication.

tquatmann:
> Looks good. May be good to avoid the now even more appearant code duplication.

Absolutely! The builder tests could need a refactor and should be templated like (most of) the model checker tests.

PR #507, opened 2024-02-21
Fixes for explicit jani model builder

tquatmann:
- Fixed building JANI models that use the same automaton multiple times in their system composition. Previously, the different instances referred to the same expression variables for the automaton location and its local variables.

- Fixed building JaniChoiceOrigins which previously referred to internal automata indices and not the ones from the JANI model.

### DIFF
--- a/src/storm/generator/JaniNextStateGenerator.cpp
+++ b/src/storm/generator/JaniNextStateGenerator.cpp
@@ -68,6 +68,8 @@ JaniNextStateGenerator<ValueType, StateType>::JaniNextStateGenerator(storm::jani
     }
     STORM_LOG_THROW(features.empty(), storm::exceptions::NotSupportedException,
                     "The explicit next-state generator does not support the following model feature(s): " << features.toString() << ".");
+    // Simplify the system compositions so that we can exclude the case where automata appear in the composition multiple times.
+    this->model.simplifyComposition();
 
     // Get the reward expressions to be build. Also find out whether there is a non-trivial one.
     bool hasNonTrivialRewardExpressions = false;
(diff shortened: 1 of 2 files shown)


src/storm/generator/JaniNextStateGenerator.cpp
```diff
@@ -1069,6 +1072,7 @@ std::vector<Choice<ValueType>> JaniNextStateGenerator<ValueType, StateType>::get
                                                                 automatonIndex, state, stateToIdCallback));
 
                     if (this->getOptions().isBuildChoiceOriginsSet()) {
+                        auto modelAutomatonIndex = model.getAutomatonIndex(parallelAutomata[automatonIndex].get().getName());
```
sjunges, line 1075: This looks complicated, but fine.

tquatmann: Yes, it is. Alternatively, one could store an index mapping between the `parallelAutomata` vector and the original indices in the jani model. However, my feeling is that the current solution is less error prone.

src/storm/generator/JaniNextStateGenerator.cpp
```diff
@@ -68,6 +68,8 @@ JaniNextStateGenerator<ValueType, StateType>::JaniNextStateGenerator(storm::jani
     }
     STORM_LOG_THROW(features.empty(), storm::exceptions::NotSupportedException,
                     "The explicit next-state generator does not support the following model feature(s): " << features.toString() << ".");
+    // Simplify the system compositions so that we can exclude the case where automata appear in the composition multiple times.
```
sjunges, line 71: So simplify also brings it in some kind of normal form that is necessary?  The name "simplify" may be a bit misleading, but the comment is sufficient :-)

sjunges:
LGTM

src/storm/generator/JaniNextStateGenerator.cpp
tquatmann: One could argue that simplify brings it to a *simple form* that is necessary. But I agree that `simplifyComposition` could be renamed to something more descriptive.

PR #508, opened 2024-02-28
Better error message when checking instantaneous reward formula without state rewards

tquatmann:
The former message was a bit misleading.


volkm (approved):
LGTM

PR #510, opened 2024-03-05
Fixed wrong size threshold if exploration time limit is given

AlexBork:
This PR removes a line that always sets the threshold for the size of the belief MDP to the heuristic value computed by Storm if it is not explicitly set different from the default.
The intended behaviour is for the belief exploration to ignore the size threshold if a time limit for the belief exploration and no explicit size threshold are given.


tquatmann:
LGTM!

PR #511, opened 2024-03-13
Jani: Support for rates in DTMCs

tquatmann:
The jani specification allows rates to occur at edges in DTMCs which are then interpreted as weights. Previously, this case was not handled properly.

This implements the weighting for explicit (sparse) model building and throws an error for symbolic model building.


tquatmann:
I'm closing this PR since DTMC rates recently got removed from the jani-spec.

PR #512, opened 2024-03-15
Optimization of robust VI

glatteis:
This PR contains small adjustments to robust VI that significantly speed up the computations (at least 10x). Profiling revealed that re-allocating the vector is costly (already pointed out by the comment) and `diameter()` changes the rounding mode on each call, which is very expensive. I could not manage to get faster sorting than `std::sort`, although I haven't tried using `boost::container::flatmap`. I will try that in a private branch later :) for now it's a good improvement.

### DIFF
--- a/src/storm/solver/helper/ValueIterationOperator.h
+++ b/src/storm/solver/helper/ValueIterationOperator.h
@@ -411,18 +411,21 @@ class ValueIterationOperator {
      */
     bool auxiliaryVectorUsedExternally{false};
 
-    template<typename ApplyValueType>
+    // Due to a GCC bug we have to add this dummy template type here
+    // https://stackoverflow.com/questions/49707184/explicit-specialization-in-non-namespace-scope-does-not-compile-in-gcc
+    template<typename ApplyValueType, typename Dummy>
     struct ApplyCache{};
 
-    template<>
-    struct ApplyCache<storm::Interval> {
+    template<typename Dummy>
+    struct ApplyCache<storm::Interval, Dummy> {
         mutable std::vector<std::pair<SolutionType, SolutionType>> robustOrder;
+        Dummy x;
     };
 
     /*!
      * Cache for robust value iteration, empty struct for other ValueTypes than storm::Interval.
      */
-    ApplyCache<ValueType> applyCache;
+    ApplyCache<ValueType, int> applyCache;
 
     /*!
      * Bitmask that indicates the start of a row in the 'matrixColumns' vector
--- a/src/storm/solver/helper/ValueIterationOperator.h
+++ b/src/storm/solver/helper/ValueIterationOperator.h
@@ -291,8 +291,8 @@ class ValueIterationOperator {
                         OperandType const& operand, OffsetType const& offsets, uint64_t offsetIndex) const {
         STORM_LOG_ASSERT(*matrixColumnIt >= StartOfRowIndicator, "VI Operator in invalid state.");
         auto result{robustInitializeRowRes<RobustDirection>(operand, offsets, offsetIndex)};
-        static std::vector<std::pair<SolutionType, SolutionType>> robustOrder;
-        robustOrder.clear();
+        AuxCompare<RobustDirection> compare;
+        applyCache.robustOrder.clear();
 
         SolutionType remainingValue{storm::utility::one<SolutionType>()};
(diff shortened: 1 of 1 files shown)


glatteis:
Hm, CI says 30.19 error: failed retrieving file 'gnutls-3.8.3-1-x86_64.pkg.tar.zst.sig' from geo.mirror.pkgbuild.com : Operation too slow. Less than 1 bytes/sec transferred the last 10 seconds

volkm:
Thanks for the changes. 10x faster sounds really good. The changes look good to me.

volkm:
Can you try restarting the failing CI job? The error should be independent of your changes.

src/storm/solver/helper/ValueIterationOperator.h
```diff
@@ -291,35 +291,41 @@ class ValueIterationOperator {
                         OperandType const& operand, OffsetType const& offsets, uint64_t offsetIndex) const {
         STORM_LOG_ASSERT(*matrixColumnIt >= StartOfRowIndicator, "VI Operator in invalid state.");
         auto result{robustInitializeRowRes<RobustDirection>(operand, offsets, offsetIndex)};
-        std::vector<std::pair<SolutionType, SolutionType>> tmp;  // TODO this reallocation is too costly.
+        static std::vector<std::pair<SolutionType, SolutionType>> robustOrder;
```
tquatmann, line 294: Can we avoid having this `static`?  I see how the cached memory is helpful, but maybe use a private member like this:
```
template <typename VT>
struct ApplyCache {};

template<>
struct ApplyCache <storm::Interval>
{
    std::vector<std::pair<SolutionType, SolutionType>> robustOrder;
};

ApplyCache<ValueType> applyCache;
```

This way, there is no robustOrder for Operands that are not using intervals.

src/storm/solver/helper/ValueIterationOperator.h
```diff
@@ -291,35 +291,41 @@ class ValueIterationOperator {
...
                 // Notice the unclear semantics here in terms of how to order things.
             } else {
-                result += operand[*matrixColumnIt] * (matrixValueIt->lower());
+                result += operand[*matrixColumnIt] * lower;
             }
-            remainingValue -= matrixValueIt->lower();
-            if (!storm::utility::isZero(matrixValueIt->diameter())) {
-                tmp.emplace_back(operand[*matrixColumnIt], matrixValueIt->diameter());
+            remainingValue -= lower;
+            // TODO Float rounding modes are not respecting optimization direction here, is this an issue?
```
tquatmann, line 307: None of the value iteration-based techniques currently respect rounding modes.
If we ever were to incorporate that, I believe changes would be required all over this place.
Thus, I'm in favour of dropping this comment as it suggests that the remaining code *does* float respect rounding modes.

src/storm/solver/helper/ValueIterationOperator.h
```diff
@@ -291,35 +291,41 @@ class ValueIterationOperator {
...
         if (storm::utility::isZero(remainingValue) || storm::utility::isOne(remainingValue)) {
             return result;
         }
-        AuxCompare<RobustDirection> compare;
-        std::sort(tmp.begin(), tmp.end(), compare);
 
-        for (auto const& valWidthPair : tmp) {
-            auto availableMass = std::min(valWidthPair.second, remainingValue);
-            result += availableMass * valWidthPair.first;
+        static AuxCompare<RobustDirection> cmp;
```
tquatmann, line 317: Again, I'd be in favour of dropping `static`. Does this have any performance impact?

tquatmann (approved):
Looks good, thanks!

Generally, I'm in favour of avoiding static variables as much as possible, so please have a look at my comments.

src/storm/solver/helper/ValueIterationOperator.h
glatteis: I'll be dropping the static and change it according to your suggestion :)

src/storm/solver/helper/ValueIterationOperator.h
glatteis: Comment deleted :)

src/storm/solver/helper/ValueIterationOperator.h
glatteis: Done :)

tquatmann:
LGTM! Thanks for the changes.

glatteis:
just FYI, I currently have no rights to merge the PR :D

PR #513, opened 2024-03-16
Get z3 version by parsing version file

volkm:
Recently, CI builds fail because they are sometimes stuck in obtaining the z3 version. See [here](https://github.com/moves-rwth/storm/actions/runs/8305864391/job/22738791859#step:3:200) for an example.
The problem might be due to compiling a small program with z3 in order to obtain the z3 version.

This PR obtains the z3 version by parsing the file `z3_version.h` instead. The old option is still present to preserve backwards compatibility to versions before [this fix](https://github.com/Z3Prover/z3/issues/1833).

### DIFF
--- a/resources/3rdparty/CMakeLists.txt
+++ b/resources/3rdparty/CMakeLists.txt
@@ -199,13 +199,23 @@ find_package(Z3 QUIET)
 # Z3 Defines
 set(STORM_HAVE_Z3 ${Z3_FOUND})
 
-
 if(Z3_FOUND)
-    # Get the z3 version by compiling and running a simple program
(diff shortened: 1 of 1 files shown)


sjunges:
I am fine with this, but it adds additional complexity that I would not include if this is actually not necessary?

volkm:
The change is currently not needed anymore.
I personally find the new way of parsing the version better than compiling a small program though. The latter can also fail due to compilation issues which are not coming from z3 (as seen when this issue was triggered).

sjunges:
Do you think we need to retain the old variant as well then?

volkm:
I checked and then z3 version file is present since 4.8.1. Even Ubuntu 20.04 ships with z3 version 4.8.7. So we should be fine with removing the old check.

PR #519, opened 2024-04-11
Jani builder: Fix referring to the wrong kind of automaton index

tquatmann:
This fixes a small error introduced in PR #507


volkm:
👍

PR #530, opened 2024-05-08
Fixed issue with const failed BEs under SEQ which are triggered by FDEP

volkm:
Needed to check if a previously disabled dependency could be enabled through the failure of a BE under a SEQ.

Additional changes:
- added test case for the issue
- moved function `getDynamicBehavior()` from `DFT` to `FDEPConflictFinder`
- use `beType` instead of `type()` for BEs,


sjunges:
I scrolled through this and I have no comments ;-) feel free to merge

volkm:
Thanks for taking a look.

PR #531, opened 2024-05-13
Handle reward bounded formulae without specified reward model

AlexBork:
This PR extends Storm to handle reward bounded until formulae without an explicitly specified reward model, i.e. a formula of the form `P=? [ "safe" Urew<3 "goal"]`. In case the reward model is not unique, an error is thrown.
This mirrors the behavior of the reward operator if no reward model is specified and simplifies property specification for models with unique reward models.
In addition, the relevant parser test case is extended.

### DIFF
--- a/src/storm/logic/TimeBoundType.h
+++ b/src/storm/logic/TimeBoundType.h
@@ -52,8 +52,7 @@ class TimeBoundReference {
     }
 
     bool hasRewardModelName() const {
-        assert(isRewardBound());
-        return static_cast<bool>(rewardName);
+        return isRewardBound() && static_cast<bool>(rewardName);
     }
 
     boost::optional<std::string> const& getOptionalRewardModelName() const {


src/storm/logic/TimeBoundType.h
```diff
@@ -11,7 +11,7 @@ enum class TimeBoundType { Steps, Time, Reward };
 
 class TimeBoundReference {
     TimeBoundType type;
-    std::string rewardName;
+    boost::optional<std::string> rewardName;
```
sjunges, line 14: I think we are trying to move away from boost::optional in favour of std::optional.

sjunges:
LGTM, but it would be good if @tquatmann could look at the parser :-)

src/storm/logic/TimeBoundType.h
```diff
@@ -41,7 +42,21 @@ class TimeBoundReference {
...
+        return rewardName.get();
+    }
+
     std::string const& getRewardName() const {
+        return getRewardModelName();
+    }
+
+    bool hasRewardModelName() const {
+        assert(isRewardBound());
+        return static_cast<bool>(rewardName);
```
tquatmann, line 56: This might be the more expected behaviour (I'm not sure?):
```suggestion
        return isRewardBound() && static_cast<bool>(rewardName);
```

tquatmann:
Parser looks good. Thanks!

AlexBork:
Thank you both for your comments. I have a more general question regarding boost::optional vs. std::optional use with respect to the parser. If I understand it correctly, the parser itself provides boost::optional when parsing optional parts, so at some point, in this case when the TimeBoundReference is created, we'd need to do a conversion to std::optional. Is this what we want to do? 
In addition, if I only change the added part to std::optional, we are inconsistent within the class itself as the rewardAccumulation still uses boost::optional, so I'd like to change that, too. My idea would be to stick to boost::optional for this PR, and in the future to make a pass over the whole parser and related classes to switch things over to std::optional at once. Thoughts?

src/storm/logic/TimeBoundType.h
AlexBork: Changed it

sjunges:
Hi,

that works! We generally discussed at some point that one big boost::optional -> std::optional change is currently not realistic, but I understand that you want to be consistent within the class right now. 

Best,
Sebastian

PR #536, opened 2024-05-19
LP solver support

sjunges:
This PR adds four features towards better support for LP based MDP model checking:

- Soplex version output in CMAKE
- Gurobi 11.0.2 support (which is the current version)
- Disabling gurobi tests from the command line to ensure that we can run CI even if we compile storm with support for Gurobi (but the CI will not have a valid license). 
- Dockerfile updates


src/test/storm_gtest.h
```diff
@@ -27,10 +27,19 @@ inline GTEST_API_ AssertionResult DoubleNearPredFormat(const char* expr1, const
 
 namespace storm {
 namespace test {
-inline void initialize() {
+extern bool noGurobi;
+
+inline void initialize(int* argc, char** argv) {
```
volkm, line 32: I had to lookup how gtest arguments are handled. Maybe add a comment:
`// GoogleTest-specific commandline arguments were already processed before and removed from argc/argv`

Dockerfile
```diff
@@ -19,6 +20,9 @@ ARG build_type=Release
 ARG no_threads=1
 # Specify CMake arguments for Storm
 ARG cmake_args="-DSTORM_PORTABLE=ON"
+# Can be ON/OFF
+ARG gurobi_support="ON"
```
volkm, line 24: I see the point of providing some convenience flags, but could we not use `cmake_args` instead? Then we have more flexibility to provide any CMake flags. We are also already using `cmake_args` to define the configurations in the CI.
The other option would be to add more on/off switches for all relevant configurations (spot, MathSAT, Cln, ...)

volkm (approved):
Thanks, LGTM. Only two smaller comments.

Dockerfile
sjunges: I dislike the default then, because you need to carefully mimic what is already in the default. But maybe we can add the arguments hardcoded and one may overwrite them (as far as i understand, cmake takes the rightmost definition as the actual one). What do you think?

volkm: We only need the PORTABLE flag as default argument to ensure platform independence. Should we simply hard-code this cmake flag and leave `cmake_args` empty? Then users can always provide their own flags.

sjunges: but how do we then set that the default is to include spot, soplex, etc in this build?

volkm: Ah okay, you want to have the dependencies enabled by default. Then I agree that it is better to have separate flags for it.

volkm: After this merge, I will create a new PR for some additional configurations: Spot, MathSat etc. I think it is better to have the changes to the Dockerfile in a separate PR because I should also adapt the deployed CI image then.

sjunges:
@volkm As this now seems to compile with Gurobi, we need to add the --noGurobi flag to the tests. How do we do that? 

I will make sure that the MILP test also listens to the noGurobi flag.

volkm:
One issue: ctest does not forward the arguments to googletest. We are currently using ctest to run all the tests and have a make target "test". There seems to be some [workaround](https://gitlab.kitware.com/cmake/cmake/-/issues/20470) for supporting arguments. Another possibility would be to define the extra arguments in cmake when defining the [googletest](https://cmake.org/cmake/help/latest/module/GoogleTest.html).

One additional small note: warnings on unknown arguments are currently not printed because the log level is set to ERROR.

sjunges:
Hmm. I am not so sure how to properly handle this then...

volkm:
Maybe not so nice, but we could add a CMake flag `DISABLE_GUROBI_TESTS` and use that instead of the `--nogurobi` cmdline flag. We loose flexibility because we have to set the flag at build time, but at least it is working.

Another idea could be to use [labels](https://gitlab.kitware.com/cmake/cmake/-/issues/18127) and then exclude tests based on the label.

volkm:
I think it looks good for now. It is a bit overkill to check the license for every test suite, but I agree that it is currently the easiest way to handle the Gurobi license.

PR #537, opened 2024-05-21
Added `--permute` option to re-order the states after building.

tquatmann:
This is useful to, e.g., investigate how much algorithm performance is affected by the order of the states.

Example:
```console

$ ./storm --qvbsroot ~/git/qcomp/benchmarks --qvbs consensus 5 steps_min  --permute bfs   | grep "Time for model checking" -B1
Result (for initial states): 768.0679927
Time for model checking: 0.201s.

$ ./storm --qvbsroot ~/git/qcomp/benchmarks --qvbs consensus 5 steps_min  --permute reverse-bfs   | grep "Time for model checking" -B1
Result (for initial states): 768.0918665
Time for model checking: 0.861s.

$  ./storm --qvbsroot ~/git/qcomp/benchmarks --qvbs consensus 5 steps_min  --permute random 1234   | grep "Time for model checking" -B1
Result (for initial states): 768.0830197
Time for model checking: 0.394s.

### DIFF
--- a/src/storm/transformer/StatePermuter.h
+++ b/src/storm/transformer/StatePermuter.h
@@ -1,0 +1,21 @@
+#include <cstdint>
+#include <memory>
+#include <vector>
+#include "storm/models/sparse/Model.h"
+
+namespace storm::transformer {
+
+/*!
+ * Applies the given permutation to the states of the given model.
+ * The permutation must be a permutation of the numbers 0, 1, ..., n-1 for n = model.getNumberOfStates().
+ * The state at position i of the input model will be moved to position permutation[i] in the output model.
+ * @tparam ValueType
+ * @param originalModel
+ * @param permutation
+ * @return
+ */
+template<typename ValueType>
+std::shared_ptr<storm::models::sparse::Model<ValueType>> permuteStates(storm::models::sparse::Model<ValueType> const& originalModel,
+                                                                       std::vector<uint64_t> const& permutation);
+
+}  // namespace storm::transformer
--- a/src/storm/utility/permutation.cpp
+++ b/src/storm/utility/permutation.cpp
@@ -1,0 +1,139 @@
+#include "storm/utility/permutation.h"
+
+#include <algorithm>
+#include <deque>
+#include <numeric>
+#include <random>
+
+#include "storm/adapters/RationalFunctionAdapter.h"
+#include "storm/adapters/RationalNumberAdapter.h"
+#include "storm/storage/BitVector.h"
+#include "storm/storage/SparseMatrix.h"
+#include "storm/utility/macros.h"
+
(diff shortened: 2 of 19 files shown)

### CODE REVIEW

File: src/storm/utility/permutation.cpp [hunk 1]
```diff
@@ -0,0 +1,139 @@
...
+    STORM_LOG_ASSERT(initialStates.size() == transitionMatrix.getRowGroupCount(), "Unexpected dimensions of initial states and transition matrix.");
+    STORM_LOG_ASSERT(initialStates.size() == transitionMatrix.getColumnCount(), "Unexpected dimensions of initial states and transition matrix.");
+    std::vector<index_type> permutation;
+    permutation.reserve(transitionMatrix.getRowGroupCount());
+
+    std::deque<index_type> stack(initialStates.begin(), initialStates.end());
+    storm::storage::BitVector discoveredStates = initialStates;
+    bool const useFifoStack = order == OrderKind::Bfs || order == OrderKind::ReverseBfs;
+
+    while (!stack.empty()) {
```
sjunges, line 85: dont we have graph algorithms which give BFS and DFS for you?

src/storm/transformer/StatePermuter.h
```diff
@@ -0,0 +1,21 @@
...
+ * Applies the given permutation to the states of the given model.
+ * The permutation must be a permutation of the numbers 0, 1, ..., n-1 for n = model.getNumberOfStates().
+ * The state at position i of the input model will be moved to position permutation[i] in the output model.
+ * @tparam ValueType
+ * @param originalModel
+ * @param permutation
+ * @return
+ */
+template<typename ValueType>
+std::shared_ptr<storm::models::sparse::Model<ValueType>> permuteStates(storm::models::sparse::Model<ValueType> const& originalModel,
```
sjunges, line 18: Can we add one test for this method?

sjunges:
LGTM, very cool functionality :-)

src/storm/utility/permutation.cpp
tquatmann: True, we have implemented these kind of BFS/DFS traversals at other places (for example [storm/utility/graph.cpp](https://github.com/moves-rwth/storm/blob/master/src/storm/utility/graph.cpp)).
However, I don't think we can easily get the permutation out of that.

We could eventually make something that is more modular and usable in all kind of graph algorithms. But imo this is something for a future refactor of graph.h

sjunges: Ah, I see, I thought you obtain a sorted list of states and then you compute the permuation from that?

tquatmann: If I see this correctly, that sorted list of states would already be the inverse of the permutation.

PR #538, opened 2024-05-21
remove some warnings

sjunges:


volkm (approved):
LGTM.
Only a minor suggestion for possible improvement.

volkm:
I tried to fix the compile issue and encountered the issue that the single-line version does not store information which state violates the assertion. In the interest of a better output, I will revert the change.
Sorry for the wrong suggestion.

PR #539, opened 2024-05-21
no nullpointer check necessary before free/delete[]

sjunges:
see https://stackoverflow.com/questions/18775608/free-a-null-pointer-anyway-or-check-first


volkm (approved):
LGTM

PR #545, opened 2024-05-22
Added configuration options to Dockerfile

volkm:
Allow to enable/disable several libraries/configs in the Dockerfile.
Also adapted the CI build configs to the new Dockerfile.


volkm:
Mathsat is currently provided via `cmake_args`. We should use the common format after https://github.com/moves-rwth/storm/issues/546 is addressed.

Issue #461, opened 2023-12-05
Use `maybe_unused` only if variable is not always unused


volkm:
Our general guideline is (thanks to @tquatmann):
- `[[maybe_unused]]` can be added whenever the parameter is used in some cases and unused in other cases (e.g. depending on compile time settings with `#if ...`). This case should be rather uncommon and actually avoidable most of the time
- If the parameter is always unused, the name can be dropped or commented out. This is decided on a case-by-case basis since (a) dropping the name potentially becomes annoying if the parameter should be used at some point and (b) keeping the name as a comment risks inconsistencies, e.g., when the parameter in other function overloads is renamed.

The use of `[[maybe_unused]]` in several files in `storm/solver/helper` should be replaced by commenting out/removing the parameters instead.

PR #547, opened 2024-05-22
Remove more warnings

AlexBork:
This removes more warnings in storm-pomdp. This includes both warnings when compiling with AppleClang on macOS using cmake option STORM_ALLWARNINGS and some unused parameter warnings occurring with GCC (as observed in the CI).

### DIFF
--- a/src/storm-pomdp/modelchecker/BeliefExplorationPomdpModelChecker.cpp
+++ b/src/storm-pomdp/modelchecker/BeliefExplorationPomdpModelChecker.cpp
@@ -17,7 +17,6 @@
 
 #include "storm/environment/Environment.h"
 #include "storm/exceptions/NotSupportedException.h"
-#include "storm/storage/Scheduler.h"
 #include "storm/utility/SignalHandler.h"
 #include "storm/utility/graph.h"
 #include "storm/utility/macros.h"
--- a/src/storm-pomdp/modelchecker/BeliefExplorationPomdpModelChecker.cpp
+++ b/src/storm-pomdp/modelchecker/BeliefExplorationPomdpModelChecker.cpp
@@ -1337,7 +1337,7 @@ void BeliefExplorationPomdpModelChecker<PomdpModelType, BeliefValueType, BeliefM
                     }
                 } else if (clipping.onGrid) {
                     // If the belief is not clippable, but on the grid, it may need to be explored, too
-                    bool inserted = beliefExplorer->addTransitionToBelief(action, successor.first, successor.second, false);
+                    beliefExplorer->addTransitionToBelief(action, successor.first, successor.second, false);
                 } else {
                     // Otherwise, the reward for all candidates is infinite, clipping does not make sense. Cut it off instead
                     absDelta += utility::convertNumber<BeliefValueType>(successor.second);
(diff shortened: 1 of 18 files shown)


src/storm-pomdp/transformer/ApplyFiniteSchedulerToPomdp.cpp
```diff
@@ -31,7 +31,7 @@ struct RationalFunctionConstructor {
 };
 
 template<typename ValueType>
-std::shared_ptr<RawPolynomialCache> getCache(storm::models::sparse::Pomdp<ValueType> const& model) {
+std::shared_ptr<RawPolynomialCache> getCache([[maybe_unused]] storm::models::sparse::Pomdp<ValueType> const& model) {
```
sjunges, line 34: In this instance, I would have expected to just dont name the model, see https://github.com/moves-rwth/storm/issues/461

sjunges:
LGTM. Many thanks for the swift action!

src/storm-pomdp/transformer/ApplyFiniteSchedulerToPomdp.cpp
AlexBork: I see your point. My idea to do it like that was that the specific instantiation with storm::RationalFunction requires the model parameter, so simply deleting it would not work. I could rework its usages?

volkm: I think Sebastian meant to only remove the variable name, either:
```
std::shared_ptr<RawPolynomialCache> getCache(storm::models::sparse::Pomdp<ValueType> const&) {
```
or 
```
std::shared_ptr<RawPolynomialCache> getCache(storm::models::sparse::Pomdp<ValueType> const& /*model*/) {
```

AlexBork: Ah, I actually didn't know that that's possible. Thanks, I will change it!

volkm: Thanks

src/storm-pomdp/modelchecker/BeliefExplorationPomdpModelChecker.cpp
```diff
@@ -1279,7 +1278,7 @@ bool BeliefExplorationPomdpModelChecker<PomdpModelType, BeliefValueType, BeliefM
     underApproximation->finishExploration();
     statistics.underApproximationBuildTime.stop();
     printUpdateStopwatch.stop();
-    STORM_PRINT_AND_LOG("Finished exploring under-approximation MDP.\nStart analysis...\n");
+    STORM_PRINT_AND_LOG("Finished exploring under-approximation MDP.\nStart analysis...\n")
```
tquatmann, line 1281: see other comment

src/storm-pomdp/modelchecker/BeliefExplorationPomdpModelChecker.cpp
```diff
@@ -1095,7 +1094,7 @@ bool BeliefExplorationPomdpModelChecker<PomdpModelType, BeliefValueType, BeliefM
                                        << " ##### " << underApproximation->getUnexploredStates().size() << " beliefs queued\n")
             if (underApproximation->getCurrentNumberOfMdpStates() > heuristicParameters.sizeThreshold && options.useClipping) {
                 STORM_PRINT_AND_LOG("##### Clipping Attempts: " << statistics.nrClippingAttempts.value() << " ##### "
-                                                                << "Clipped States: " << statistics.nrClippedStates.value() << "\n");
+                                                                << "Clipped States: " << statistics.nrClippedStates.value() << "\n")
```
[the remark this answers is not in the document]
AlexBork: STORM_PRINT and STORM_PRINT_AND_LOG do not require the ;
With STORM_ALLWARNINGS I get a warning about empty statements if I leave it in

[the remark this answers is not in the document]
tquatmann: Ahh, I see. Thanks!

volkm:
Small note regarding the semicolon: I think the issue comes from [here](https://github.com/moves-rwth/storm/blob/master/src/storm/utility/macros.h#L65). Should we remove the semicolons here?

AlexBork:
I would prefer to always require a semicolon, though I think just removing it there doesn't work (I did a quick test). I'll open an issue where we can potentially discuss this.

PR #548, opened 2024-05-22
a little less rational function dependencies

sjunges:


volkm (approved):
A little less conversation between header files is always a good idea. Thanks!

Issue #364, opened 2023-05-09
Soplex export location


sjunges:
As @volkm wrote:

We are not using add_imported_library in the CMake file for soplex. 

This would resolve one instance of anohter flaw:
https://github.com/moves-rwth/storm/issues/358

sjunges:
It turns out that we are adding it to the list of STORM_DEP_TARGETS, which should be the right behavior which we also have for other libraries.

Issue #358, opened 2023-04-26
Storm exports LOCATION-NOTFOUND


tquatmann:
My `stormConfig.cmake` file contained

```
set_target_properties(libsoplex-pic PROPERTIES IMPORTED_LOCATION "LOCATION-NOTFOUND")
```

which means that e.g. stormpy does not compile because it somehow can not link with a library named  `LOCATION-NOTFOUND`

We should somehow make sure that either this is not exported in the first place or at least detected on the stormpy side.

tquatmann:
Yes, this also solves the issue for me. I think there are two sides of this issue:

* Find out and fix why using soplex does this
* Generally make sure that "LOCATION-NOTFOUND" is not exported by any library

volkm:
We are not using `add_imported_library` in the CMake file for soplex. No idea whether this has anything to do with it, but I guess it does not hurt to add it and try whether it changes anything?

PR #549, opened 2024-05-22
Soplex cmake export fixed

sjunges:
The problem was that soplex export its location only for the release configuration and we were not handling that case. 

fixes https://github.com/moves-rwth/storm/issues/358. I think it also resolves https://github.com/moves-rwth/storm/issues/364

Whether we should have additional sanity checks on the created cmake config is another question. I think it is hard to do that really well, so i would rather not do that right now.


volkm:
Interesting behaviour to only export for the Release. Thanks for the fix. LGTM.

PR #552, opened 2024-05-23
Modify printing macros to require a trailing semicolon

### LINKED ISSUE #551: Inconsistent `STORM_PRINT` and `STORM_PRINT_AND_LOG` macros
The above-mentioned macros are not surrounded by `do{ ... } while(false)` as the other similar macros.
This is somewhat of a hack to enforce that macro invocations are terminated with a `;`. It seems that one can also just drop the `;` after `std::cout.flush()`.

The current state might trigger warnings when a `;` is added anyway, see for example the discussion on #547 

https://github.com/moves-rwth/storm/blob/f50c3282cc6642949b6f33868206c54aafed8a81/src/storm/utility/macros.h#L52-L73

AlexBork:
Adds wrapping `do{...} while` to STORM_PRINT and STORM_PRINT_AND_LOG macro definitions to require their uses to be followed by a semicolon. Also adjusts occurences where previously the semicolon was not present.

Resolves #551


volkm (approved):
LGTM.

tquatmann (approved):

PR #556, opened 2024-05-27
Integrate sin, cos and constants to Storm Jani parser

MarcoLm993:
This PR introduces support to the `sin` and `cos` operators, and to the PI and Euler number constants.

In particular, for the sin and cos operators I introduced a new expression type called `trigonometric`, that can be used later on in case other trigonometric operators will be integrated in storm. 

Please let me know if I missed any additional place where the aforementioned operators need to be defined.

### DIFF
--- a/src/storm-parsers/parser/JaniParser.cpp
+++ b/src/storm-parsers/parser/JaniParser.cpp
@@ -12,6 +12,7 @@
 #include "storm/storage/jani/TemplateEdge.h"
 #include "storm/storage/jani/expressions/JaniExpressions.h"
 #include "storm/storage/jani/visitor/CompositionInformationVisitor.h"
+#include "storm/storage/jani/visitor/JaniExpressionSubstitutionVisitor.h"
 
 #include "storm/storage/jani/types/ArrayType.h"
 #include "storm/storage/jani/types/BasicType.h"
@@ -847,7 +848,11 @@ std::pair<std::unique_ptr<storm::jani::JaniType>, storm::expressions::Type> Jani
                                 "Upper bound for bounded real variable " << variableName << "(scope: " << scope.description << ") must be numeric");
                 if (lowerboundExpr.isInitialized() && upperboundExpr.isInitialized() && !lowerboundExpr.containsVariables() &&
                     !upperboundExpr.containsVariables()) {
-                    STORM_LOG_THROW(lowerboundExpr.evaluateAsRational() <= upperboundExpr.evaluateAsRational(), storm::exceptions::InvalidJaniException,
+                    using SubMap = std::map<storm::expressions::Variable, storm::expressions::Expression>;
+                    storm::expressions::JaniExpressionSubstitutionVisitor<SubMap> transcendentalVisitor(SubMap(), true);
+                    const storm::RationalNumber lowerboundValue = transcendentalVisitor.substitute(lowerboundExpr).evaluateAsRational();
+                    const storm::RationalNumber upperboundValue = transcendentalVisitor.substitute(upperboundExpr).evaluateAsRational();
+                    STORM_LOG_THROW(lowerboundValue <= upperboundValue, storm::exceptions::InvalidJaniException,
                                     "Lower bound must not be larger than upper bound for bounded real variable " << variableName
                                                                                                                  << "(scope: " << scope.description << ").");
                 }
--- a/src/storm-parsers/parser/JaniParser.cpp
+++ b/src/storm-parsers/parser/JaniParser.cpp
@@ -125,7 +125,7 @@ std::pair<storm::jani::Model, std::vector<storm::jani::Property>> JaniParser<Val
     uint_fast64_t featuresCount = parsedStructure.count("features");
     STORM_LOG_THROW(featuresCount < 2, storm::exceptions::InvalidJaniException, "features-declarations can be given at most once.");
     if (featuresCount == 1) {
-        auto allKnownModelFeatures = storm::jani::getAllKnownModelFeatures();
+        const auto allKnownModelFeatures = storm::jani::getAllKnownModelFeatures();
         for (auto const& feature : parsedStructure.at("features")) {
             std::string featureStr = getString<ValueType>(feature, "Model feature");
             bool found = false;
--- a/src/storm-parsers/parser/JaniParser.cpp
+++ b/src/storm-parsers/parser/JaniParser.cpp
@@ -47,8 +47,8 @@ template<typename ValueType>
 const bool JaniParser<ValueType>::defaultVariableTransient = false;
 const std::string VARIABLE_AUTOMATON_DELIMITER = "_";
 template<typename ValueType>
-const std::set<std::string> JaniParser<ValueType>::unsupportedOpstrings({"sin",  "cos",  "tan",   "cot",   "sec",   "csc",   "asin", "acos",
-                                                                         "atan", "acot", "asec",  "acsc",  "sinh",  "cosh",  "tanh", "coth",
+const std::set<std::string> JaniParser<ValueType>::unsupportedOpstrings({"tan",  "cot",  "sec",   "csc",   "asin",  "acos",  "atan",
+                                                                         "acot", "asec", "acsc",  "sinh",  "cosh",  "tanh",  "coth",
                                                                          "sech", "csch", "asinh", "acosh", "atanh", "asinh", "acosh"});
 
 template<typename ValueType>
@@ -1333,6 +1333,16 @@ storm::expressions::Expression JaniParser<ValueType>::parseExpression(Json const
                 ensureNumericalType(arguments[0], opstring, 0, scope.description);
                 ensureNumericalType(arguments[1], opstring, 1, scope.description);
                 return storm::expressions::logarithm(arguments[0], arguments[1]);
+            } else if (opstring == "cos") {
+                arguments = parseUnaryExpressionArguments(expressionStructure, opstring, scope, returnNoneInitializedOnUnknownOperator, auxiliaryVariables);
+                assert(arguments.size() == 1);
+                ensureNumericalType(arguments[0], opstring, 0, scope.description);
+                return storm::expressions::cos(arguments[0]);
+            } else if (opstring == "sin") {
+                arguments = parseUnaryExpressionArguments(expressionStructure, opstring, scope, returnNoneInitializedOnUnknownOperator, auxiliaryVariables);
+                assert(arguments.size() == 1);
+                ensureNumericalType(arguments[0], opstring, 0, scope.description);
+                return storm::expressions::sin(arguments[0]);
             } else if (opstring == "aa") {
                 STORM_LOG_THROW(expressionStructure.count("exp") == 1, storm::exceptions::InvalidJaniException,
                                 "Array access operator requires exactly one exp (at " + scope.description + ").");
@@ -1430,6 +1440,16 @@ storm::expressions::Expression JaniParser<ValueType>::parseExpression(Json const
                     return storm::expressions::Expression();
                 }
                 STORM_LOG_THROW(false, storm::exceptions::InvalidJaniException, "Unknown operator " << opstring << " in " << scope.description << ".");
+            }
+        }
+        if (expressionStructure.count("constant") == 1) {
+            // Convert constants to a numeric value (only PI and Euler number, as in Jani specification)
+            const std::string constantStr = getString<ValueType>(expressionStructure.at("constant"), scope.description);
+            if (constantStr == "π") {
+                return expressionManager->rational(M_PI);
+            }
+            if (constantStr == "e") {
+                return expressionManager->rational(std::exp(1.0));
             }
         }
         STORM_LOG_THROW(
--- a/src/storm-parsers/parser/JaniParser.cpp
+++ b/src/storm-parsers/parser/JaniParser.cpp
@@ -1446,10 +1446,14 @@ storm::expressions::Expression JaniParser<ValueType>::parseExpression(Json const
             // Convert constants to a numeric value (only PI and Euler number, as in Jani specification)
             const std::string constantStr = getString<ValueType>(expressionStructure.at("constant"), scope.description);
             if (constantStr == "π") {
-                return expressionManager->rational(M_PI);
+                return std::make_shared<storm::expressions::TranscendentalNumberLiteralExpression>(
+                           *expressionManager, storm::expressions::TranscendentalNumberLiteralExpression::TranscendentalNumber::PI)
+                    ->toExpression();
             }
             if (constantStr == "e") {
-                return expressionManager->rational(std::exp(1.0));
+                return std::make_shared<storm::expressions::TranscendentalNumberLiteralExpression>(
+                           *expressionManager, storm::expressions::TranscendentalNumberLiteralExpression::TranscendentalNumber::E)
+                    ->toExpression();
             }
         }
         STORM_LOG_THROW(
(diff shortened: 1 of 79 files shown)


src/storm-parsers/parser/JaniParser.cpp
```diff
@@ -1432,6 +1442,16 @@ storm::expressions::Expression JaniParser<ValueType>::parseExpression(Json const
                 STORM_LOG_THROW(false, storm::exceptions::InvalidJaniException, "Unknown operator " << opstring << " in " << scope.description << ".");
             }
         }
+        if (expressionStructure.count("constant") == 1) {
```
sjunges, line 1445: Here, the expression now represents an approximation of the actual expression in the jani file. This is something to discuss a bit.

src/storm/storage/expressions/Expression.cpp
```diff
@@ -548,6 +548,18 @@ Expression logarithm(Expression const& first, Expression const& second) {
         second.getBaseExpressionPointer(), BinaryNumericalFunctionExpression::OperatorType::Logarithm)));
 }
 
+Expression cos(Expression const& first) {
+    return Expression(std::shared_ptr<BaseExpression>(new UnaryNumericalFunctionExpression(first.getBaseExpression().getManager(),
```
sjunges, line 552: Looking at e.g. https://github.com/moves-rwth/storm/blob/77ad64ed21225b3f5d91bb606796ec64eab8c279/src/storm/storage/expressions/Expression.cpp#L469 i would expect a type check here as well.

sjunges:
Hi Marco,

great!

- I think that it is debatable whether pi should be the same as 3.14.... I currently would argue that as an expression, it should be a function wihtout arguments (a symbol, that is) and only upon evaluating it, it should be interpreted with an approximation.  @tquatmann @volkm 

- It would be good if you could add some small tests, i.e., a small programme with sin/cos and pi/e and show that this gets properly parsed. 

- We probably do not handle model building for jani models with these constructs. Thus, I guess that the Jani Builders should get methods that check whether these ingredients are in the jani model and refuse working with them.  One approach here would be to use the model features of Jani. If trig is not added as a model feature, the parser should complain (or at least during a 'finalize' call) and the model builders should check whether this model feature is present. Let us know if you have any questions about this!

Best,
Sebastian

src/storm-parsers/parser/JaniParser.cpp
MarcoLm993: If I understand it correctly, you suggest to have something like this for PI (and the same for `e`):
```cpp
if (constantStr == "π") {
  return storm::expressions::pi();
}
```
Did I get it right?

sjunges: Indeed!

MarcoLm993:
Hi Sebastian,

thanks for the very quick feedback!

* Point 1, I replied in the initial comment, I think i understand your point.
* Point 2, I already have a small jani model to check that the new features are parsed correctly, I can add that in STORM, too, for the unit test!
* Point 3, I think I understood your point, I will try to have a look and reach out in case of doubts!

Best,
Marco

src/storm-parsers/parser/JaniParser.cpp
MarcoLm993: Hello,
I started looking into implementing this and I am not sure on what would be the preferred way to go.
My current understanding is that we need to have an additional class inherited from `BaseExpression`, that we can call `ConstantValueExpression` (or anything else), and this would provide an enum for the ConstantValue to store (i.e. PI and E), together with the same methods as for the `RationalLiteralExpression` class.

Taking this path means that some more boilerplate code needs to be written (for now I only found all the classes inherited from `ExpressionVisitor`, but maybe there is more), so I want to ask whether this is what you had in mind with your comment above, before I start implementing everything that is required.

sjunges: @tquatmann your input would be greatly appreciated here

src/storm/storage/expressions/UnaryNumericalFunctionExpression.cpp
```diff
@@ -96,6 +108,11 @@ std::shared_ptr<BaseExpression const> UnaryNumericalFunctionExpression::simplify
                 case OperatorType::Floor:
                 case OperatorType::Ceil:
                     break;
+                case OperatorType::Cos:
+                case OperatorType::Sin:
+                    // The operands expect a real value. Raise an exception
+                    STORM_LOG_THROW(false, storm::exceptions::InvalidOperationException, "Cannot apply cos or sin to an integer value.");
```
tquatmann, line 114: My understanding of the JANI-spec is that e.g. `sin(10)` would be valid in JANI as int is also a valid numeric type.

src/storm-parsers/parser/JaniParser.cpp
sjunges: I personally would indeed tend to use a ConstantValueExpression or maybe a RealLiteralExpression?

tquatmann: What you describe is the most correct and most complete way to implement this. However, unless someone sees an important use case for π and e outside of jani models, I suggest to reduce the mentioned boilerplate code a bit:

- Put your new expression class [here](https://github.com/moves-rwth/storm/tree/master/src/storm/storage/jani/expressions), where we collect all expressions that are somewhat special for JANI. Given the current naming scheme, the new class should have suffix `...LiteralExpression`, maybe `TranscendentalNumberLiteralExpression` ? 

- Extend the [JaniExpressionVisitor](https://github.com/moves-rwth/storm/blob/master/src/storm/storage/jani/visitor/JaniExpressionVisitor.h) and every class that inherits from this (fortunately, these are only a hand full and it's almost always clear what to do).

- Extend the `substituteConstantsInPlace` function of the JANI model, so that it can (optionally, controlled by an argument) approximate transcendentalNumberLiterals. Apply this substitution, before building the model.

The advantage of this approach is that you can still export the parsed jani without approximating the constants.
For example, when using `storm-conv --jani input.jani --constants N=4 --tojani output.jani`, the constants should still be present.

MarcoLm993: Indeed, having the new `...LiteralExpression` class in `jani/expressions` seems more appropriate for the current PR, thanks for your suggestion!

MarcoLm993: Hi @tquatmann ,

I just pushed a first attempt at integrating the `TranscendentalNumberLiteralExpression` in the JaniParser.
I am not sure this is 100% what you had in mind, since it resulted in an unexpected amount of changes around the code. Please let me know if there is anything that you think should be done differently.

Also, the unit tests and all the remaining comments will come at a later commit. This is just to make sure the changes related to this comment match your requests.

MarcoLm993: I just pushed some more changes I needed to do in order to have storm-conv working as you described.
Somehow this ended up touching an unexpected amount of files, I hope this is still fine with you, but of course I am open to any suggestion you might have.

In the next days I will address the last few comments you had (namely a missing type check and accepting integers as argument of trigonometric operators).

MarcoLm993:
Hello @sjunges and @tquatmann ,

I just pushed the last changes related to this PR. Now it should be good to go, if you do not spot anything else to change. :)

Looking forward to your inputs.

Best,
Marco

src/storm/storage/jani/Model.cpp
```diff
@@ -1100,52 +1100,52 @@ Model& Model::replaceUnassignedVariablesWithConstants() {
...
 
 Model Model::substituteConstants() const {
     Model result(*this);
     result.replaceUnassignedVariablesWithConstants();
-    result.substituteConstantsInPlace();
+    result.substituteConstantsInPlace(false);
     return result;
 }
 
 Model Model::substituteConstantsFunctions() const {
```
MarcoLm993, line 1145: I added to this method the substitution of Transcendental numbers with Rational numbers, too.
Do you think it makes sense to reflect that in the name of the method, too? Something like `substituteConstantsFunctionsTranscendentals`?

src/storm/generator/JaniNextStateGenerator.cpp
```diff
@@ -156,6 +158,7 @@ bool JaniNextStateGenerator<ValueType, StateType>::canHandle(storm::jani::Model
     features.remove(storm::jani::ModelFeature::DerivedOperators);
     features.remove(storm::jani::ModelFeature::Functions);  // can be substituted
     features.remove(storm::jani::ModelFeature::StateExitRewards);
+    features.remove(storm::jani::ModelFeature::TrigonometricFunctions);
```
sjunges, line 161: Is this true? If so, I think this should be covered in a test. 
If you are not sure, I would not allow this right now?

src/storm/storage/expressions/Type.h
```diff
@@ -187,7 +196,7 @@ class IntegerType : public BaseType {
     virtual bool isIntegerType() const override;
 
    private:
-    static const uint64_t mask = (1ull << 62);
+    static const uint64_t mask = (1ull << 58);
```
sjunges, line 199: Is there a reason these have changed?

src/storm/storage/jani/TemplateEdgeDestination.cpp
```diff
@@ -15,8 +15,9 @@ TemplateEdgeDestination::TemplateEdgeDestination(std::vector<Assignment> const&
     // Intentionally left empty.
 }
 
-void TemplateEdgeDestination::substitute(std::map<storm::expressions::Variable, storm::expressions::Expression> const& substitution) {
-    assignments.substitute(substitution);
+void TemplateEdgeDestination::substitute(std::map<storm::expressions::Variable, storm::expressions::Expression> const& substitution,
+                                         bool const& substituteTranscendentalNumbers) {
```
sjunges, line 19: bool by const& triggers a warning on my system

src/storm/generator/JaniNextStateGenerator.cpp
MarcoLm993: Hello Sebastian,

I am using it in the smc extension we are developing in the CONVINCE, so I am quite convinced this is working.
However, having one more test never hurts! :)

Some references about the smc use case:
- test: https://github.com/convince-project/smc_storm/blob/main/test/test_models.cpp#L58
- where this is used: https://github.com/convince-project/smc_storm/blob/main/src/model_checker/state_generation.cpp#L42

src/storm/storage/jani/TemplateEdge.cpp
```diff
@@ -85,15 +85,16 @@ OrderedAssignments& TemplateEdge::getAssignments() {
...
 
-void TemplateEdge::substitute(std::map<storm::expressions::Variable, storm::expressions::Expression> const& substitution) {
-    guard = substituteJaniExpression(guard, substitution);
+void TemplateEdge::substitute(std::map<storm::expressions::Variable, storm::expressions::Expression> const& substitution,
+                              bool const& substituteTranscendentalNumbers) {
+    guard = substituteJaniExpression(guard, substitution, substituteTranscendentalNumbers);
 
     for (auto& assignment : assignments) {
-        assignment.substitute(substitution);
+        assignment.substitute(substitution, substituteTranscendentalNumbers);
```
sjunges, line 93: is this maybe something that can be avoided by having a sensible default for this?

src/storm/generator/JaniNextStateGenerator.cpp
sjunges: If you want to make sure that we dont break this, a test definitively does not hurt :-)

src/storm/storage/expressions/Type.h
MarcoLm993: Yes, it was required for the type comparison here implemented [here](https://github.com/moves-rwth/storm/pull/556/files/af3337abf9f01100c009d5e38ee5867ceafd84b9#diff-7a2243d7193a5ddf87c896ae38a351a8daf446c1622176d0c5d9d29e1c43c3ddR301): to ensure that a transcendental number type is considered a non-integer number by the various model visitors, its mask must evaluate to a value higher than the one inn IntegerType.
Since in this implementation the mask is generated by setting a single bit in a uint64 variable, RationalType was getting the highest possible value, and IntegerType the one right before that. For this reason, I had to shift all of them to make space for the newly implemented TranscendentalNumberType.

src/storm/storage/jani/TemplateEdge.cpp
MarcoLm993: I am not sure, since this is set in the JaniModel's methods to substitute constants (around [here](https://github.com/moves-rwth/storm/pull/556/files/af3337abf9f01100c009d5e38ee5867ceafd84b9#diff-1271277540852224a7a247ef9c5259a6f3a3bbde0f1d44ba862c8e0f7c9ecb36R1133)) and must be propagated down to the various entries of the model. The value depends whether we want to preserve the constants (e.g. storm-conv) or not (when we run the evaluation).
A way to not have the this argument would be to set a static variable for the jani visitor, to signal whether we want to substitute transcendental numbers or not, but I wouldn't recommend it, since it makes the code execution less clear.

src/storm/storage/jani/TemplateEdgeDestination.cpp
MarcoLm993: I can remove the reference. On my system quite some warning are triggered, so it is hard to notice new ones.
Are there more warnings that were added by this PR?

MarcoLm993: I had a look in the repository and `bool const&` seems to be a common pattern for function definitions in STORM.

Of course I can switch the ones I added to be `bool const`, by dropping the reference, but maybe it would make sense to make that consistent throughout the whole repository. The same could be done for all other basic types (int, float, double, char)

src/storm/generator/JaniNextStateGenerator.cpp
MarcoLm993: I adapted the test_trigonometry jani file to not expand forever and added a test for it in the `ExplicitJaniModelBuilderTest` gtest, (that is the only place I found with similar tests).
The new unit test can be found in this commit: https://github.com/moves-rwth/storm/pull/556/commits/6b54f7c5a1b5e01fb5ab0c645bfde906d4a87d03
On doing it, I found that transcendental numbers were not handled correctly in the constant variables bounds, but now they are handled, too.

src/storm/storage/jani/TemplateEdgeDestination.cpp
MarcoLm993: I changed all declarations of `bool const& substituteTranscendentalNumbers` to not pass the value by reference anymore here: https://github.com/moves-rwth/storm/pull/556/commits/6f016ca1cb72ee0eb2a64cfc97c43f2f33150df0
On my machine, this seems not to throw warnings anymore.

src/storm/storage/jani/types/BoundedType.cpp
```diff
@@ -66,10 +66,10 @@ storm::expressions::Expression const& BoundedType::getUpperBound() const {
 void BoundedType::substitute(std::map<storm::expressions::Variable, storm::expressions::Expression> const& substitution) {
```
tquatmann, line 66: Why does this not pass a `bool substituteTranscendentalNumbers` ?

tquatmann (commented):
Thanks for doing all this! 👍 

There are some rather minor comments left. Apart from that, this is looking good!

I believe we could eventually replace all these `substitute` methods by a `JaniTraverser` but that is something for a different PR.

MarcoLm993 (commented):
Hi, thanks for your review.
I addressed the comments from @tquatmann and pushed them in the last commit, and I added one comment about the bit shift for the Type maks.
Let me know what you think about it.

tquatmann (approved):
Thanks for looking into my comments. To me, this PR looks ready to be merged!

src/storm/storage/expressions/Type.h
tquatmann: Thanks for clarifying. I just wanted to make sure that I didn't miss some detail concerning the unused masks.

sjunges:
Many thanks for all the work.

MarcoLm993:
I am the one thanking you for the discussions and for looking into this :)

PR #562, opened 2024-05-29
CI to create Docker releases

volkm:
Add Github Action which builds and deploy Docker images for a given tag (e.g. `1.2.3` or `stable`).


sjunges:
LGTM

PR #563, opened 2024-05-31
Follow general output precision in DFT approximation

volkm:
- The DFTModelChecker specified its own precision instead of following the default.
- Made `setOutputDigits` private.


sjunges:
LGTM!

PR #566, opened 2024-06-04
Update shipped glpk to version 5.0

tquatmann:


sjunges:
Does it make sense to remove some of the doc files, in particular pdfs etc to keep our repo small where possible?

tquatmann:
Good point, I removed the `doc` and the `examples` folder, which are both not needed for building.

PR #568, opened 2024-06-07
Improved LpMinMaxLinearEquationSolver, set relevant values in topo solvers

tquatmann:
* Added options to tweak the LP encoding
* Use RawMode instead of storm::expressions
* Add ViToLp Method to warm-start the procedure
* Topological solvers can now (optionally, if enabled) set relevant values for each SCC

See [this](https://doi.org/10.1007/978-3-031-30823-9_24) publication for further details on some of these improvements.


sjunges:
This looks good! Should we also have a vi-to-lp environment in our tests here: https://github.com/moves-rwth/storm/blob/6d0e4f415028413330f223cf081cdfc320bbb3dc/src/test/storm/modelchecker/prctl/mdp/MdpPrctlModelCheckerTest.cpp#L231

PR #571, opened 2024-06-12
Fix assertion in SparseMatrix

volkm:
Assertion was violated for empty matrix.

### DIFF
--- a/src/storm/storage/SparseMatrix.cpp
+++ b/src/storm/storage/SparseMatrix.cpp
@@ -2382,14 +2382,24 @@ typename SparseMatrix<ValueType>::rows SparseMatrix<ValueType>::getRowGroup(inde
 
 template<typename ValueType>
 typename SparseMatrix<ValueType>::const_iterator SparseMatrix<ValueType>::begin(index_type row) const {
-    STORM_LOG_ASSERT(row < this->getRowCount() || row == 0, "Row " << row << " exceeds row count " << this->getRowCount() << ".");
+    STORM_LOG_ASSERT(row < this->getRowCount(), "Row " << row << " exceeds row count " << this->getRowCount() << ".");
     return this->columnsAndValues.begin() + this->rowIndications[row];
 }
 
 template<typename ValueType>
 typename SparseMatrix<ValueType>::iterator SparseMatrix<ValueType>::begin(index_type row) {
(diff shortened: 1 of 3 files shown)


sjunges:
Is it on purpose that I can now iterate over row 20 if there are no rows, but i cannot iterate over row 20 if there are 5 rows?

volkm:
Good point. I revised the assertion.

src/storm/storage/SparseMatrix.cpp
```diff
@@ -2382,22 +2382,25 @@ typename SparseMatrix<ValueType>::rows SparseMatrix<ValueType>::getRowGroup(inde
...
 
 template<typename ValueType>
 typename SparseMatrix<ValueType>::iterator SparseMatrix<ValueType>::begin(index_type row) {
+    STORM_LOG_ASSERT(row < this->getRowCount() || row == 0, "Row " << row << " exceeds row count " << this->getRowCount() << ".");
     return this->columnsAndValues.begin() + this->rowIndications[row];
 }
 
 template<typename ValueType>
 typename SparseMatrix<ValueType>::const_iterator SparseMatrix<ValueType>::end(index_type row) const {
+    STORM_LOG_ASSERT(row < this->getRowCount(), "Row " << row << " exceeds row count " << this->getRowCount() << ".");
```
sjunges, line 2397: for empty matrices, you cannot call end now?

sjunges:
What is the use case to be able to iterate over a non-existing row?

volkm:
In my view, it should be possible to create an empty matrix. But then the new test in this PR fails, because in the constructor we call `updateNonzeroEntryCount` which tries to iterate over all entries.
My goal is to support the iterators for empty matrices as well, such that for example `begin() != end()`  always works.

sjunges:
I agree that begin() should work, but I dont know whether begin(0) needs to work?

volkm:
I think we need to support 0 because it is the default argument for `begin()`.

sjunges:
I see. Then I am ok with this solution, but would personally have considered seperating the implementation of the functions without argument.

The comment about the "end" code still stands? As in, shouldnt one be able to also call end(0) now?

tquatmann:
I also think that `begin()` and `begin(index_type row)` should be two different overloads. Then, `begin(i)` is only valid iff row `i` exists and its the same for `end`.

volkm:
Thanks for your comments. I decoupled `begin()` and `begin(row)`. This is also now consistent with `end` which already had both these functions.

sjunges:
LGTM

PR #573, opened 2024-07-01
Misc solver fixes

tquatmann:
This PR provides minor fixes related to solvers:

- gmmxx settings: removed jacobi method from possible options
- Rational Search always requires a unique solution
- VI2PI: configure VI environment, flag requirements of VI solver as checked


volkm:
LGTM

PR #574, opened 2024-07-01
Fixed incorrectly flagging terminal states as 'unexplored'. 

### LINKED ISSUE #559: "Unexplored" label is always added
The "unexplored" label introduced in #521 seems to be added even if all states are explored.

For the command
```
./bin/storm --prism ../resources/examples/testfiles/dtmc/die.pm
```
the resulting model still has the "unexplored" label:
```
Model type: 	DTMC (sparse)
States: 	13
Transitions: 	20
Reward Models:  none
State Labels: 	3 labels
   * unexplored -> 0 item(s)
   * deadlock -> 0 item(s)
   * init -> 1 item(s)
Choice Labels: 	none
```

Or is the "unexplored" label supposed to always be present (like "init" and "deadlock")?

tquatmann:
Fixes #559


volkm (approved):
LGTM. Thanks.

PR #575, opened 2024-07-02
Removed RewardMeasureType

### LINKED ISSUE #541: What do we do with rewardMeasureType
In my quest for https://github.com/moves-rwth/storm/issues/154
We have some occurences of rewardMeasureType but I dont believe that this is currenlty really supported.

volkm:
Resolves #541

While going through the files I also replaced `ifndef` by `pragma once` and applied autoformat on the include orders.


sjunges:
LGTM

PR #578, opened 2024-07-07
Fixed SparseModelMemoryProduct with transition rewards and nondetermism

tquatmann:
We currently require that the row grouping of the transition **probability** matrix is also present in the transition **reward** matrices.

Specifically, we [assert](https://github.com/moves-rwth/storm/blob/58e174b11158ece3bb6b05339ba4759e32d475a2/src/storm/models/sparse/Model.cpp#L76) that `rewardMatrix.isSubmatrixOf(transitionMatrix)` is true when constructing a model and `isSubmatrixOf` checks whether the row grouping is identical.


volkm (approved):
LGTM

PR #580, opened 2024-07-10
Fixed issue in DFT symmetry computation.

volkm:
Added compatibility check for corresponding colour class sizes.

Fixes https://github.com/moves-rwth/stormpy/issues/172


sjunges:
LGTM!

PR #581, opened 2024-07-11
Moved Storm main binary to storm-cli folder

volkm:
Separated storm library and storm binary.

Smaller changes: made CMake files consistent and removed an old workaround.


sjunges:
Great, this simplifies some modifications that I wanted to do in the future!!

PR #584, opened 2024-07-23
Use 4 threads in CI

volkm:
Increased number of Github runners from 2 to 4. Leads to faster CI runs (see [old](https://github.com/moves-rwth/storm/actions/runs/10026195486/usage) and [new](https://github.com/volkm/storm/actions/runs/10056576731/usage) timings).


sjunges:
LGTM! Great!

PR #585, opened 2024-07-23
Revised number of build jobs for resources

volkm:
The multi-threaded build lead to build issues in cudd for me (Docker on M3): `make resources -j2` lead to the error `write jobserver: Bad file descriptor`. Single-threaded build with `make resources -j1` was successful. I set the number of threads for cudd to 1 now.

For some other resources, I enabled multi-threaded build.


sjunges:
I dont really understand where this bug comes from, but the fix seems fine :-)

volkm:
I also have no real idea what the root cause of this bug is. I searched a bit but could not find anything meaningful apart from disabling multi-threaded build. So that is what we do now :).

PR #598, opened 2024-08-23
Bump build system

### LINKED ISSUE #286: Support for C++20
As a mid-term goal, it would be good to check whether storm can be compiled with C++20. 

Known problems:
- [x] ModernJson, see https://github.com/moves-rwth/storm/issues/284
- [ ] Run tests

volkm:
- Use C++20 and fixed some includes (resolves #286)
- Bump CMake to 3.22
- Removed support for Debian 11 and Ubuntu 20.04
- Regularly enable CLN and Ginac again for Apple Silicon

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required (VERSION 3.22)
 
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
 
 # Set project name
 project (storm CXX C)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,39 +157,30 @@ set(STORM_TEST_RESOURCES_DIR "${PROJECT_SOURCE_DIR}/resources/examples/testfiles
 # Auto-detect operating system.
 set(MACOSX 0)
 set(LINUX 0)
+set(APPLE_SILICON 0)
 if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
 	# Mac OS
 	set(OPERATING_SYSTEM "Mac OS")
         set(MACOSX 1)
+	if(${CMAKE_SYSTEM_PROCESSOR} MATCHES arm64)
+		set(APPLE_SILICON 1)
+	endif()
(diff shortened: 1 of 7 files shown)


sjunges:
This is maybe more general than this PR, but should we indeed still rely on CLN and GiNaC where possible?

volkm:
Do we have an idea of the performance difference when using CLN or GMP?

volkm:
For the files `ExplicitQualitativeCheckResult.cpp` and `ExplicitQuantitativeCheckResult.cpp` I moved the include `RationalNumberAdapter.h` before the usual header file in order for the compilation to be successful. I have no idea why the include order is relevant here. Feel free to suggest a better fix.

CMakeLists.txt
```diff
@@ -157,39 +157,30 @@ set(STORM_TEST_RESOURCES_DIR "${PROJECT_SOURCE_DIR}/resources/examples/testfiles
...
-    set(LINUX 1)
+	message(WARNING "We are unsure about your operating system.")
+	set(OPERATING_SYSTEM "Linux")
+	set(LINUX 1)
 ENDIF()
 message(STATUS "Storm - Detected operating system ${OPERATING_SYSTEM}.")
 
 set(SHIPPED_CARL_USE_CLN_NUMBERS ON)
 set(SHIPPED_CARL_USE_GINAC ON)
-# Detect Apple Silicon and adjust settings
```
sjunges, line 180: I think we should still emit a warning, given that we still have various bugs on apple silicon....

src/storm/modelchecker/results/ExplicitQualitativeCheckResult.cpp
```diff
@@ -1,4 +1,6 @@
-#include "storm/modelchecker/results/ExplicitQualitativeCheckResult.h"
+#include "storm/adapters/RationalNumberAdapter.h"  // Must come first
+
+#include "ExplicitQualitativeCheckResult.h"
```
sjunges, line 3: Please prefix this with storm/modelchecker/results as this will otherwise soon not compile anymore

PR #605, opened 2024-08-28
Fixed issues with DFT representant computation

volkm:
Representants were not properly updated when the element was part of a spare module but also of the top module.
Also added a test case triggering this issue.


sjunges:
LGTM

PR #612, opened 2024-09-03
Fix handling of SMGs with non-synchronizing but labeled actions.

### LINKED ISSUE #609: Error when using action labels to specify players in SMGs
When specifying the players in an SMG using action labels, storm incorrectly complains, that there are modules with unlabeled commands. Prism games can correctly handle this situation.

Steps for reproducing:

```
action_labels.prism
===================

smg

global state : [0..2];

player p1 [a1] endplayer
player p2 [a2] endplayer
player p3 [a3] endplayer

module m1
	[a1] state=0 -> 1/2 : (state'=1) + 1/2 : (state'=2);
endmodule

module m2
	[a2] state=1 -> 1/2 : (state'=0) + 1/2 : (state'=2);
endmodule

module m3
	[a3] state=2 -> 1/2 : (state'=0) + 1/2 : (state'=1);
endmodule
```

Call storm with:
```bash
./storm --prism action_labels.prism
```

Output:
```
Storm 1.8.1

Date: Fri Aug 30 17:24:02 2024
Command line arguments: --prism action_labels.prism
Current working directory: /home/jannik/ba/storm_test

Time for model input parsing: 0.166s.

ERROR (PrismNextStateGenerator.cpp:663): Module m1 is not owned by any player but has at least one enabled, unlabeled command.
ERROR (storm-cli.cpp:63): An exception caused Storm to terminate. The message of the exception is: WrongFormatException: Module m1 is not owned by any player but has at least one enabled, unlabeled command.
```

tquatmann:
Previously, all non-synchronizing commands were  assigned to the player owning the module, even if the command has an action label. However, the correct behavior (as far as I understand the specification) is that only unlabeled commands are associated to the module player.

Relevant snippet from the [PRISM-games website](https://www.prismmodelchecker.org/games/modelling.php#smgs): 
> For a turn-based game, it needs to be specified which player controls each state. In PRISM-games, this is done by proving a list of action names and/or module names for each player. This player then controls any state where all the transitions available are either labelled with one of these actions or are unlabelled and belong to one of these modules

Fixes #609

### DIFF
--- a/resources/examples/testfiles/smg/action_labels.prism
+++ b/resources/examples/testfiles/smg/action_labels.prism
@@ -1,0 +1,19 @@
+smg
+
+global state : [0..2];
+
+player p1 [a1] endplayer
+player p2 [a2] endplayer
+player p3 [a3] endplayer
+
+module m1
+	[a1] state=0 -> 1/2 : (state'=1) + 1/2 : (state'=2);
+endmodule
+
+module m2
+	[a2] state=1 -> 1/2 : (state'=0) + 1/2 : (state'=2);
+endmodule
+
+module m3
+	[a3] state=2 -> 1/2 : (state'=0) + 1/2 : (state'=1);
+endmodule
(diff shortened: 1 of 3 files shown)


sjunges:
LGTM

PR #617, opened 2024-09-05
Extended handling of terminal states for POMDPs

### LINKED ISSUE #603: Missing handling of terminal states for POMDPs
States that are marked as terminal while building a POMDP model (for example due to violating the left-hand side of an Until) are not handled appropriately. By default, for terminal states, only one self-loop without action label is added. This, however, may lead to violation of the requirement that states with the same observation must have the same actions enabled.

This can be worked around with by using option ``--buildfull``, but terminal states should be handled properly to reduce model building times.

This issue is based on the discussion on issue #48.

AlexBork:
This PR extends the handling of states marked as terminal in the PrismNextStateGenerator. In particular, terminal states in POMDPs are expanded with self-loops for all enabled actions, preventing erroneous behavior if there are other states with the same observation as a terminal state present in the model.

Resolves #603


tquatmann:
I don't like the code duplication introduced here.
However, I see that this is the lesser evil as the alternatives would be

- doing this "properly";  requiring a more involved rewrite of the PrismNextStateGenerator or
- ignoring the issue

LGTM

tquatmann (approved):

PR #619, opened 2024-09-09
Re-introduce sampling of the derivative into CLI

linusheck:
When we changed storm-pars's CLI interface, we accidentally removed the derivative sampling option. I reintroduced it under the "sampling" mode:

```
dtmc

const p1;
const p2;
const p3;

// this isnt simple :)
module test
	s : [0..5] init 0;
	
	[] s=0 -> p1 : (s'=1) + p2 : (s'=2) + 1-p1-p2 : (s'=3);
	[] s=1 -> p1 : (s'=2) + p2 : (s'=3) + 1-p1-p2 : (s'=4);
	[] s=2 -> 0.5 : (s'=0) + 0.5 : (s'=5);
	[] s=3 -> 0.5 : (s'=1) + 0.5 : (s'=5);
	[] s=4 -> 1 : (s'=4);
	[] s=5 -> 1 : (s'=5);
endmodule

label "target" = s=4;
```

```
storm-pars --prism non_graph_preserving.pm -prop "P=? [F \"target\"]" --mode sampling  --samples "p1=0.1,p2=0.1" --sample-derivative
```

=>

```
Model checking property "1": P=? [F "target"] ...
Derivative w.r.t. p1:
Result (initial states) for instance [p1=1/10, p2=1/10]: 0.02889586935

Derivative w.r.t. p2:
Result (initial states) for instance [p1=1/10, p2=1/10]: -0.5451254454

Overall time for sampling all instances: 0.000s
```


src/storm-pars/settings/modules/SamplingSettings.cpp
```diff
@@ -25,6 +26,7 @@ SamplingSettings::SamplingSettings() : ModuleSettings(moduleName) {
                                                    "Sets whether it can be assumed that the samples are graph-preserving.")
                         .build());
     this->addOption(storm::settings::OptionBuilder(moduleName, sampleExactOptionName, false, "Sets whether to sample using exact arithmetic.").build());
+    this->addOption(storm::settings::OptionBuilder(moduleName, sampleDerivativeOptionName, false, "Sets whether to sample the derivatives instead..").build());
```
sjunges, line 29: Don't we always solve the system, also when computing derivatives? So is the "instead" maybe confusing / unnecessary?

sjunges:
Great, thanks!

src/storm-pars/settings/modules/SamplingSettings.cpp
linusheck: ok will change it!

sjunges:
The code still does not report the actual model checking results on the CLI, does it?

glatteis:
> The code still does not report the actual model checking results on the CLI, does it?

Nope :D didn’t interpret your first comment like that. Will change it so it also reports the original results :)

sjunges:
LGTM!

PR #621, opened 2024-09-10
Support for discounted properties in MDPs and DTMCs

AlexBork:
This PR adds support for checking cumulative and total reward properties with a given discount factor on MDPs and DTMCs.
- Extends the parser to allow discounted properties, proposed syntax is ``R=? [Cdiscount=_factor_]``
- Introduces helper classes for checking discounted properties
- Extends the respective PCTL model checking classes with support for discounted cumulative and total reward properties, implemented analogous to the existing undiscounted implementations

### DIFF
--- a/src/storm/logic/DiscountedCumulativeRewardFormula.h
+++ b/src/storm/logic/DiscountedCumulativeRewardFormula.h
@@ -1,0 +1,78 @@
+#pragma once
+
+#include "storm/logic/PathFormula.h"
+
+#include "storm/logic/TimeBound.h"
+#include "storm/logic/TimeBoundType.h"
+
+namespace storm {
+namespace logic {
+class DiscountedCumulativeRewardFormula : public PathFormula {
+   public:
+    DiscountedCumulativeRewardFormula(storm::expressions::Expression const discountFactor, TimeBound const& bound,
+                                      TimeBoundReference const& timeBoundReference = TimeBoundReference(TimeBoundType::Time),
+                                      boost::optional<RewardAccumulation> rewardAccumulation = boost::none);
+    DiscountedCumulativeRewardFormula(storm::expressions::Expression const discountFactor, std::vector<TimeBound> const& bounds,
+                                      std::vector<TimeBoundReference> const& timeBoundReferences,
+                                      boost::optional<RewardAccumulation> rewardAccumulation = boost::none);
+
+    virtual ~DiscountedCumulativeRewardFormula() = default;
+
+    virtual bool isDiscountedCumulativeRewardFormula() const override;
+    virtual bool isCumulativeRewardFormula() const override;
+    virtual bool isRewardPathFormula() const override;
+
+    bool isMultiDimensional() const;
+    unsigned getDimension() const;
+
+    virtual boost::any accept(FormulaVisitor const& visitor, boost::any const& data) const override;
+
+    virtual void gatherReferencedRewardModels(std::set<std::string>& referencedRewardModels) const override;
+    virtual void gatherUsedVariables(std::set<storm::expressions::Variable>& usedVariables) const override;
+
+    virtual std::ostream& writeToStream(std::ostream& out, bool allowParentheses = false) const override;
+
+    TimeBoundReference const& getTimeBoundReference() const;
+    TimeBoundReference const& getTimeBoundReference(unsigned i) const;
+
+    bool isBoundStrict() const;
+    bool isBoundStrict(unsigned i) const;
+    bool hasIntegerBound() const;
+    bool hasIntegerBound(unsigned i) const;
+
+    storm::expressions::Expression const& getBound() const;
+    storm::expressions::Expression const& getBound(unsigned i) const;
+
+    storm::expressions::Expression const& getDiscountFactor() const;
+
+    template<typename ValueType>
+    ValueType getBound() const;
+
+    template<typename ValueType>
+    ValueType getBound(unsigned i) const;
+
+    template<typename ValueType>
+    ValueType getDiscountFactor() const;
+
+    template<typename ValueType>
+    ValueType getNonStrictBound() const;
+
+    std::vector<TimeBound> const& getBounds() const;
+
+    bool hasRewardAccumulation() const;
+    RewardAccumulation const& getRewardAccumulation() const;
+    std::shared_ptr<DiscountedCumulativeRewardFormula const> stripRewardAccumulation() const;
+
+    std::shared_ptr<DiscountedCumulativeRewardFormula const> restrictToDimension(unsigned i) const;
+
+   private:
+    static void checkNoVariablesInBound(storm::expressions::Expression const& bound);
+    static void checkNoVariablesInDiscountFactor(storm::expressions::Expression const& factor);
+
+    storm::expressions::Expression const discountFactor;
+    std::vector<TimeBoundReference> timeBoundReferences;
+    std::vector<TimeBound> bounds;
+    boost::optional<RewardAccumulation> rewardAccumulation;
+};
+}  // namespace logic
+}  // namespace storm
--- a/src/storm/modelchecker/helper/DiscountingHelper.cpp
+++ b/src/storm/modelchecker/helper/DiscountingHelper.cpp
@@ -1,0 +1,146 @@
+#include "DiscountingHelper.h"
+#include "storm/environment/solver/MinMaxSolverEnvironment.h"
+#include "storm/exceptions/IllegalFunctionCallException.h"
+#include "storm/solver/helper/DiscountedValueIterationHelper.h"
+#include "storm/solver/helper/SchedulerTrackingHelper.h"
+#include "storm/storage/SparseMatrix.h"
+
+namespace storm {
+namespace modelchecker {
+namespace helper {
+
+template<typename ValueType, bool TrivialRowGrouping>
+DiscountingHelper<ValueType, TrivialRowGrouping>::DiscountingHelper(storm::storage::SparseMatrix<ValueType> const& A, ValueType discountFactor)
+    : localA(nullptr), A(&A), discountFactor(discountFactor) {
+    progressMeasurement = storm::utility::ProgressMeasurement("iterations");
+    storm::storage::SparseMatrix<ValueType> discountedMatrix(A);
+    discountedMatrix.scaleRowsInPlace(std::vector<ValueType>(discountedMatrix.getRowCount(), discountFactor));
+    discountedA = discountedMatrix;
+}
+
+template<typename ValueType, bool TrivialRowGrouping>
+DiscountingHelper<ValueType, TrivialRowGrouping>::DiscountingHelper(storm::storage::SparseMatrix<ValueType> const& A, ValueType discountFactor,
+                                                                    bool trackScheduler)
+    : localA(nullptr), A(&A), discountFactor(discountFactor), trackScheduler(trackScheduler) {
+    progressMeasurement = storm::utility::ProgressMeasurement("iterations");
+    storm::storage::SparseMatrix<ValueType> discountedMatrix(A);
+    discountedMatrix.scaleRowsInPlace(std::vector<ValueType>(discountedMatrix.getRowCount(), discountFactor));
+    discountedA = discountedMatrix;
+}
+
+template<typename ValueType, bool TrivialRowGrouping>
+void DiscountingHelper<ValueType, TrivialRowGrouping>::setUpViOperator() const {
+    if (!viOperator) {
+        viOperator = std::make_shared<solver::helper::ValueIterationOperator<ValueType, TrivialRowGrouping>>();
+        viOperator->setMatrixBackwards(this->discountedA);
+    }
+    /*if (this->choiceFixedForRowGroup) {
+        // Ignore those rows that are not selected
+        assert(this->initialScheduler);
+        auto callback = [&](uint64_t groupIndex, uint64_t localRowIndex) {
+            return this->choiceFixedForRowGroup->get(groupIndex) && this->initialScheduler->at(groupIndex) != localRowIndex;
+        };
+        viOperator->setIgnoredRows(true, callback);
+    }*/
+}
+
+template<typename ValueType, bool TrivialRowGrouping>
+void DiscountingHelper<ValueType, TrivialRowGrouping>::showProgressIterative(uint64_t iteration) const {
+    progressMeasurement->updateProgress(iteration);
+}
+
+template<typename ValueType, bool TrivialRowGrouping>
+bool DiscountingHelper<ValueType, TrivialRowGrouping>::solveWithDiscountedValueIteration(storm::Environment const& env,
+                                                                                         std::optional<OptimizationDirection> dir, std::vector<ValueType>& x,
+                                                                                         std::vector<ValueType> const& b) const {
+    storm::solver::helper::DiscountedValueIterationHelper<ValueType, TrivialRowGrouping> viHelper(viOperator);
+    uint64_t numIterations{0};
+    auto viCallback = [&](solver::SolverStatus const& current) {
+        return current;
+        showProgressIterative(numIterations);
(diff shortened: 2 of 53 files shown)


sjunges:
This is absolutely great. 

I have two questions:
- Does this code support convergence detection based on the Bellman residual error?
- Do we throw relevant error messages when doing, e.g., weak bismulation?

Best,
Sebastian

AlexBork:
Yes, convergence is detected using the Bellman residual, see ``src/storm/solver/helper/DiscountedValueIterationHelper.cpp``. I have also tested convergence using only the number of iterations as the criterion, i.e. letting the iteration run until it is guaranteed that we are precise enough. That was however highly inefficient as the number of iterations for guaranteed convergence is usually far larger than the actual number needed to reach the precision.

The error messages are not implemented, I will take care of it! I'm happy about any pointers where such problems might occur.

src/storm/logic/DiscountedCumulativeRewardFormula.h
```diff
@@ -0,0 +1,78 @@
+#pragma once
+
+#include "storm/logic/PathFormula.h"
+
+#include "storm/logic/TimeBound.h"
+#include "storm/logic/TimeBoundType.h"
+
+namespace storm {
+namespace logic {
+class DiscountedCumulativeRewardFormula : public PathFormula {
```
sjunges, line 10: Would the amount of code duplication shrink if you instead use another base class?

src/storm/modelchecker/helper/DiscountingHelper.cpp
```diff
@@ -0,0 +1,146 @@
...
+    discountedA = discountedMatrix;
+}
+
+template<typename ValueType, bool TrivialRowGrouping>
+void DiscountingHelper<ValueType, TrivialRowGrouping>::setUpViOperator() const {
+    if (!viOperator) {
+        viOperator = std::make_shared<solver::helper::ValueIterationOperator<ValueType, TrivialRowGrouping>>();
+        viOperator->setMatrixBackwards(this->discountedA);
+    }
+    /*if (this->choiceFixedForRowGroup) {
```
sjunges, line 37: ....? :-)

src/storm/solver/helper/DiscountedValueIterationHelper.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+namespace storm::solver::helper {
+
+template<typename ValueType, storm::OptimizationDirection Dir, bool Relative>
+class DiscountedVIOperatorBackend {
+   public:
+    DiscountedVIOperatorBackend(ValueType const& precision, ValueType const& discountFactor, ValueType const& maximalAbsoluteReward)
+        : precision{precision},
+          discountFactor{discountFactor},
+          bound{(((storm::utility::one<ValueType>() - discountFactor) * precision) / (2 * discountFactor))} {
```
sjunges, line 15: I am slightly confused by the factor two here? Can we add a comment that clarifies this bound variable? :-)

AlexBork: I checked and the factor 2 is indeed incorrect, thanks for spotting that! I added a reference to the Russell/Norvig book

src/storm/modelchecker/helper/DiscountingHelper.cpp
AlexBork: Ah, copy artifacts :D removed it now

src/storm/logic/DiscountedCumulativeRewardFormula.h
AlexBork: Yes, that makes things a lot more compact. Thanks for the hint

AlexBork:
Thanks @sjunges for the review! I incorporated your comments now.

sjunges:
Thanks! @tquatmann do you want to have a look before merging?

src/storm/modelchecker/helper/DiscountingHelper.h
```diff
@@ -0,0 +1,67 @@
...
+class Environment;
+namespace modelchecker {
+namespace helper {
+template<typename ValueType, bool TrivialRowGrouping = false>
+class DiscountingHelper : public SingleValueModelCheckerHelper<ValueType, storm::models::ModelRepresentation::Sparse> {
+   public:
+    DiscountingHelper(storm::storage::SparseMatrix<ValueType> const& A, ValueType discountFactor);
+    DiscountingHelper(storm::storage::SparseMatrix<ValueType> const& A, ValueType discountFactor, bool trackScheduler);
+
+    void setUpViOperator() const;
```
tquatmann, line 18: Wouldn't it be simpler to make`setUpViOperator` private and call it within `solveWithDiscountedValueIteration`?

src/storm/modelchecker/prctl/helper/SparseDtmcPrctlHelper.cpp
```diff
@@ -379,6 +380,61 @@ std::vector<ValueType> SparseDtmcPrctlHelper<ValueType, RewardModelType>::comput
...
+    // Only compute the result if the model has at least one reward this->getModel().
+    STORM_LOG_THROW(!rewardModel.empty(), storm::exceptions::InvalidPropertyException, "Missing reward model for formula. Skipping formula.");
+
+    // Compute the reward vector to add in each step based on the available reward models.
+    std::vector<ValueType> totalRewardVector = rewardModel.getTotalRewardVector(transitionMatrix);
+
+    // Initialize result to the zero vector.
+    std::vector<ValueType> result(transitionMatrix.getRowGroupCount(), storm::utility::zero<ValueType>());
+
+    auto multiplier = storm::solver::MultiplierFactory<ValueType>().create(env, transitionMatrix);
```
tquatmann, line 405: This does not use the `discountFactor`, right?

tquatmann (commented):
Thanks for the great work! 

- Do we (have to) enforce somewhere that the discount factor is strictly between 0 and 1? For both formula types (total and cumulative formulas)? I might have missed this in my review
- I'd suggest to add more tests, in particular in `FormulaParserTest`, `DtmcPrctlModelCheckerTest`, and `SchedulerGenerationMdpPrctlModelCheckerTest`. The existing test also doesn't consider `DiscountedCumulativeRewardFormula`s

AlexBork:
I have switched the order of discount factor and bound in the cumulative discounted reward formula syntax. This prevents problems if a time bound is given without a keyword as the expression for the discount factor was not correctly parsed.
For example previously "R=? [Cdiscount=(0.9) <5]" could not be parsed as "(0.9) <5" was interpreted as one expression for the discount factor.
The new syntax is now "R=? [C<5discount=(0.9)]".

AlexBork:
Yes, I will make a push to have this fully ready for 1.11.

tquatmann:
Can confirm the problem. I created issue #753

src/storm/modelchecker/prctl/helper/SparseDtmcPrctlHelper.cpp
```diff
@@ -20,6 +21,7 @@
 #include "storm/modelchecker/prctl/helper/rewardbounded/MultiDimensionalRewardUnfolding.h"
 #include "storm/modelchecker/results/ExplicitQuantitativeCheckResult.h"
 
+#include "storm/environment/solver/MinMaxSolverEnvironment.h"
```
tquatmann: Is this include necessary?

src/storm/modelchecker/prctl/helper/SparseDtmcPrctlHelper.cpp
```diff
@@ -1,5 +1,6 @@
 #include "storm/modelchecker/prctl/helper/SparseDtmcPrctlHelper.h"
 
+#include "../../../environment/solver/NativeSolverEnvironment.h"
```
tquatmann: For consistency, please use an absolute path (and move to the other environment include(s))

src/storm/modelchecker/prctl/helper/SparseDtmcPrctlHelper.cpp
```diff
@@ -424,13 +426,42 @@ std::vector<ValueType> SparseDtmcPrctlHelper<ValueType, RewardModelType>::comput
...
+            STORM_LOG_THROW(false, storm::exceptions::NotSupportedException,
+                            "Solving discounted total reward objectives is currently only supported by the native solver using value-iteration-based methods.");
+        } else {
+            STORM_LOG_WARN("The default method does not support solving discounted total reward Falling back to value iteration.");
+        }
+    }
+
+    STORM_LOG_WARN_COND(env.solver().native().getMethod() != storm::solver::NativeLinearEquationSolverMethod::IntervalIteration,
+                        "Interval iteration is not supported for discounted total reward objectives. Falling back to value iteration. Note that value "
+                        "iteration guarantees soundness for discounted objectives.");
```
tquatmann: I don't think we should try to look at the settings for linear equation solvers here, i.e., I'd suggest

```cpp
    if (env.solver().isForceExact()) {
        STORM_LOG_WARN("Exact solving of discounted total reward objectives is currently not supported. The result will be an approximative");
    }
```

If we at some point implement other solution methods for discounted properties, there should be a dedicated setting for that algorithm (similar to long-run average right now)

src/storm/modelchecker/prctl/helper/SparseMdpPrctlHelper.cpp
```diff
@@ -1009,12 +1010,34 @@ MDPSparseModelCheckingHelperReturnType<SolutionType> SparseMdpPrctlHelper<ValueT
...
+    if (!(env.solver().minMax().getMethod() == storm::solver::MinMaxMethod::ValueIteration ||
+          env.solver().minMax().getMethod() == storm::solver::MinMaxMethod::SoundValueIteration ||
+          env.solver().minMax().getMethod() == storm::solver::MinMaxMethod::OptimisticValueIteration ||
+          env.solver().minMax().getMethod() == storm::solver::MinMaxMethod::IntervalIteration)) {
+        STORM_LOG_THROW(false, storm::exceptions::NotSupportedException,
+                        "Solving discounted total reward objectives is currently only supported using value-iteration-based methods.");
+    }
+    STORM_LOG_WARN_COND(env.solver().minMax().getMethod() != storm::solver::MinMaxMethod::IntervalIteration,
+                        "Interval iteration is not supported for discounted total reward objectives. Falling back to value iteration. Note that value "
+                        "iteration guarantees soundness for discounted objectives.");
```
tquatmann: See above.

src/storm/modelchecker/prctl/helper/SparseDtmcPrctlHelper.cpp
tquatmann: If we keep things like that, keep in mind that we nowadays also have guessing value iteration #724 .

src/storm/solver/multiplier/Multiplier.cpp
```diff
@@ -81,8 +81,24 @@ void Multiplier<ValueType>::repeatedMultiplyAndReduceWithFactor(Environment cons
     progress.setMaxCount(n);
     progress.startNewMeasurement(0);
     for (uint64_t i = 0; i < n; ++i) {
+        std::transform(x.begin(), x.end(), x.begin(), [factor](ValueType& c) { return c * factor; });
         multiplyAndReduce(env, dir, x, b, x);
+        if (storm::utility::resources::isTerminate()) {
```
tquatmann, line 88: `progress.updateProgress(i);` seems to be missing (apparently also at other places in this file).

src/test/storm/modelchecker/prctl/mdp/DiscountingMdpPrctlModelCheckerTest.cpp
```diff
@@ -0,0 +1,389 @@
+#include "test/storm_gtest.h"
```
tquatmann: Can this be merged with the MdpPrctlModelCheckerTest to reduce code duplications? (Same for the DTMC variant)?
I know that some environments there don't apply but these could be skipped if they don't work

tquatmann:
The problem with the tests here should hopefully be resolved, once #754 is merged.

src/storm/modelchecker/prctl/helper/SparseDtmcPrctlHelper.cpp
AlexBork: I agree that this is clunky, so I'll remove the case distinctions. Otherwise it will be a pain to maintain.
I'm then not sure if a warning or an error should be thrown as "exact" should mean that I am using an exact solution method (if I'm not very explicitly forcing Storm to use something different).
Opinions?

AlexBork: Sidenote: an option to use rational numbers, but approximative algorithms would be nice at some point

src/test/storm/modelchecker/prctl/mdp/DiscountingMdpPrctlModelCheckerTest.cpp
AlexBork: Right now that would make sense. Though if we want to extend Storm to support different solution methods, I think it would be better to keep then separate.
If you think it's better to separate them later, I can merge the tests for now.

src/storm/modelchecker/prctl/helper/SparseDtmcPrctlHelper.cpp
tquatmann: From an API point of view, you already can have `ValueType=storm::RationalNumber` and `env.solver().isForceExact() == false`. But you cannot trigger this behaviour from command line.

Regarding `WARN` vs. `THROW`: There are good arguments for both. I believe in other cases (time bounded reachability in CTMC comes to mind) we just THROW.

src/test/storm/modelchecker/prctl/mdp/DiscountingMdpPrctlModelCheckerTest.cpp
tquatmann: I think we should find a way to get rid of those massive code duplications in the tests. The `Lra*Test.cpp` classes are guilty of this, too. But that's something for a different PR :)

src/test/storm/modelchecker/prctl/mdp/MdpPrctlModelCheckerTest.cpp
```diff
@@ -245,6 +245,7 @@ class SparseDoubleLPEnvironment {
         storm::Environment env;
         env.solver().minMax().setMethod(storm::solver::MinMaxMethod::LinearProgramming);
         env.solver().minMax().setPrecision(storm::utility::convertNumber<storm::RationalNumber>(1e-6));
+        env.solver().minMax().setRelativeTerminationCriterion(false);
```
tquatmann: This seems unrelated to the given PR.
```suggestion
```

src/test/storm/modelchecker/prctl/mdp/MdpPrctlModelCheckerTest.cpp
```diff
@@ -260,6 +261,7 @@ class SparseDoubleViToLPEnvironment {
         storm::Environment env;
         env.solver().minMax().setMethod(storm::solver::MinMaxMethod::ViToLp);
         env.solver().minMax().setPrecision(storm::utility::convertNumber<storm::RationalNumber>(1e-6));
+        env.solver().minMax().setRelativeTerminationCriterion(false);
```
tquatmann: This seems unrelated to the given PR.
```suggestion
```

tquatmann (approved):

src/test/storm/modelchecker/prctl/mdp/MdpPrctlModelCheckerTest.cpp
AlexBork: It is necessary for result to be computed correctly with the configuration. Alternatively I can configure the test to be skipped for configs without the absolute termination criterion set

tquatmann:
LGTM! Can be merged now :)

PR #622, opened 2024-09-10
Made BitVector conform to the std::ranges::range concept

weiss-i-net:
The BitVector iterators weren't default-construtable and were missing postfix increment operators and thus not fulfilling the range concept (https://en.cppreference.com/w/cpp/ranges/range). Changing this makes it possible to use BitVector with C++20s range adaptors. E.g. for code like this:

```c++
for (auto i : some_bitvector | std::views::filter(some_lambda)) {
  do_something(i);
}
```

### DIFF
--- a/src/storm/storage/BitVector.cpp
+++ b/src/storm/storage/BitVector.cpp
@@ -103,7 +103,7 @@ BitVector::const_reverse_iterator& BitVector::const_reverse_iterator::operator++
     currentIndex = getNextIndexWithValue<true, true>(dataPtr, lowerBound, --currentIndex);
     return *this;
 }
-BitVector::const_reverse_iterator& BitVector::const_reverse_iterator::operator++(int) {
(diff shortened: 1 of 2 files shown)


volkm:
Thanks for the PR.

I personally find it a bit strange that a default constructor is needed (which is not properly initialized), but this seems to be [okay](https://stackoverflow.com/questions/78121564/are-there-requirements-for-the-default-constructed-stdforward-iterator-state-f).
Maybe add some static assert of the form `static_assert(std::ranges::forward_range<BitVector>);` to ensure that the rightf range concept is supported?

weiss-i-net:
I had the same thought and read the same stack overflow thread :D

The static_assert is a good idea. What would be a good place to put it?

tquatmann:
>  What would be a good place to put it?

I don't think we have a 'proper' place for these kind of things.  
I'd suggest in `BitVector.h` right after the class declaration?

PR #623, opened 2024-09-20
Part of a fix for a double free issue in StormPy

TheGreatfpmK:
Read more information here: https://github.com/moves-rwth/stormpy/pull/185

In short, pybind11 can cause problems while exposing smart pointer objects. With this change we don't need to access the BeliefMdpExplorer object via the function getInteractiveBeliefExplorer anymore, as the only functionality we needed in that object (that is the setFMSchedValueList function) is being exposed into the BeliefExplorationPomdpModelChecker class.

### DIFF
--- a/src/storm-pomdp/modelchecker/BeliefExplorationPomdpModelChecker.cpp
+++ b/src/storm-pomdp/modelchecker/BeliefExplorationPomdpModelChecker.cpp
@@ -1458,6 +1458,11 @@ BeliefExplorationPomdpModelChecker<PomdpModelType, BeliefValueType, BeliefMDPTyp
 }
 
 template<typename PomdpModelType, typename BeliefValueType, typename BeliefMDPType>
+void BeliefExplorationPomdpModelChecker<PomdpModelType, BeliefValueType, BeliefMDPType>::setFMSchedValueList(std::vector<std::vector<std::unordered_map<uint64_t, ValueType>>> valueList) {
+    this->interactiveUnderApproximationExplorer->setFMSchedValueList(valueList);
+}
+
+template<typename PomdpModelType, typename BeliefValueType, typename BeliefMDPType>
 BeliefValueType BeliefExplorationPomdpModelChecker<PomdpModelType, BeliefValueType, BeliefMDPType>::rateObservation(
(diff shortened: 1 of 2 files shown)


volkm:
Looks good to me, but @AlexBork has probably the better insights.

Could you adapt the code formatting?

TheGreatfpmK:
I fixed the formatting. I would wait for approval from Alex, as maybe there's a better way to do this. And also while this seems to have fixed the initial issues, there still is another memory-related issue we found right now so I will take a look at it to check whether we need any more changes in Storm before we merge this.

TheGreatfpmK:
Ok this should be everything we need to update in Storm to fix the issues, so when Alex approves it, you can merge.

AlexBork:
I'm back in the office and will take care of this.

src/storm-pomdp/modelchecker/BeliefExplorationPomdpModelChecker.cpp
```diff
@@ -1457,6 +1457,12 @@ BeliefExplorationPomdpModelChecker<PomdpModelType, BeliefValueType, BeliefMDPTyp
     return interactiveUnderApproximationExplorer;
 }
 
+template<typename PomdpModelType, typename BeliefValueType, typename BeliefMDPType>
+void BeliefExplorationPomdpModelChecker<PomdpModelType, BeliefValueType, BeliefMDPType>::setFMSchedValueList(
+    std::vector<std::vector<std::unordered_map<uint64_t, ValueType>>> valueList) {
+    this->interactiveUnderApproximationExplorer->setFMSchedValueList(valueList);
```
AlexBork, line 1463: I think the ``this->`` is not necessary here. For consistency with the rest of the code, I'd remove it.

AlexBork (commented):
One small point, otherwise this looks good. Thank you! Can you check if the suggested change doesn't break anything in your change?

src/storm-pomdp/modelchecker/BeliefExplorationPomdpModelChecker.cpp
TheGreatfpmK: Yes agreed I removed it.

TheGreatfpmK:
Thanks Alex! I resolved your suggestion. If there's nothing else we can merge

PR #627, opened 2024-11-01
iMC changes outside of storm-pars

linusheck:
Draft PR containing all changes I made for iMC algorithms that are outside of `storm-pars`. Tests in the storm-pars part (no PR for this yet) cover most lines in this PR.

In more detail, this PR contains:
- a new function for `FlexibleSparseMatrix` that adds new rows
- a `SchedulerTrackingHelper` for iMCs and a interval (=robust) scheduler index in the minmax solvers for warm-starting
- interval support for the `(Iterative)MinMaxLinearEquationSolver`
- an `addNotCurrentModel` function for SMT solvers that adds the constraint that the current model is not a solution
- AGNUM support for Z3
- interval MC end component elimination (`RobustMaximalEndComponentDecomposition`, `RobustMaximalEndComponent`)
- optimisations in `ValueIterationOperator` for iMCs
- a couple more functions for `storm::Interval`


src/storm/solver/IterativeMinMaxLinearEquationSolver.h
```diff
@@ -1,6 +1,8 @@
 #pragma once
 
+#include <variant>
```
sjunges, line 3: Do we need this include here? It is a standard header, so we can maybe add it to the PCH, but still?

src/storm/solver/IterativeMinMaxLinearEquationSolver.h
```diff
@@ -39,9 +40,9 @@ class IterativeMinMaxLinearEquationSolver : public StandardMinMaxLinearEquationS
    private:
     MinMaxMethod getMethod(Environment const& env, bool isExactMode) const;
 
-    bool solveInducedEquationSystem(Environment const& env, std::unique_ptr<LinearEquationSolver<ValueType>>& linearEquationSolver,
+    bool solveInducedEquationSystem(Environment const& env, std::unique_ptr<LinearEquationSolver<SolutionType>>& linearEquationSolver,
                                     std::vector<uint64_t> const& scheduler, std::vector<SolutionType>& x, std::vector<ValueType>& subB,
-                                    std::vector<ValueType> const& originalB) const;
+                                    std::vector<ValueType> const& originalB, OptimizationDirection dir) const;
```
sjunges, line 45: do we write this as const& ? it does not really matter for the compiler

src/storm/solver/TopologicalMinMaxLinearEquationSolver.cpp
```diff
@@ -151,6 +151,7 @@ bool TopologicalMinMaxLinearEquationSolver<ValueType, SolutionType>::internalSol
         }
 
         // If requested, we store the scheduler for retrieval.
+        // TODO when is this code called??????? im a bit confused
```
sjunges, line 154: @tquatmann

src/storm/solver/helper/SchedulerTrackingHelper.h
```diff
@@ -1,22 +1,74 @@
 #pragma once
 #include <memory>
+#include <optional>
```
sjunges, line 3: we use boost optional and not std::optional here...?

src/storm/solver/helper/ValueIterationOperator.cpp
```diff
@@ -3,6 +3,8 @@
 #include <optional>
 
 #include "storm/adapters/RationalNumberAdapter.h"
+#include "storm/adapters/RationalNumberForward.h"
```
sjunges, line 6: If we already include the adapter, then there should be no need to inlcude the forward

src/storm/solver/helper/ValueIterationOperator.cpp
```diff
@@ -44,13 +46,33 @@ void ValueIterationOperator<ValueType, TrivialRowGrouping, SolutionType>::setMat
...
     } else {
-        matrixColumns.push_back(StartOfRowIndicator);  // Indicate start of first row
-        for (auto rowIndex : indexRange<Backward>(0, numRows)) {
-            for (auto const& entry : matrix.getRow(rowIndex)) {
-                matrixValues.push_back(entry.getValue());
-                matrixColumns.push_back(entry.getColumn());
+        if constexpr (std::is_same<ValueType, storm::Interval>::value) {
+            applyCache.hasOnlyConstants.clear();
+            applyCache.hasOnlyConstants.grow(matrix.getRowCount());
+            // TODO Implement hasTwoSuccessors
```
sjunges, line 52: TODO?

src/storm/solver/helper/ValueIterationOperator.h
```diff
@@ -291,11 +310,28 @@ class ValueIterationOperator {
...
+                    STORM_LOG_THROW(false, storm::exceptions::NotImplementedException, "Value Iteration is not implemented with pairs and interval-models.");
+                    // Notice the unclear semantics here in terms of how to order things.
+                } else {
+                    result += operand[*matrixColumnIt] * lower;
+                }
+            }
+            return result;
+        }
+
+        // TODO I think this is a problem if we have probabilities and a state that is going to the vector, we don't count that
```
sjunges, line 329: TODO?

src/storm/solver/helper/ValueIterationOperator.h
```diff
@@ -291,11 +310,28 @@ class ValueIterationOperator {
...
+        // Currently "fixed in preprocessing"
+        // It's different for rewards (same problem somewhere else, search for word "octopus" in codebase)
```
sjunges, line 331: ? :-)

src/storm/solver/helper/ValueIterationOperator.h
linusheck: I think we should just check these conditions before and make them an error because there's no way to fix it in valueiterationoperator.

src/storm/solver/helper/ValueIterationOperator.cpp
linusheck: I'll just remove the `hasTwoSuccessors` thing for now :)

src/storm/solver/IterativeMinMaxLinearEquationSolver.h
linusheck: like this, it's consistent with the rest of the methods

src/storm/solver/TopologicalMinMaxLinearEquationSolver.cpp
tquatmann: When we want to produce an optimal scheduler for the given MinMax problem, the `trackScheduler` flag will be set to true. 
The code below is doing a single iteration of value iteration to find out the optimal choices.

src/storm/solver/IterativeMinMaxLinearEquationSolver.cpp
```diff
@@ -619,28 +652,56 @@ bool IterativeMinMaxLinearEquationSolver<ValueType, SolutionType>::solveEquation
...
-    }
+    // This code duplication is necessary because the helper class is different for the two cases.
+    if (this->A->hasTrivialRowGrouping()) {
+        storm::solver::helper::ValueIterationHelper<ValueType, true, SolutionType> viHelper(viOperatorTriv);
 
-    if (!this->isCachingEnabled()) {
-        clearCache();
-    }
+        uint64_t numIterations{0};
+        auto viCallback = [&](SolverStatus const& current) {
```
sjunges, line 660: the viCallback definition can be moved out of the if, right?

sjunges:
Does this depend on #588? That PR is still not ready for merging, is it?

glatteis:
> Does this depend on #588? That PR is still not ready for merging, is it?

It doesn’t depend on that PR I‘m pretty sure, because that PR mainly changes thing in storm-pars

tquatmann (commented):
Thanks for the great work!

I think some test cases for the robust mec decomposition and the scheduler tracking for robust value iteration would help with maintaining the new features.

src/storm/adapters/Z3ExpressionAdapter.cpp
```diff
@@ -166,6 +166,9 @@ storm::expressions::Expression Z3ExpressionAdapter::translateExpression(z3::expr
                         return manager.rational(storm::utility::convertNumber<storm::RationalNumber>(std::string(Z3_get_numeral_string(expr.ctx(), expr))));
                     }
                 }
+            case Z3_OP_AGNUM:
+                return manager.rational(storm::utility::convertNumber<storm::RationalNumber>(
+                    std::string(Z3_get_numeral_string(expr.ctx(), Z3_get_algebraic_number_lower(expr.ctx(), expr, 16)))));
```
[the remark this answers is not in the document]
linusheck: I agree that the hardcoded precision is bad, but I'm not sure how to make this configurable in a way that makes sense.

src/storm/solver/MinMaxLinearEquationSolver.h
```diff
@@ -200,6 +205,9 @@ class MinMaxLinearEquationSolver : public AbstractEquationSolver<SolutionType> {
     /// The scheduler choices that induce the optimal values (if they could be successfully generated).
     mutable boost::optional<std::vector<uint_fast64_t>> schedulerChoices;
 
+    /// For interval models, this index points into the schedulerChoices and is a state -> index map.
+    mutable boost::optional<std::vector<uint_fast64_t>> robustSchedulerIndex;
```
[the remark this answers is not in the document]
linusheck: In an MDP, the scheduler choices are just one number for each row group - the row that was chosen. In an iMC (this is currently the only interval model that supports extracting schedulers here), a scheduler is an order of entries within a single row. In the current representation, the `schedulerChoices` from `robustSchedulerIndex[i]` to `robustSchedulerIndex[i+1]` (exclusive) are this permutation. So, the `robustSchedulerIndex[i]` currently just counts how many matrix entries come before row `i`.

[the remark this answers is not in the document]
linusheck: I'll put this into the comment

src/storm/storage/RobustMaximalEndComponentDecomposition.h
```diff
@@ -0,0 +1,112 @@
...
+#include "storm/storage/Decomposition.h"
+#include "storm/storage/RobustMaximalEndComponent.h"
+#include "storm/storage/StronglyConnectedComponent.h"
+#include "storm/storage/sparse/StateType.h"
+#include "storm/utility/OptionalRef.h"
+
+namespace storm::storage {
+
+/*!
+ * This class represents the decomposition of a nondeterministic model into its maximal end components.
```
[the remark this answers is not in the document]
linusheck: I deleted the RobustMaximalEndComponent :)

The RobustMaximalEndComponentDecomposition has almost entirely different code to the MaximalEndComponentDecomposition right now, but I totally see that we should merge them to support models like iMDPs in the future. It might be a bit fiddly to implement Haddad-Monmege's algorithm for iMDPs while not regressing performance on standard MDPs.

For now, it should be possible to add a `bool robust` and switch between the different algorithms. I'll attempt that next week :)

src/storm/adapters/Z3ExpressionAdapter.cpp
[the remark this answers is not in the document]
tquatmann: Imo, just a small comment would already help to make it obvious that some rounding happens here.

Also, the `Z3ExpressionAdapter` could have a flag `bool AllowRoundingAlgebraicNumbers` that one has to set to true in contexts where algebraic numbers should be expected and where rounding would be fine. Otherwise, I can see that this could become a somewhat unexpected source of approximation error.

src/storm/storage/RobustMaximalEndComponentDecomposition.h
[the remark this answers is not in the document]
linusheck: > the `robust` flag can be a compile-time constant (set to `std::is_same_v<ValueType, storm::Interval>` or as extra template parameter). Then, this should be possible without runtime penalities for standard MDPs

It's a bit more difficult:
- the robust maximal end component decomposition needs the vector as well, not just the matrix, so the method signatures are different
- the current end component decomposition is based on `MaximalEndComponent`, but that class has a bunch of methods that are nonsensical for robust end components, like the choices per state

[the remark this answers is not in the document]
linusheck: The `vector` is just the vector belonging to the transition matrix when writing it as a system of equations, as is usually done in Storm. E.g., the state rewards for expected rewards, or the probability to reach one in reachability probabilities. I'll add that as a comment

src/storm/adapters/Z3ExpressionAdapter.cpp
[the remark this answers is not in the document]
linusheck: > The Z3ExpressionAdapter could have a flag bool AllowRoundingAlgebraicNumbers that one has to set to true in contexts where algebraic numbers should be expected and where rounding would be fine. Otherwise, I can see that this could become a somewhat unexpected source of approximation error.

I agree but I'm not sure on what level I should expose this. The user just interacts with an SMT solver they get from an SMT solver factory - should they be able to say something about rounding on that level? Should Z3SMTSolver have a function "allowRounding"?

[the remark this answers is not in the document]
sjunges: I think it would be indeed nice to have this on the level of SMT solvers. In particular, I could phantom that one in the future would like to add something like dReal where we would run into the same issues.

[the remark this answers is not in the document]
linusheck: I've checked it out and I don't believe that I have time to implement this right now, but I will do it later. I will open an issue.

src/storm/storage/RobustMaximalEndComponentDecomposition.h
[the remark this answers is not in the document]
linusheck: > I propose to merge the PR with a RobustMaximalEndComponentDecomposition and we can try to reduce the code duplications in a separate PR. I believe C++20 concepts and requires expressions could handle this quite elegantly.

Let's do that :)

linusheck:
If we agree to postpone these last two things (we've agreed on the end component decomposition refactoring but not yet on the SMT solver interfaces), we can merge this PR. Currently, the only place where SMT results are rounded are after big-step preprocessing. The changes to fix the SMT solver APIs are a lot of source code changes, but very doable later. See issue https://github.com/moves-rwth/storm/issues/713.

linusheck:
seems that spot is failing :(

volkm:
One issue with Spot will be fixed by #703, the other ones are tracked in #707. Other than that your CI seems to be successful.
We should merge it after version 1.10 is released.

PR #629, opened 2024-11-11
Observationtraceunfolder update

sjunges:
- Update trace unfolder such that it can also support computing reachability probabilities and that conditional probs must be computed via a dedicated routine for conditional probs. The old behavior is available via the "rejectionSamplingMode"
- Also fix a problem in the belieftracker reset function.


src/storm-pomdp/transformer/ObservationTraceUnfolder.cpp
```diff
@@ -36,32 +39,52 @@ std::shared_ptr<storm::models::sparse::Mdp<ValueType>> ObservationTraceUnfolder<
...
+        // The violated state is only used if we do no use the rejection semantics.
+        ++newStateIndex;
+    }
+    // Add this initial state:
+    uint64_t initialState = newStateIndex;
+    ++newStateIndex;
+
+    unfoldedToOldNextStep[initialState] = actualInitialStates.getNextSetIndex(0);
 
+    uint64_t resetDestination = options.rejectionSampling ? initialState : violatedState;  // Should be initial state for the standard semantics.
```
tquatmann, line 72:
```suggestion
    uint64_t const violatedState = newStateIndex;
    if (!options.rejectionSampling) {
        // The violated state is only used if we do no use the rejection semantics.
        ++newStateIndex;
    }
    // Add this initial state:
    uint64_t const initialState = newStateIndex;
    ++newStateIndex;

    unfoldedToOldNextStep[initialState] = actualInitialStates.getNextSetIndex(0);

    uint64_t const resetDestination = options.rejectionSampling ? initialState : violatedState;  // Should be initial state for the standard semantics.
```

tquatmann (approved):
LGTM! Just a very minor suggestion to improve readability a bit.

PR #630, opened 2024-11-12
No RationalNumber in ExplicitQualitativeCheckResult.h

linusheck:
Removes the need to know about `storm::RationalNumber` in `ExplicitQualitativeCheckResult.h` by removing the default template parameter. My PAYNT build broke because this was the case.


linusheck:
Sadly, this doesn't save us from having to do a 
```
```
because state valuations also mention rational numbers.

tquatmann:
LGTM

PR #632, opened 2024-11-13
Two Monotonicity Bugfixes

SvStein:
1.  Bug: Not all states were added to the order properly when using PLA bounds. They were simply set to done when all their successors could be ordered via PLA bounds, but not put into the order. They were at best only added trivially (between top and bottom) when required later as another state's successor. These states are now no longer set to done so that the main loop will still consider them and integrate them into the order properly.
2. Bug: The zeroconf benchmark only resulted in no increasing parameter (with PLA) or one increasing parameter (without PLA), whereas the correct (and previously attained) result is that both parameters are increasing. The order was correct and did lend itself to concluding the correct result but local monotonicity was not checked at all states, specifically not those from the statesToHandle vector (whose purpose is currently unclear). This restriction was now lifted so that local monotonicity is checked on these too if the order has been expanded with them.


linusheck:
LGTM. Am I allowed to just merge it @tquatmann ? :)

PR #640, opened 2024-11-19
iMC changes inside of storm-pars

linusheck:
Merge https://github.com/moves-rwth/storm/pull/627 first


src/storm/solver/IterativeMinMaxLinearEquationSolver.cpp
```diff
@@ -630,10 +630,12 @@ bool IterativeMinMaxLinearEquationSolver<ValueType, SolutionType>::solveEquation
         }
         storm::Environment const& environmentOfSolver = environmentOfSolverStorage ? *environmentOfSolverStorage : env;
 
-        solveInducedEquationSystem(environmentOfSolver, linEqSolver, this->getInitialScheduler(), x, *auxiliaryRowGroupVector, b, dir);
-        // If we were given an initial scheduler and are maximizing (minimizing), our current solution becomes
-        // always less-or-equal (greater-or-equal) than the actual solution.
-        guarantee = maximize(dir) ? SolverGuarantee::LessOrEqual : SolverGuarantee::GreaterOrEqual;
+        bool success = solveInducedEquationSystem(environmentOfSolver, linEqSolver, this->getInitialScheduler(), x, *auxiliaryRowGroupVector, b, dir);
+        if (success) {
```
sjunges, line 694: What if this is not successful?

src/storm/solver/IterativeMinMaxLinearEquationSolver.cpp
```diff
@@ -183,7 +183,7 @@ void IterativeMinMaxLinearEquationSolver<ValueType, SolutionType>::extractSchedu
     }
 
     // Set the correct choices.
-    STORM_LOG_WARN_COND(!viOperatorTriv && !viOperatorNontriv,
+    STORM_LOG_WARN_COND(viOperatorTriv || viOperatorNontriv,
```
sjunges: This warning is confusing, in particular because it reads like a tautology. 

Also, was this really always the wrong way around?

src/storm/solver/helper/SchedulerTrackingHelper.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 #include <memory>
+#include <optional>
```
sjunges: What has changed in this file that we need to include a header here?

src/storm-pars/api/region.h
```diff
@@ -300,49 +410,30 @@ std::unique_ptr<storm::modelchecker::RegionRefinementCheckResult<ValueType>> che
...
-    auto regionChecker =
-        initializeRegionModelChecker(env, model, task, engine, generateSplitEstimates, allowModelSimplification, preconditionsValidated, monotonicitySetting);
-    if (maxSplitsPerStepThreshold && maxSplitsPerStepThreshold < std::numeric_limits<uint64_t>::max()) {
-        regionChecker->setMaxSplitDimensions(maxSplitsPerStepThreshold.value());
-    }
-    auto res = regionChecker->computeExtremalValue(env, region, dir, precision.is_initialized() ? precision.get() : storm::utility::zero<ValueType>(),
-                                                   absolutePrecision, boundInvariant);
-    STORM_LOG_ASSERT(res.first.isConstant(), "result must be a constant");
-    return {storm::utility::convertNumber<storm::RationalNumber>(res.first.constantPart()), std::move(res.second)};
+    // TODO: allow passing these settings? Maybe also pass monotone parameters?
```
sjunges: Review this todo?

src/storm-pars/api/region.h
```diff
@@ -270,6 +319,70 @@ std::unique_ptr<storm::modelchecker::RegionCheckResult<ValueType>> checkRegionsW
     return checkRegionsWithSparseEngine(model, task, regions, engine, hypotheses, sampleVerticesOfRegions);
 }
 
+template<typename ValueType>
```
sjunges: Do we really want this in the API header and not in another file. 

Also, I think we use Settings for things from the command line and environments/options otherwise?

src/storm-pars/derivative/GradientDescentInstantiationSearcher.h
```diff
@@ -16,6 +16,8 @@
 #include "storm/exceptions/WrongFormatException.h"
 #include "storm/models/sparse/Dtmc.h"
 #include "storm/utility/Stopwatch.h"
+#include "storm/utility/logging.h"
```
sjunges: This is typically a sign that we are putting implemetnations in header files... Do we really want that here?

src/storm-pars/modelchecker/instantiation/SparseMdpInstantiationModelChecker.cpp
```diff
@@ -229,6 +229,13 @@ std::unique_ptr<CheckResult> SparseMdpInstantiationModelChecker<SparseModelType,
     return result;
 }
 
+template<typename SparseModelType, typename ConstantType>
+bool SparseMdpInstantiationModelChecker<SparseModelType, ConstantType>::isProbabilistic(
+    storm::utility::parametric::Valuation<typename SparseModelType::ValueType> const& valuation) {
```
sjunges, line 234: Looking at the implementaiton, isn't this checking whether the valuation is well-defined? 

I also would like to notice that this method is implemented in a way that is quite expensive.  Where is it used?

src/storm-pars/analysis/MonotonicityKind.cpp
```diff
@@ -0,0 +1,34 @@
+#include "storm-pars/analysis/MonotonicityKind.h"
```
sjunges: Do we want to move this file and others to modelcehcker/region/monotonicity?

src/storm-pars/settings/modules/ParametricSettings.cpp
```diff
@@ -39,9 +39,12 @@ ParametricSettings::ParametricSettings() : ModuleSettings(moduleName) {
                         .setShortName(transformContinuousShortOptionName)
                         .build());
     this->addOption(storm::settings::OptionBuilder(moduleName, useMonotonicityName, false, "If set, monotonicity will be used.").build());
-    this->addOption(
-        storm::settings::OptionBuilder(moduleName, timeTravellingEnabledName, false, "Enabled time travelling (flip transitions to improve PLA bounds).")
-            .build());
+    this->addOption(storm::settings::OptionBuilder(moduleName, bigStepEnabledName, false, "Enables big step transitions.")
+                        .addArgument(storm::settings::ArgumentBuilder::createBooleanArgument("time-travel", "Enable time-travelling")
```
sjunges: mentions time travelling

src/storm-pars/settings/modules/ParametricSettings.cpp
```diff
@@ -73,8 +76,16 @@ pars::utility::ParametricMode ParametricSettings::getOperationMode() const {
     return *mode;
 }
 
+bool ParametricSettings::isBigStepEnabled() const {
+    return this->getOption(bigStepEnabledName).getHasOptionBeenSet();
+}
+
 bool ParametricSettings::isTimeTravellingEnabled() const {
-    return this->getOption(timeTravellingEnabledName).getHasOptionBeenSet();
+    return this->getOption(bigStepEnabledName).getArgumentByName("time-travel").getValueAsBoolean();
```
sjunges: Duplication?

src/storm-pars/transformer/BigStep.h
```diff
@@ -1,6 +1,14 @@
 #pragma once
 
+#include <carl/core/FactorizedPolynomial.h>
```
sjunges: Includes to carl should go via the adapters to catch warnings etc.

We need to discuss what to do with the cache and the real algebraic numbers. 

Either way, this is polluting our header space right now.

src/storm-pars/transformer/RobustParameterLifter.h
```diff
@@ -0,0 +1,182 @@
+#pragma once
+
+#include <memory>
+#include <set>
+#include <unordered_map>
+#include <vector>
+
+#include "storm-pars/analysis/Order.h"
```
sjunges: With some forward declarations, we can probably reduce compile times significantly here.

src/storm-pars/transformer/IntervalEndComponentPreserver.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include <memory>
+#include <set>
+#include <unordered_map>
+#include <vector>
+
+#include "storm/adapters/RationalFunctionForward.h"
```
sjunges: Forward declarations to avoid completely polluting our headers would be great.

src/storm/solver/IterativeMinMaxLinearEquationSolver.cpp
linusheck: Then your gurantee would be `None`, which it is by default, but I'll explicitely add the branch to mark that clearly

src/storm/solver/IterativeMinMaxLinearEquationSolver.cpp
linusheck: > Also, was this really always the wrong way around?

Seems like in the code in master the warning triggers exactly when the vi operator is not being setup.

It's easiest to read this if warning gets triggered in the `if` branch right below, which I'll do

src/storm-pars/derivative/GradientDescentInstantiationSearcher.h
linusheck: Indeed, this include was uncessary

src/storm-pars/modelchecker/instantiation/SparseMdpInstantiationModelChecker.cpp
linusheck: > Looking at the implementaiton, isn't this checking whether the valuation is well-defined?

Yes, I can rename it

> I also would like to notice that this method is implemented in a way that is quite expensive. Where is it used?

Before checking whether the center of the model is SAT / UNSAT in PLA. So we'll use the model instantiator to instantiate the model anyway, but we currently don't re-use the result. The instantiation checks are quite cheap for PLA so it's not that important there

linusheck: TODO: Write a comment

src/storm-pars/api/region.h
linusheck: Consistent with Storm terminology, this seems to be an `Options` struct like `AbstractionRefinementOptions` or `BeliefExplorationPomdpModelCheckerOptions`.
I'll rename this to `Options` and put it into its own file.

linusheck: I gave the `MonotonicitySetting` in that class the same treatment

src/storm-pars/transformer/BigStep.h
linusheck: I removed every carl include from the .h by moving the implementations to the .cpp and adding a `RawUnivariatePolynomial` to the `RationalFunctionForward`.

src/storm-pars/api/analysis.h
sjunges, line 1: Maybe just remove this file? I don't know, it doesnt feel like part of an API.

src/storm-pars/transformer/BigStep.h
```diff
@@ -0,0 +1,360 @@
+#pragma once
+
+#include <carl/core/FactorizedPolynomial.h>
+#include <carl/core/MultivariatePolynomial.h>
```
sjunges: Something is different than what your comment suggested? :-)

linusheck: i just forgot to push, sorry

linusheck: pushed now :)

src/storm-pars/api/analysis.h
linusheck: Done

src/storm-pars/parser/ParameterRegionParser.cpp
```diff
@@ -78,7 +78,7 @@ storm::storage::ParameterRegion<ParametricType> ParameterRegionParser<Parametric
     Valuation upperBoundaries;
     std::vector<std::string> parameterBoundaries;
     CoefficientType bound = storm::utility::convertNumber<CoefficientType>(regionBound);
-    STORM_LOG_THROW(0 < bound && bound < 1, storm::exceptions::WrongFormatException, "Bound must be between 0 and 1, " << bound << " is not.");
+    // STORM_LOG_THROW(0 < bound && bound < 1, storm::exceptions::WrongFormatException, "Bound must be between 0 and 1, " << bound << " is not.");
```
sjunges: Remove? Do we need to set warnings / errors in other types of verification where regions are used?

src/test/storm-pars/modelchecker/region/monotonicity/MonotonicityHelperTest.cpp
sjunges, line 1: I would suggest to match the file structure with the changes in the main sources?

src/storm-pars/parser/ParameterRegionParser.cpp
linusheck: Not in ones using PL, that is already happening. Indeed, for GD, we should have a warning. I'll remove this one.

src/test/storm-pars/modelchecker/region/monotonicity/MonotonicityHelperTest.cpp
linusheck: done :)

src/storm-pars-cli/storm-pars.cpp
```diff
@@ -82,6 +84,20 @@ std::vector<storm::storage::ParameterRegion<ValueType>> parseRegions(std::shared
     } else if (regionSettings.isRegionBoundSet()) {
         result = storm::api::createRegion<ValueType>(regionSettings.getRegionBoundString(), *model);
     }
+    if (regionSettings.isAssumeGraphPreservingSet()) {
+        for (auto const& region : result) {
```
tquatmann, line 91:
```suggestion
        // We want to warn the user in case the model is actually not graph preserving.
        // However, determining graph-preservingness precisely is hard.
        // As an approximation, we only check if the region intersects 0 or 1.
        for (auto const& region : result) {
```

tquatmann: Add a comment that this is not meant to be a precise check for graph-preservation

src/storm-pars/modelchecker/instantiation/SparseCtmcInstantiationModelChecker.cpp
```diff
@@ -23,6 +23,13 @@ std::unique_ptr<CheckResult> SparseCtmcInstantiationModelChecker<SparseModelType
     return modelChecker.check(env, *this->currentCheckTask);
 }
 
+template<typename SparseModelType, typename ConstantType>
+bool SparseCtmcInstantiationModelChecker<SparseModelType, ConstantType>::isWellDefined(
+    storm::utility::parametric::Valuation<typename SparseModelType::ValueType> const& valuation) {
+    auto const& instantiatedModel = modelInstantiator.instantiate(valuation);
+    return instantiatedModel.getTransitionMatrix().isProbabilistic();
```
tquatmann: Transition matrices of CTMCs store the exit rates, so it's fine that rows sum up to values > 1.
Instead, we should check if all entries are non-negative here.

src/test/storm-pars/transformer/BigStepTest.cpp
```diff
@@ -71,27 +73,33 @@ void testModel(std::string programFile, std::string formulaAsString, std::string
...
-    auto resultPLATT = plaTT.getBoundAtInitState(env, region[0], storm::OptimizationDirection::Minimize);
+    auto plaTT = storm::api::initializeRegionModelChecker<storm::RationalFunction>(env, sharedDtmc, checkTask,
+                                                                                   storm::modelchecker::RegionCheckEngine::RobustParameterLifting);
+    auto resultPLATT = plaTT->getBoundAtInitState(env, region[0], storm::OptimizationDirection::Minimize);
 
     ASSERT_TRUE(resultPLA < resultPLATT) << "Time-Travelling did not make bound better";
 }
 
-class TimeTravelling : public ::testing::Test {
+class BigStep : public ::testing::Test {
```
tquatmann, line 102: also rename the file? :)

src/storm/adapters/RationalFunctionForward.h
```diff
@@ -13,12 +13,16 @@
...
+#include <carl/core/UnivariatePolynomial.h>
 #include <carl/core/VariablePool.h>
 #include "storm/adapters/RationalNumberAdapter.h"
 #endif
 
 namespace carl {
 
+template<typename P>
+class UnivariatePolynomial;
+
```
tquatmann, line 25: There is both, a forward declaration and an include. I assume one of them isn't necessarily.

tquatmann (approved):
I have a few nitpickings, but nothing serious. LGTM!

src/storm-pars/modelchecker/instantiation/SparseCtmcInstantiationModelChecker.cpp
linusheck: Good catch!

src/storm/adapters/RationalFunctionForward.h
linusheck: The include is only if STORM_CARL_SUPPORTS_FWD_DECL is false

sjunges: Which basically only happens for old versions of carl, but that is another discussion.

tquatmann:
This can be merged now, right? :)

linusheck:
I think so

PR #642, opened 2024-12-06
Added storm-pars linker dependency to storm-dft

volkm:
`storm-dft/transformations/DftInstantiator.cpp` depends on `storm-pars/utility/parametric.h`

### DIFF
--- a/src/storm-dft/CMakeLists.txt
+++ b/src/storm-dft/CMakeLists.txt
@@ -15,7 +15,7 @@ set_target_properties(storm-dft PROPERTIES DEFINE_SYMBOL "")
 list(APPEND STORM_TARGETS storm-dft)
 set(STORM_TARGETS ${STORM_TARGETS} PARENT_SCOPE)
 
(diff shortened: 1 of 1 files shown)


sjunges:
Depending on a header should not be a reason to link with a library?

volkm:
That is true and for me, compilation worked before as well.
There was a report on some build issues by a student which lead to this change. We could keep it open for now and wait whether others have the same issue.

sjunges:
This seems a nuclear option that I would like not to use. 

I think we could reconsider whether the necessary lines can go to the carl adapters? That should definitively solve the issue?

volkm:
I will look into moving the functionality to the carl adapter.

AlexBork:
A student of mine has also encountered what appears to be the same problem. Has there been any update on this?

volkm:
After looking at the `DftInstantiator` I figured that it would be better to use existing functions for `substitute`/`evaluate`. This now means that `storm-dft` definitely requires the dependency on the `storm-pars` library. But seeing that future extensions building on parametric DFT will eventually need the parametric library anyway, I think it should be okay.

This should hopefully fix the issues of the students.

sjunges:
LGTM

PR #644, opened 2024-12-11
Improved file handling especially property file

### LINKED ISSUE #643: Confusing error message for --prop option
If one gives to the --prop option a non-existent file, Storm issues a relevant error message:

> Command line arguments: -v --prism model.ma --prop nothing.prop
> ...
> ERROR (FormulaParser.cpp:97): Could not parse formula: nothing.prop
> ERROR (storm-cli.cpp:63): An exception caused Storm to terminate. The message of the exception is: WrongFormatException: Could not parse formula: nothing.prop

But, if the name of the non-existing file starts with a number,  the message changes:

> Command line arguments: -v --prism model.ma --prop 3.prop
> ...
> ERROR (FormulaParserGrammar.cpp:414): Expected expression 3 to be of boolean type.
> ERROR (storm-cli.cpp:63): An exception caused Storm to terminate. The message of the exception is: WrongFormatException: Expected expression 3 to be of boolean type.

This is confusing, since Storm no longer mentions that the file is non-existent, but lets the user think that the
contents of the file are improper, e.g. the 3rd expression in the file is badly typed. If the user has a .prop file
but made a mistake when naming this file on the command line, a lot of time is wasted trying to understand
the issue. 

Even if a formula can follow --prop, "3.prop" should not be taken for a formula because: (1) formula names cannot start
with digits, and (2) it has a ".prop" extension.

PS: By the way, why are Storm error messages always repeated twice? Displaying them once would be sufficient.

volkm:
- Consistently use `openFile` / `closeFile` helper functions
- Introduce namespace `storm::io` for all files in `storm/io`
- Warn if a property string looks like a link to a property file but the file does not exist (Resolves #643). The check warns if the property strings contains any of `.prop`, `.prctl`, `.pctl` or `.csl` (following the [Prism documentation](https://www.prismmodelchecker.org/manual/PropertySpecification/PropertiesFiles))

### DIFF
--- a/src/storm-cli-utilities/model-handling.h
+++ b/src/storm-cli-utilities/model-handling.h
@@ -635,20 +635,20 @@ void exportSparseModel(std::shared_ptr<storm::models::sparse::Model<ValueType>> 
 
     if (ioSettings.isExportBuildSet()) {
         switch (ioSettings.getExportBuildFormat()) {
-            case storm::exporter::ModelExportFormat::Dot:
+            case storm::io::ModelExportFormat::Dot:
                 storm::api::exportSparseModelAsDot(model, ioSettings.getExportBuildFilename(), ioSettings.getExportDotMaxWidth());
                 break;
-            case storm::exporter::ModelExportFormat::Drn:
+            case storm::io::ModelExportFormat::Drn:
                 storm::api::exportSparseModelAsDrn(model, ioSettings.getExportBuildFilename(),
                                                    input.model ? input.model.get().getParameterNames() : std::vector<std::string>(),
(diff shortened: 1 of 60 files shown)


volkm:
Each call to `parseProperties` currently checks whether a potential file with this name exists. File I/O operations can be costly and this can introduce some unnecessary overhead. I would suggest to add dedicated functions for `parsePropertiesFromString` and `parsePropertiesFromFile` to avoid this overhead if the input format is clear.

sjunges:
I agree with your second comment (on having dedicated functions), but this is a seperate issue, isn't it?

sjunges:
Changes LGTM!

PR #645, opened 2024-12-12
Check if Eigen patch was already applied

### LINKED ISSUE #616: Eigen patch step fails on rebuild
When Eigen is rebuilt (for example through changing `CMakeLists.txt`) the patch step through `git apply` fails because the patch was already applied before.

Workaround: remove the directory `build/resources/3rdparty/StormEigen-3.4.1alpha/`.

volkm:
Fixes #616 

`git apply --reverse --check` checks if the patch was already applied (by trying to reverse it) and otherwise applies the patch.


sjunges:
Great! LGTM

PR #649, opened 2024-12-16
Change printing behavior in belief exploration

### LINKED ISSUE #648: POMDP belief exploration prints information by default
When I write, in Stormpy

```
belmc = stormpy.pomdp.BeliefExplorationModelCheckerDouble(pomdp, options)
result = belmc.check(properties[0].raw_formula, [])
```

I get output to the command line:

```
Heuristically selected an under-approximation MDP size threshold of 42801.
Finished exploring under-approximation MDP.
Start analysis...
```

Which should not be the default behavior.

AlexBork:
Changes output in BeliefExplorationPomdpModelChecker to prevent unwanted printing when functions are used in stormpy.

Resolves #648


AlexBork:
On a side note, I think it would make sense to document somewhere what exactly the logging and printing macros do and where to use and avoid using which macros.

volkm:
For the documentation, I would suggest to extend https://github.com/moves-rwth/storm/blob/master/doc/developers.md

volkm (approved):
LGTM

PR #650, opened 2024-12-22
Added two functions for more convenient access to state valuations

tquatmann:
* `prism::Program::getAllExpressionVariables` now has the possibility to not include constants
* added `storage::sparse::StateValuations::getIntegerValues()` and similar for other types, to provide access to all state values of a given variable

### DIFF
--- a/src/storm/storage/sparse/StateValuations.cpp
+++ b/src/storm/storage/sparse/StateValuations.cpp
@@ -164,6 +164,34 @@ storm::RationalNumber const& StateValuations::getRationalValue(storm::storage::s
     return valuation.rationalValues[variableToIndexMap.at(rationalVariable)];
 }
 
+storm::storage::BitVector StateValuations::getBooleanValues(storm::expressions::Variable const& booleanVariable) const {
+    storm::storage::BitVector result(getNumberOfStates(), false);
+    for (uint64_t stateIndex = 0; stateIndex < getNumberOfStates(); ++stateIndex) {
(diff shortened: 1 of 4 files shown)


src/storm/storage/sparse/StateValuations.cpp
```diff
@@ -164,6 +164,34 @@ storm::RationalNumber const& StateValuations::getRationalValue(storm::storage::s
     return valuation.rationalValues[variableToIndexMap.at(rationalVariable)];
 }
 
+storm::storage::BitVector StateValuations::getBooleanValues(storm::expressions::Variable const& booleanVariable) const {
+    storm::storage::BitVector result(getNumberOfStates(), false);
+    for (uint64_t stateIndex = 0; stateIndex < getNumberOfStates(); ++stateIndex) {
+        if (getBooleanValue(stateIndex, booleanVariable)) {
```
volkm, line 170: In this case, I would suggest to directly iterate over the valuations instead of using the convenience function `getBooleanValue` which comes with some overhead. For example, we can look up the index of `booleanVariable` once, instead of doing it for each state.

volkm (commented):
Looks good. I left one comment which could lead to better performance.

src/storm/storage/sparse/StateValuations.cpp
tquatmann: You're absolutely right. Thanks!

PR #651, opened 2025-01-08
Support conversion in function parametric::evaluate

volkm:
Added automatic conversion to the desired return type in `parametric::evaluate`. This allows to replace the common construct
```
storm::utility::convertNumber<ConstantType>(storm::utility::parametric::evaluate(function, valuation));
```
by the simpler version
```
storm::utility::parametric::evaluate<ConstantType>(function, valuation);
```
If the types are equal, then of course no conversion is performed.

Drawback is that the include `constants.h` is now moved to the header file `parametric.h` but I think this is acceptable.


tquatmann (commented):
I'm not sure about this, but if we really want to avoid the additional header inclusion of `constants.h` in `parametric.h`, we could use explicit template instantiations in the cpp file  (for double, cln rationals, and gmp rationals).

Apart from that: LGTM!

volkm:
Thanks for the suggestion @tquatmann. I have added explicit template instantiations and moved the include back to the cpp file.

PR #653, opened 2025-01-14
PcaaWeightVectorChecker: Avoid division by zero bug

tquatmann:


sjunges:
Thanks so much, LGTM. (and well-spotted!!!)

volkm:
The CI error is there since the weekend. My guess it that the newest Clang version has some changes which now result in an error. I opened an [issue](https://github.com/moves-rwth/storm/issues/654).

sjunges:
Do you agree we can merge this?

volkm:
Yes, we can merge it. The failed CI is due to a different issue.

PR #657, opened 2025-01-20
Speed up of MEC decomposition by adjusting loops

j-ide:
I streamlined the decomposition into MECs by preventing to iterate over all remaining candidates for every end component and by keeping track of MECs that contain an accepting end component

### DIFF
--- a/src/storm/modelchecker/helper/ltl/SparseLTLHelper.cpp
+++ b/src/storm/modelchecker/helper/ltl/SparseLTLHelper.cpp
@@ -108,7 +108,7 @@ storm::storage::BitVector SparseLTLHelper<ValueType, Nondeterministic>::computeA
 
         for (uint64_t i = 0; i < transitionMatrix.getRowGroupCount(); i++) {
             uint64_t mec_index = state_to_mec[i];
-            if (mec_index != MAX && is_amec[mec_index]) {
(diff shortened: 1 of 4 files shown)


tquatmann:
Thanks for spotting these! I made another pass over the suggested changes:

For the MECDecomposition, there was still some segmantation fault in the tests. I assume there was some index off-by-one error. I made a rewrite of your optimization with a bit less indexing-magic (thanks to std::span :) )

For the LTL part, I noted that we can even set all states in a MEC to accepting as soon as it contains one EC that is accepting (because the remaining states in the MEC can reach the accepting EC almost surely). However, this also meant that the scheduler generation needed some updates.

Finally, I ran some benchmarks on MDPs with reachability probability / expected reward queries, where there are end-components to eliminate (as a possible preprocessing step before doing value iteration).

image
 
"Old" is the current main branch, "New" is the state of this PR. The first plot shows that this is much faster. The other two plots compare runtime when doing EC elimination or not. The old code (mid) suggested that end component elimination is not always helpful. The new code (right) suggests that it is always a good idea to eliminate end components first (on this kind of MDP benchmarks)

sjunges:
LGTM!  Many thanks!

As an aside @tquatmann should we update the section in the revised practitioner's guide based on this?

Issue #661, opened 2025-02-02
Building JANI with choice labels is not supported


TheGreatfpmK:
Building JANI models with "--buildchoicelab" is not supported.

For example running:
> ./build/bin/storm --jani resources/examples/testfiles/ma/ftwc.jani --buildchoicelab

throws:
```
ERROR (JaniNextStateGenerator.cpp:57): JANI next-state generator cannot generate choice labels.
ERROR (storm-cli.cpp:63): An exception caused Storm to terminate. The message of the exception is: NotSupportedException: JANI next-state generator cannot generate choice labels.
```

sjunges:
I think this should be reasonably easy to add. I will make a PR but I could use some help in testing that code.

TheGreatfpmK:
Sure, I have some JANI MDPs on which I can try it asap.

PR #662, opened 2025-02-02
Jani choicelab

sjunges:
Add choice label building for JANI. Fixes https://github.com/moves-rwth/storm/issues/661


TheGreatfpmK:
I tried the implementation on csma.-2-2 model (both JANI and PRISM can be found here: [csma.2-2.v1.zip](https://github.com/user-attachments/files/18641365/csma.2-2.v1.zip)) and it seems it works mostly fine, however, there is one additional label with no name when parsing JANI.

PRISM model output:
Model type:     MDP (sparse)
States:         1038
Transitions:    1282
Choices:        1054
Reward Models:  none

State Labels:   2 labels
   * deadlock -> 0 item(s)
   * init -> 1 item(s)

Choice Labels:  8 labels
   * send1 -> 12 item(s)
   * send2 -> 12 item(s)
   * end2 -> 20 item(s)
   * cd -> 6 item(s)
   * time -> 844 item(s)
   * busy2 -> 30 item(s)
   * busy1 -> 30 item(s)
   * end1 -> 20 item(s)

JANI model output:
Model type:     MDP (sparse)
States:         1038
Transitions:    1282
Choices:        1054
Reward Models:  none

State Labels:   2 labels
   * deadlock -> 0 item(s)
   * init -> 1 item(s)

Choice Labels:  9 labels
   * send1 -> 12 item(s)
   * send2 -> 12 item(s)
   * end2 -> 20 item(s)
   * cd -> 6 item(s)
   * time -> 844 item(s)
   * busy2 -> 30 item(s)
   *   **-> 80 item(s)**
   * busy1 -> 30 item(s)
   * end1 -> 20 item(s)

TheGreatfpmK:
Ok I see what's happening. If there are empty action labels [] in the PRISM file then the corresponding edges in the JANI file don't have any "action" property. While building models from PRISM these are ignored when it comes to choice labeling, but here a new "empty" label is created for these choices. In this zip [jani-empty-labels-example.zip](https://github.com/user-attachments/files/18866786/jani-empty-labels-example.zip) there are 2 versions of the same model. In tiny_rewards all actions have labels and therefore this is not a problem, in tiny_rewards2 there is one action with no label leading to this behaviour.

src/storm/generator/JaniNextStateGenerator.cpp
```diff
@@ -1070,19 +1073,25 @@ std::vector<Choice<ValueType>> JaniNextStateGenerator<ValueType, StateType>::get
...
+                    if (this->getOptions().isBuildChoiceLabelsSet()) {
+                        if (actionIndex != storm::jani::Model::SILENT_ACTION_INDEX) {
+                            result.back().addLabel(model.getAction(actionIndex).getName());
+                        }
+                    }
                 }
             }
         } else {
             // If the element has more than one set of edges, we need to perform a synchronization.
+            // TODO I (SJ) do not fully understand the meaning of this assertion; after all, it is about the output.
```
tquatmann:
```suggestion
            // We require that some output action for the synchronisation must have been set before.
            // This might be the silent action, if the Jani model does not specify an output action.
```
Hope this clarifies? It's essentially just a sanity check

tquatmann (approved):
LGTM!

PR #669, opened 2025-02-21
Assertions with type check for variables

volkm:
Added assertions to check the variable types.
Was raised in https://github.com/moves-rwth/stormpy/issues/210


tquatmann (approved):
LGTM

PR #675, opened 2025-02-28
Added check that Z3 is available in test

volkm:
Check for z3 was missing


sjunges:
LGTM; but is this just because the model builder needs an SMT solver? :-)

volkm:
Yes, all model building requires Z3 as far as I know

PR #677, opened 2025-03-02
Prism/Update: handle irrational probability expressions

### LINKED ISSUE #673: exponent of power operater must be an integer
hiii there,

i ran storm with a prism model which has math function pow( x, y ), and i got an error saying
"exponent of power operater must be an integer"

but the model seems fine on prism.

// pow func expression in my model
const double dtet_min = 0.0000931;
const double dtet_max = 0.046;
const double dtet_exp = 3.8;
const double dtet_denom = pow(13, dtet_exp);

tquatmann:
Some sanity checks tried to evaluate probability expressions as rational numbers, but the `pow` operator with non-integer coefficients might yield irrational numbers (e.g. pow(2,0.5)). The sanity checks now only apply, if the (simplified) expressions are literals.

We still reject such inputs if Storm is run with `--exact` flag.

Fixes #673

### DIFF
--- a/src/storm/storage/prism/Update.cpp
+++ b/src/storm/storage/prism/Update.cpp
@@ -17,8 +17,12 @@ Update::Update(uint_fast64_t globalIndex, storm::expressions::Expression const& 
       assignments(assignments),
       variableToAssignmentIndexMap(),
       globalIndex(globalIndex) {
-    STORM_LOG_ASSERT(likelihoodExpression.containsVariables() || likelihoodExpression.evaluateAsRational() >= 0,
-                     "Negative likelihood expressions are not allowed.");
+    STORM_LOG_ASSERT(
+        [&likelihoodExpression]() {
+            auto simplified = likelihoodExpression.simplify();
+            return !simplified.isLiteral() || simplified.evaluateAsRational() >= 0;
+        }(),
+        "Negative likelihood expressions are not allowed.");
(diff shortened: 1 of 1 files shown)


src/storm/storage/prism/Update.cpp
```diff
@@ -79,10 +83,9 @@ Update Update::substitute(std::map<storm::expressions::Variable, storm::expressi
     for (auto const& assignment : this->getAssignments()) {
         newAssignments.emplace_back(assignment.substitute(substitution));
     }
-    auto newLikelihoodExpression = this->getLikelihoodExpression().substitute(substitution);
-    STORM_LOG_THROW(likelihoodExpression.containsVariables() || likelihoodExpression.evaluateAsRational() >= 0, storm::exceptions::IllegalArgumentException,
+    auto newLikelihoodExpression = this->getLikelihoodExpression().substitute(substitution).simplify();
+    STORM_LOG_THROW(!newLikelihoodExpression.isLiteral() || newLikelihoodExpression.evaluateAsRational() >= 0, storm::exceptions::IllegalArgumentException,
                     "Substitution yielding negative probabilities in '" << this->getLikelihoodExpression() << "' are not allowed.");
-    // FIXME: The expression could be simplified, but 1/K (where K is an int) is then resolved to 0, which is incorrect (for probabilities).
```
tquatmann, line 85: Note: This old comment no longer applies since the `simplify()` function no longer simplifies divisions with a remainder

sjunges:
LGTM, thanks!

PR #679, opened 2025-03-04
Expression parsing: detect and handle integer overflows

### LINKED ISSUE #652: Properties containing rationals with either the denominator or numerator containing a number >= $2^63$ causes the property not to be recognized
When calling storm on any model with a property such as: `"P>9223372036854775808/9223372036854775808 [F \"goal\"]"` You get the following error:
```
ERROR (SpiritErrorHandler.h:26): Parsing error at 1:3:  expecting <expression>, here:
        P>9223372036854775808/9223372036854775808 [F "goal"]
          ^

ERROR (properties.cpp:54): WrongFormatException: Parsing error at 1:3:  expecting <expression>, here:
        P>9223372036854775808/9223372036854775808 [F "goal"]
          ^
Note that the used API function does not have access to model variables. If the property you tried to parse contains model variables, it will not be parsed correctly.
ERROR (storm-cli.cpp:63): An exception caused Storm to terminate. The message of the exception is: WrongFormatException: WrongFormatException: Parsing error at 1:3:  expecting <expression>, here:
        P>9223372036854775808/9223372036854775808 [F "goal"]
          ^
Note that the used API function does not have access to model variables. If the property you tried to parse contains model variables, it will not be parsed correctly.
```
This can for example be obtained using the command:
`storm -drn storm/resources/examples/testfiles/dtmc/crowds-5-5.drn --prop "P>9223372036854775808/9223372036854775808 [F \"goal\"]"`

I would expect either the error to say that a too large number was found or for it to be accepted.

tquatmann:
When parsing large integers, the changes in this PR yield a more informative error message like this:

```
ERROR (ExpressionCreator.cpp:253): Overflow when parsing integer literal '92348905723462223372036854775807' as a 64 bit integer. Consider appending '.0' to the number to parse it as an (arbitrary precision) float.
ERROR (SpiritErrorHandler.h:26): Parsing error at 1:33:  expecting <no 64-bit integer overflow>, here:
        92348905723462223372036854775807/309779851562500000000000000000000000000000000000000000 [F (observe0 > 1)];
 ```

Fixes #652 

Notes:
- We might get rid of having 64-bit integer literals and just use arbitrary precision numbers in IntegerLiteralExpressions. However, this might affect performance (I'm not sure if that would be noticeable) and requires some more changes at different places.
- Parsing rationals like the one above as RationalLiteralExpressions would be an option. However, I'm a bit unsure if e.g. "4/2" should become an IntegerLiteral or a RationalLiteral, i.e., is it equivalent to "2" or "2.0". I'm also not sure if (in some contexts) integer division has to be considered.

### DIFF
--- a/src/storm-parsers/parser/ExpressionCreator.cpp
+++ b/src/storm-parsers/parser/ExpressionCreator.cpp
@@ -243,9 +243,18 @@ storm::expressions::Expression ExpressionCreator::createRationalLiteralExpressio
     }
 }
 
-storm::expressions::Expression ExpressionCreator::createIntegerLiteralExpression(int64_t value, bool&) const {
-    if (this->createExpressions) {
-        return manager.integer(value);
+storm::expressions::Expression ExpressionCreator::createIntegerLiteralExpression(storm::RationalNumber const& value, bool&, bool& overflow) const {
+    STORM_LOG_ASSERT(storm::utility::isInteger(value), "Expected integer value.");
+    auto const min = storm::utility::convertNumber<storm::RationalNumber>(std::numeric_limits<int64_t>::min());
+    auto const max = storm::utility::convertNumber<storm::RationalNumber>(std::numeric_limits<int64_t>::max());
+    overflow = value < min || value > max;
+    if (overflow) {
+        STORM_LOG_ERROR("Overflow when parsing integer literal '"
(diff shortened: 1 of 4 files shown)


sjunges:
Great! 

I agree that there is a possibility to always use arbitrary-precision integers and I don't think the overhead would be noticable. However, there seem to be more urgent quality-improvements than that one? 

Sebastian

PR #685, opened 2025-03-07
Build without Dd support

### LINKED ISSUE #590: Installation without CUDD
as CUDD seems to be a potential trouble maker: Is there a possibility to build storm without CUDD installed? Will that be feasible?

This would also address https://github.com/moves-rwth/storm/issues/586

volkm:
Fixes #590 

- [x] Introduce CMake options `STORM_USE_CUDD`/ `STORM_USE_SYLVAN` to disable Dd libraries, and use CMake flags `STORM_HAVE_CUDD` / `STORM_HAVE_SYLVAN`.
- [x] Build only with Sylvan
- [x] Build only with CUDD
- [x] Build without any Dd library
- [ ] Extend CI tests to test different configurations -> separate PR


volkm:
I need some input on how to best handle disabling Dd libraries. I see two options:
1. Disabling support completely by removing the template instantiations using `ifdef`. I started doing this in this PR, see [this example](https://github.com/moves-rwth/storm/pull/685/files#diff-8bdc81d8e95a1613e0f079430b65f6af59a893c2e51ad281a10eaa2f68923f01L1615).
2. Only disabling the calls to the Dd library in the `InternalBdd` etc. classes. A similar approach is used for e.g. the [GurobiSolver](https://github.com/moves-rwth/storm/blob/master/src/storm/solver/GurobiLpSolver.cpp#L747).

The first version has the advantage that build times are reduced if the dependency is not available. We also get compile time errors if there are issues with the missing dependency. The disadvantage is that in case of no Dd library, we have to disable all parts using Symbolic representations because we cannot link against any Dd datastructure. An example is the [AbstractModelchecker](https://github.com/moves-rwth/storm/blob/master/src/storm/modelchecker/AbstractModelChecker.cpp#L158) which has a default dummy Dd Representation even for Sparse models. Another example are the tests, which need to be explicitly removed in case of missing dependencies, see [this example](https://github.com/moves-rwth/storm/pull/685/files#diff-83e952b1813a2cd90a7d5944ccd5df10d4dac9fbf37743bf7470c4d159d6e99aR301).

The second version has the advantage that we need fewer changes, because we only need to locally disable the functions. Compilation against the `InternalX` classes can still be done and the tests can for example simply be skipped, if the Dd library is unavailable. The disadvantage is that we still need to build all instantiations even though the library is not available and we might only see issues during runtime (when the exception is thrown) and not at build time.

sjunges:
Hmm.

I think option 1 sound unreasonable right now, although the fact that this is unreasonable shows that we have too tightly integrated some parts of the code base with eachother, i.e., an abstract model checker should not need to know about Dd-based model checkers.  That may, however, be another issue that is not related to the ability to not compile against a dd library. 

I am not too worried about the overhead in compile times. That should be really minimal as these objects can and will be minimal. I am mostly concerned about the disability to figure out that things will not work out properly at compile time and the fact that we are compiling 100 different versions of our code.

volkm:
Thanks for your opinion. I am also leaning towards the second option for now, because it is easier to integrate.

Trying out the first options definitely raised some interesting issues and we now know better what the possible pitfalls are. For now, I will start with compiling against dummy internal Dd libraries.

volkm:
Building with only CUDD, with only Sylvan and without any DD library is now supported.

I will add a separate PR to check all the different configurations. I will try to have a weekly CI job which tests more "exotic" configurations instead of checking all configurations all the time.

CMakeLists.txt
```diff
@@ -499,13 +521,7 @@ endif()
...
 
-export_option(STORM_HAVE_GMM)
-export_option(STORM_HAVE_GUROBI)
-export_option(STORM_HAVE_SPOT)
-export_option(STORM_HAVE_SOPLEX)
-export_option(STORM_HAVE_XERCES)
-export_option(STORM_HAVE_Z3)
-
+# Export options
 include(export)
```
sjunges, line 525: Why move the export_option and include(export) away from eachother?

volkm: I wanted to move everything to to with the 3rdparty libraries to one location and therefore moved the export options as well.
We are also defining export_option in multiple other locations. For instance `STORM_USE_CLN_EA`, `STORM_BUILD_TEST`, etc is defined in other places.

I am open for moving all export_options in one place though.

src/storm-dft/modelchecker/SFTBDDChecker.h
```diff
@@ -32,17 +34,17 @@ class SFTBDDChecker {
     /**
      * \return The internal DFT
      */
-    std::shared_ptr<storm::dft::storage::DFT<ValueType>> getDFT() const noexcept;
```
sjunges, line 35: I wonder what the best way would be to still be able to write noexcept. In the end, for any constructible object, you will never throw an exception on this method....

src/storm/exceptions/MissingLibraryException.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#include "storm/exceptions/BaseException.h"
+#include "storm/exceptions/ExceptionMacros.h"
+
+namespace storm {
+namespace exceptions {
+
+STORM_NEW_EXCEPTION(MissingLibraryException)
```
sjunges, line 9: If we have this functionality, a TODO for our list of issues is to use this error message more consistently?

src/storm/storage/dd/cudd/InternalCuddDdManager.h
```diff
@@ -1,16 +1,18 @@
-#ifndef STORM_STORAGE_DD_INTERNALCUDDDDMANAGER_H_
-#define STORM_STORAGE_DD_INTERNALCUDDDDMANAGER_H_
+#pragma once
+
+#include "storm-config.h"
```
sjunges: in other places, you moved the stormconfig after the libraries....

src/storm/exceptions/MissingLibraryException.h
volkm: Done: https://github.com/moves-rwth/storm/issues/817

src/storm-dft/modelchecker/SFTBDDChecker.h
volkm: In my understanding if Sylvan is unavailable, we can still construct the object (even though it does not do anything) but will then throw an exception.

src/storm/storage/dd/cudd/InternalCuddDdManager.h
volkm: Thanks, good point.

volkm:
Removed the temporary CI tests again. Will merge it after the CI was successful.

PR #686, opened 2025-03-08
Support for macOS builds in CI

### LINKED ISSUE #684: Require newer version of AppleClang
Following #682 it seems we need to require AppleClang >= 15. We should check the minimal compiler version.

volkm:
#### Added CI builds on macOS:
- Xcode 14.3 on macOS 13 (Intel)
- Xcode 15.4 on macOS 14 (ARM)
- Xcode 16.2 on macOS 15 (ARM)

(The CI also shows that the ARM processors nearly halve the compilation times.)

#### Bumped the minimal required compiler versions.
- AppleClang needs at least Xcode version 14.3. XCode version 14.2 fails, see [this test](https://github.com/volkm/storm/actions/runs/13721879189/job/38378974632). Fixes #684
- Bumped Clang and GCC versions to the first versions supporting the flag ` -std=c++20`. The required compiler versions are probably still too low. But I did not want to run extensive tests as the CMake version is most likely the main restriction at the moment anyway.

#### Enabled the stack checks again on AppleClang >= 15.
- AppleClang 14 [still failed with the checks](https://github.com/volkm/storm/actions/runs/13724840840/job/38388686191).

### DIFF
--- a/.github/workflows/buildtest.yml
+++ b/.github/workflows/buildtest.yml
@@ -180,8 +180,8 @@ jobs:
       - name: Run tests
         run: docker exec ci bash -c "cd /opt/storm/build; ctest test --output-on-failure"
 
-  distroTests:
-    name: Distro Tests (${{ matrix.distro }}, ${{ matrix.buildType }})
+  linuxTests:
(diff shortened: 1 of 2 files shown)


sjunges:
LGTM!! Many thanks!

PR #687, opened 2025-03-10
Warning on potential modeling issues in DFT

volkm:
Added a separate function to check for potential modeling issues in DFT. These issues do not prevent analysis but might lead to unexpected results.
We consider two potential issues for now:
- spare modules which are initially activated because they overlap with the top module
- We allow that elements in the primary module of a SPARE gate are also connected to other parts of the fault tree. In an edge case, this can lead to unexpected behavior though, if the SPARE parent is also part of the spare module but not directly connected to the top module. In this instance, the SPARE is never activated.


sjunges:
LGTM!

PR #690, opened 2025-03-20
Set BDD variable ordering for SFT from file

volkm:
The variable ordering for the SFT->BDD translation is based on the topological ordering of the BEs by default. I added the flag `--be-order` which allows to provide a file with the order of BEs which is then used as the variable ordering.


src/storm-dft/parser/BEOrderParser.cpp
```diff
@@ -0,0 +1,58 @@
+#include "BEOrderParser.h"
+
+#include <boost/algorithm/string.hpp>
```
sjunges, line 3: Rather than this, could we use c++20 ranges ?

src/storm-dft/parser/BEOrderParser.cpp
```diff
@@ -0,0 +1,58 @@
+#include "BEOrderParser.h"
```
sjunges, line 1: I think we have a habit of writing includes differently. I would suggest to also do that for consistency (although I don't really mind...)

sjunges:
Minor comments, otherwise LGTM

src/storm-dft/parser/BEOrderParser.cpp
volkm: Done.

src/storm-dft/parser/BEOrderParser.cpp
volkm: I used c++20 ranges now, but it is not as nice as I had hoped.
I could not split on multiple delimiters and I did not find a good replacement for `boost::trim` (but this is not too important as whitespace is removed anyway.)

Happy to receive some suggestions how to do it better, but it should work for now.

volkm: Using ranges lead to new issues. I opt to keep using boost for now and will open a separate issue.

PR #691, opened 2025-03-24
Add cut-off memory node information

### LINKED ISSUE #688: POMDP - Add information about which memory nodes of external FSC was used in cut-offs
This is a small feature request aimed probably at @AlexBork.

Currently, the BeliefExplorationPomdpModelChecker returns the computed result as BeliefExplorationPomdpModelChecker::Result struct which contains DTMC with states/choices labelled so that the strategy can be constructed from it. The problem is that if an external FSC was provided to the BeliefMdpExplorer it will only report that external FSC was used in a cut-off via the label "finite_mem" however no information about which memory node of the FSC was used is provided. This information is crucial to construct the strategy into an FSC, which would be great for the SAYNT algorithm (the current output is a bit of mess with many different objects).

I did some digging and it seems that at the level of belief exploration, this information is stored in the belief MDP https://github.com/moves-rwth/storm/blob/master/src/storm-pomdp/modelchecker/BeliefExplorationPomdpModelChecker.cpp#L1255 in the form of "mem_node_X" label where X is the index of the used memory node, however when the DTMC for the result is created this information is dropped (here  https://github.com/moves-rwth/storm/blob/master/src/storm-pomdp/modelchecker/BeliefExplorationPomdpModelChecker.cpp#L652 and here  https://github.com/moves-rwth/storm/blob/master/src/storm-pomdp/modelchecker/BeliefExplorationPomdpModelChecker.cpp#L541) and is just replaced with the "finite_mem" label. Maybe we can just add this information to the end of the "finite_mem" label?

AlexBork:
This PR extends the labeling for induced Markov chains of belief MDPs with an indication which node of an FSC is used for cut-offs when external cut-off values are provided.
This is in particular important for obtaining the scheduler from the belief exploration, e.g. in the context of SAYNT.

Resolves #688

### DIFF
--- a/src/storm-pomdp/modelchecker/BeliefExplorationPomdpModelChecker.cpp
+++ b/src/storm-pomdp/modelchecker/BeliefExplorationPomdpModelChecker.cpp
@@ -2,7 +2,6 @@
 
 #include <tuple>
 
-#include "../../storm/api/export.h"
 #include "storm-pomdp/analysis/FiniteBeliefMdpDetection.h"
 #include "storm-pomdp/analysis/FormulaInformation.h"
 #include "storm-pomdp/transformer/MakeStateSetObservationClosed.h"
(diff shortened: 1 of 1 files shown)


sjunges:
LGTM!

PR #699, opened 2025-04-14
Retrieving sub-optimal values from Gurobi LP solver

cxlvinchau:
The `GurobiLpSolver` only allows retrieving values of sub-optimal solutions when an optimal solution is found. Is this behaviour intended? If so, then this PR can be ignored. Since Gurobi still stores sub-optimal solutions when no optimal solution is found (within a given timelimit), I have changed the corresponding methods. Thank you!

### DIFF
--- a/src/storm/solver/GurobiLpSolver.cpp
+++ b/src/storm/solver/GurobiLpSolver.cpp
@@ -590,7 +590,7 @@ uint64_t GurobiLpSolver<ValueType, RawMode>::getSolutionCount() const {
 
 template<typename ValueType, bool RawMode>
 ValueType GurobiLpSolver<ValueType, RawMode>::getContinuousValue(Variable const& variable, uint64_t const& solutionIndex) const {
-    if (!this->isOptimal()) {
+    if (!this->currentModelHasBeenOptimized) {
         STORM_LOG_THROW(!this->isInfeasible(), storm::exceptions::InvalidAccessException,
                         "Unable to get Gurobi solution from infeasible model (" << GRBgeterrormsg(**environment) << ").");
         STORM_LOG_THROW(!this->isUnbounded(), storm::exceptions::InvalidAccessException,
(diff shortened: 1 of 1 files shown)


tquatmann:
Hey, thank you for the input!
The current behaviour is indeed intended. We currently provide no interface for terminating an LPSolver prematurely (correct me if I'm wrong?). This means that (with the current code) a non-optimal solution means that something went unexpectedly wrong and should be handled.

I'm closing this for now, but feel free to elaborate on potential use cases of sub-optimal solutions. I'm also happy to have the LPSolver interface extended towards something like a time limit.

cxlvinchau:
Hi Tim,

thanks for getting back to me! We're currently developing a tool for computing witnesses via MILPs which builds on Storm. Quite often the solver seems to be able to obtain good solutions within a reasonable amount of time, but requires a lot of time to prove optimality. Hence, retrieving sub-optimal solutions would be nice :)

As far as I know, Storm itself also uses MILPs to obtain memoryless strategies for multi-objective queries, so perhaps this could also be useful there?

Currently there is indeed no direct way to set a time limit. However, when using the Gurobi solver, it is possible to place a `gurobi.env` file in the working directory (i.e. from which Storm is called), containing a timelimit parameter (see https://support.gurobi.com/hc/en-us/articles/14126085438481-How-do-I-set-parameters-in-Gurobi).

Being somewhat familiar with Gurobi and Storm's LP solver interface, I'd be happy to contribute to such an extension towards a time limit. Feel free to let me know what you think!

tquatmann:
Hi Calvin,

interesting, I wasn't aware of the `gurobi.env` but this is quite convenient to try things out. Thanks!

I agree that a time limit function is useful and I'd  happily accept a PR on this :)

I suggest to add something like `setTimeLimit(uint64_t seconds)` in the `GurobiLPSolver` class. When retrieving a value, one can then assert `currentModelHasBeenOptimized || timeLimitIsSet`.


I'm not sure about infeasible or unbounded cases, though. It seems strange to ask for an infeasible solution(?)

volkm:
FYI: I think you can directly set it via [C++](https://docs.gurobi.com/projects/optimizer/en/current/concepts/parameters/examples.html) without having to use a file.

cxlvinchau:
Hi Tim,

thanks a lot for your suggestion. I am a bit short on time right now, but will be happy to create a PR on this once I have some more time again :)

PR #702, opened 2025-05-06
Propagate STORM_PORTABLE to carl

volkm:
CMake flag `STORM_PORTABLE` was not propagated to carl, yielding issues in Docker containers.


AlexBork (changes requested):
The flag does not properly propagate. There needs to be an intermediate flag added in ``resources/3rdparty/CMakeLists.txt``, in particular Line 341 needs to be extended. This intermediate flag then needs to be used here.

I also suggest adding some output about the ``STORM_PORTABLE`` flag being set, though this might be out of scope of this PR.

AlexBork (approved):
LGTM, thanks for fixing the issue

PR #703, opened 2025-05-07
Added fallback URL for Spot

volkm:
As suggested by the Spot developer [here](https://lists.lre.epita.fr/hyperkitty/list/spot@lrde.epita.fr/message/2FTD6F5XWXD2JSWRMPNZSTOBNCE6KK7R/), we can use both lre.epita.fr and lrde.epita.fr as download options for Spot.
Another possibility would be to add a [Webarchive link](https://web.archive.org/web/*/https://www.lre.epita.fr/dload/spot*), if really needed.

### DIFF
--- a/doc/update_resources.md
+++ b/doc/update_resources.md
@@ -60,7 +60,7 @@ Remove directories that are not needed, e.g, `rm -r doc examples html css benchm
 
 ## Spot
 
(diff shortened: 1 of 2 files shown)


volkm:
Remaining failures in CI are due to #707

Issue #705, opened 2025-05-13
Formula parser fails for formulas with labels that contain special characters and whitespaces


TheGreatfpmK:
In the DRN format, you can have state labels that include special characters and even whitespaces. These are parsed without problem in model parser. For example if you generate DRN from PRISM which used some prism formula to denote goal state you can see a label that looks something like this: "((x = 2) & (y = 0))".

If you try to parse property containing label "((x = 2) & (y = 0))" you will get the following error:

```
ERROR (SpiritErrorHandler.h:26): Parsing error at 1:2:  expecting <basic path formula>, here:
        F "((x = 2) & (y = 0))"]
         ^

ERROR (properties.cpp:53): WrongFormatException: Parsing error at 1:2:  expecting <basic path formula>, here:
        F "((x = 2) & (y = 0))"]
         ^
Note that the used API function does not have access to model variables. If the property you tried to parse contains model variables, it will not be parsed correctly.
ERROR (storm-cli.cpp:63): An exception caused Storm to terminate. The message of the exception is: WrongFormatException: WrongFormatException: Parsing error at 1:2:  expecting <basic path formula>, here:
        F "((x = 2) & (y = 0))"]
```

This is because in the property grammar the label can only contain alphanumeric symbols and underscore.

TheGreatfpmK:
I'm including an example of DRN model with such labels. 

[maze-label-problem.zip](https://github.com/user-attachments/files/20189375/maze-label-problem.zip)

You can try for example the property: Pmax=? [F \"((x = 2) & (y = 0))\"]

sjunges:
We discussed this and we currently do not believe we want to support labels with white spaces. This is really not the idea of the DRN file. The upcoming UMB format will have more robust handling of this. 

Feel free to open the issue to further discuss this.

tquatmann:
Currently, the best workaround is to avoid the expression-style labels in PRISM if you want to export those labels to an explicit format like DRN where variables like `x` and `y` in your case have no meaning. 

Note that things become hairy anyways, even if we would allow such label formulas: On the PRISM model you'd have the formula 

```
Pmax=? [F ((x = 2) & (y = 0))]
```

whereas on the DRN you'd consider (note the double quotes)

```
Pmax=? [F "((x = 2) & (y = 0))"]
```

PR #706, opened 2025-05-13
Fixing formula parser

TheGreatfpmK:
Trying to fix https://github.com/moves-rwth/storm/issues/705. Not sure if there are some edge cases where this would fail. I also did not test this properly but for the example of PRISM formula-like label it works.


tquatmann:
Following the discussions in #705 I'm closing this as we believe the suggested changes are a bit too liberal and error prone in other contexts.

Issue #707, opened 2025-05-13
Multiple out-of-bounds access in vectors


volkm:
The latest GCC/Clang versions enable debug assertion by default and as a result, the tests trigger multiple assertions for out-of-bounds access of a vector. See [this CI run](https://github.com/moves-rwth/storm/actions/runs/14966673272/job/42038055423) for an example.

I list the stacktraces for the failing tests in the following.

### [SparseCtmcCslHelper](https://github.com/moves-rwth/storm/blob/master/src/storm/modelchecker/csl/helper/SparseCtmcCslHelper.cpp#L682):

```
[ RUN      ] StatePermuterTest.ClusterTest
/usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/15.1.1/../../../../include/c++/15.1.1/bits/stl_vector.h:1263: reference std::vector<double>::operator[](size_type) [_Tp = double, _Alloc = std::allocator<double>]: Assertion '__n < this->size()' failed.

Process terminating with default action of signal 6 (SIGABRT)
   at 0xFFCF74C: __pthread_kill_implementation (pthread_kill.c:44)
   by 0xFF75DBF: raise (raise.c:26)
   by 0xFF5D579: abort (abort.c:73)
   by 0xFC17420: std::__glibcxx_assert_fail(char const*, int, char const*, char const*) (assert_fail.cc:41)
   by 0x413919C: std::vector<double, std::allocator<double> >::operator[](unsigned long) (in /opt/storm/build/bin/test-transformer)
   by 0xB1BAD4F: _ZN5storm12modelchecker6helper19SparseCtmcCslHelper29computeTransientProbabilitiesIdLb1ETnNSt9enable_ifIXsr5storm12NumberTraitsIT_EE19SupportsExponentialEiE4typeELi0EEESt6vectorIS5_SaIS5_EERKNS_11EnvironmentERKNS_7storage12SparseMatrixIS5_EEPKSA_S5_S5_SA_S5_ (src/storm/modelchecker/csl/helper/SparseCtmcCslHelper.cpp:682)
[...]

PR #708, opened 2025-05-13
Fixed out-of-bounds access in Fox-Glynn

volkm:
Fixes the first instance of the vector out-of-bounds access in #707

### DIFF
--- a/src/storm/modelchecker/csl/helper/SparseCtmcCslHelper.cpp
+++ b/src/storm/modelchecker/csl/helper/SparseCtmcCslHelper.cpp
@@ -674,7 +674,7 @@ std::vector<ValueType> SparseCtmcCslHelper::computeTransientProbabilities(Enviro
         // The following computes a vector v such that
         // v[i] = foxGlynnResult.totalWeight - sum_{j=0}^{i} foxGlynnResult.weights[j]
         //      = sum_{j=i+1}^{n-1} foxGlynnResult.weights[j]  for i=0,...,n-1
-        // and then sets foxGlynnResult.totalWeight = v / uniformizationRate.
+        // and then sets foxGlynnResult.weights = v / uniformizationRate.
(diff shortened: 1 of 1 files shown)


tquatmann (commented):
I believe there are a few more changes needed: 

- The loop condition needs to be adapted so that the loop is executed `foxGlynnResult.weights.size()` times.
- In the else branch, the decrement of `r` needs to happen after setting weights[r] (otherwise, the last entry of weights would never be set

I suggest this:

```cpp
        // The following computes a vector v such that
        // v[i] = foxGlynnResult.totalWeight - sum_{j=0}^{i} foxGlynnResult.weights[j]
        //      = sum_{j=i+1}^{n-1} foxGlynnResult.weights[j]  for i=0,...,n-1
        // and then sets foxGlynnResult.totalWeight = v / uniformizationRate.
        // We do this in place and with numerical stability in mind. Note that the weights commonly range to values from 1e-200 to 1e+200
        uint64_t l{0ull}, r{foxGlynnResult.weights.size() - 1};
        ValueType sumLeft{storm::utility::zero<ValueType>()}, sumRight{storm::utility::zero<ValueType>()};
        while (l <= r) {
            if (foxGlynnResult.weights[l] < foxGlynnResult.weights[r]) {
                sumLeft += foxGlynnResult.weights[l];
                foxGlynnResult.weights[l] = (foxGlynnResult.totalWeight - sumLeft) / uniformizationRate;
                ++l;
            } else {
                auto const tmp = foxGlynnResult.weights[r];
                foxGlynnResult.weights[r] = sumRight / uniformizationRate;
                sumRight += tmp;
                --r;
            }
        }
```

volkm:
Thanks for the proposed changes, Tim. I integrated them.
There was one remaining issue where at some point `l=r=0` could occur which would then lead to an underflow due to `r--`.

It should be good now, but it would be good if you could give it a second look as well.

tquatmann:
Ohh right, good catch! LGTM

Issue #297, opened 2022-09-14
Bug in Rmax scheduler extraction with Rmax being infinity


sjunges:
When model checking the following MDP and `Rmax=? [F "goal"]` I get `inf` as a result and a scheduler that takes the first action everywhere. If I do `Rmin=?` I seem to get the same scheduler and `5.9...` as a result. If i apply the scheduler using python, i get `5.9...` as a result in both cases.

```
// Exported by storm
// Original model type: MDP
@type: MDP
@parameters

@reward_models
steps 
@nr_states
24
@nr_choices
108
@model
state 0 [0] init
	action 0 [0]
		1 : 0.08333333333
		3 : 0.08333333333
		5 : 0.08333333333
		6 : 0.08333333333
		8 : 0.08333333333
		10 : 0.08333333333
		12 : 0.08333333333
		13 : 0.08333333333
		14 : 0.08333333333
		16 : 0.08333333333
		18 : 0.08333333333
		20 : 0.08333333333
state 1 [1]
	action 0 [0]
		1 : 0.1
		3 : 0.9
	action 1 [0]
		2 : 0.1
		4 : 0.9
	action 2 [0]
		1 : 0.9
		21 : 0.1
	action 3 [0]
		2 : 0.9
		22 : 0.1
	action 4 [0]
		1 : 0.1
		21 : 0.9
	action 5 [0]
		2 : 0.1
		22 : 0.9
	action 6 [0]
		1 : 0.9
		3 : 0.1
	action 7 [0]
		2 : 0.9
		4 : 0.1
state 2 [1]
	action 0 [0]
		1 : 0.1
		3 : 0.9
	action 1 [0]
		2 : 0.1
		4 : 0.9
	action 2 [0]
		1 : 0.9
		21 : 0.1
	action 3 [0]
		2 : 0.9
		22 : 0.1
	action 4 [0]
		1 : 0.1
		21 : 0.9
	action 5 [0]
		2 : 0.1
		22 : 0.9
	action 6 [0]
		1 : 0.9
		3 : 0.1
	action 7 [0]
		2 : 0.9
		4 : 0.1
state 3 [1]
	action 0 [0]
		3 : 0.1
		5 : 0.9
	action 1 [0]
		4 : 0.1
		5 : 0.9
	action 2 [0]
		1 : 0.1
		3 : 0.9
	action 3 [0]
		2 : 0.1
		4 : 0.9
	action 4 [0]
		1 : 0.9
		3 : 0.1
[...]

PR #720, opened 2025-05-25
Fix computing schedulers for max reach. reward properties (fixes #683)

### LINKED ISSUE #683: Incorrect scheduler for MDP expected rewards in the infinity states
We noticed that the scheduler for infinity states in MDPs when computing expected rewards is sometimes incorrect. I'm attaching a simple MDP in PRISM [inf-reward-bad-scheduler-mdp.zip](https://github.com/user-attachments/files/19112963/inf-reward-bad-scheduler-mdp.zip). When you run model checking with property `Rmax=? [F "goal"]` it correctly decides the value to be infinity, however, if you extract the scheduler and apply it to the MDP you get a DTMC where the same property doesn't give you infinity anymore. This is because the scheduler chooses the action `a3` instead of `a4`, which would ensure that the state `s=2` is a trap state.

From our investigation, the culprit is the following function: https://github.com/moves-rwth/storm/blob/8dc90e95a3e8e0b1674752978896a39aff6db8d2/src/storm/utility/graph.cpp#L599-L620

The comment in the function is correct on how this should be computed, but the code is wrong. For me, the trap states should only be those infinity states which are part of some end component where every state is an infinity state. Here line 610 only checks if all the successors of the selected choice are infinity states but does not check whether they are all part of end component.

tquatmann:
I also added the examples from #297 and #683 as test cases.

### DIFF
--- a/resources/examples/testfiles/mdp/tiny_rewards2.nm
+++ b/resources/examples/testfiles/mdp/tiny_rewards2.nm
@@ -1,0 +1,24 @@
+// Adapted from https://github.com/moves-rwth/storm/issues/297
+mdp
+
+module main
+    s : [0..3];
+
+[a0] s=0 -> 0.1: (s'=0) + 0.9: (s'=2);
+[a1] s=0 -> 0.1: (s'=1) + 0.9: (s'=2);
+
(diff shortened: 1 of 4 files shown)


volkm:
LGTM.

PR #724, opened 2025-06-02
Implement Guessing Value Iteration

mahdi-jfri:
Implementing the algorithm proposed in https://arxiv.org/abs/2505.06769.

### DIFF
--- a/src/storm/solver/helper/GuessingValueIterationHelper.cpp
+++ b/src/storm/solver/helper/GuessingValueIterationHelper.cpp
@@ -1,19 +1,16 @@
-#include <numeric>
-#include <queue>
 #include "GuessingValueIterationHelper.h"
 
-#include "storm/environment/solver/OviSolverEnvironment.h"
+#include "storm/storage/SparseMatrix.h"
+
+#include "storm/utility/Extremum.h"
 #include "storm/utility/ProgressMeasurement.h"
-#include "storm/utility/SignalHandler.h"
 #include "storm/utility/vector.h"
 
 #include "storm/exceptions/NotSupportedException.h"
 
 #include "storm/utility/macros.h"
 
-namespace storm {
-namespace solver {
-namespace helper {
+namespace storm::solver::helper {
 
 namespace gviinternal {
 
@@ -35,7 +32,6 @@ IterationHelper<ValueType>::IterationHelper(const storage::SparseMatrix<ValueTyp
         rowIndications.push_back(matrixValues.size());
     }
     rowGroupIndices = &matrix.getRowGroupIndices();
-    //    fillSortedRowGroups()
 }
 
 template<typename ValueType>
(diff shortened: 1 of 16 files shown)


mahdi-jfri:
I believe this is ready. I have tried to conform to the code base's style. Any comments are appreciated.

volkm:
Thank you for contributing the algorithm. We will take a look into it.

src/storm/solver/helper/GuessingValueIterationHelper.cpp
```diff
@@ -0,0 +1,305 @@
...
+    upperX.swap(xy.second);
+}
+
+template<typename ValueType, bool TrivialRowGrouping>
+template<OptimizationDirection Dir>
+std::pair<VerifyResult, SolverStatus> GuessingValueIterationHelper<ValueType, TrivialRowGrouping>::tryVerify(
+    std::vector<ValueType>& lowerX, std::vector<ValueType>& upperX, const std::vector<ValueType>& b, uint64_t& numIterations, IndexType rowGroupToGuess,
+    ValueType guessValue, ValueType precision, std::function<SolverStatus(GVIData<ValueType> const&)> const& iterationCallback) {
+    auto guessLower = lowerX;
+    auto guessUpper = upperX;
```
tquatmann: This creates copies of the large vectors which might mean expensive re-allocations. 
Consider caching these somewhere?

src/storm/solver/helper/GuessingValueIterationHelper.h
```diff
@@ -0,0 +1,96 @@
...
+template<typename ValueType>
+struct GVIData {
+    std::vector<ValueType> const& x;
+    std::vector<ValueType> const& y;
+    SolverStatus const status;
+};
+
+namespace gviinternal {
+
+typedef uint32_t IndexType;
```
tquatmann: Similar algorithms all take 64 indices. What are the consequences of changing this to 64 bits?

src/storm/solver/helper/GuessingValueIterationHelper.h
```diff
@@ -0,0 +1,96 @@
...
+
+    void swipeWeights(std::vector<ValueType>& weights);
+};
+
+}  // namespace gviinternal
+
+template<typename ValueType, OptimizationDirection Dir>
+class GVIBackend;
+
+using IndexType = gviinternal::IndexType;
```
tquatmann: I'd be in favour of avoiding this type definition within the storm::solver::helper namespace. Same for `VerifyResult`

src/storm/solver/helper/GuessingValueIterationHelper.cpp
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    // verification has failed. iterate until convergence
+    return doIterations<Dir>(lowerX, upperX, b, numIterations, {}, precision, iterationCallback);
+}
+
+template<typename ValueType, bool TrivialRowGrouping>
+GuessingValueIterationHelper<ValueType, TrivialRowGrouping>::GuessingValueIterationHelper(
+    std::shared_ptr<ValueIterationOperator<ValueType, TrivialRowGrouping>> viOperator, const storage::SparseMatrix<ValueType>& matrix)
+    : viOperator(viOperator), iterationHelper(matrix) {
```
tquatmann, line 246: I might be overlooking something but the `iterationHelper` changes the stored values.
Does this mean that `solveEquations` can only be called once and if I want to solve multiple equation systems with the same matrix, I need to create a new `GuessingValueIterationHelper`?

If yes, I suggest to have the `iterationHelper` not as a class member. This also allows to put the definition of `class IterationHelper` inside the cpp file.

tquatmann (commented):
Thank you for the contribution! I agree with @sjunges that everything outside the `GuessingValueIterationHelper` looks fine.

src/storm/solver/helper/GuessingValueIterationHelper.cpp
mahdi-jfri: I don't think you need to create a new `GuessingValueIterationHelper`, since it holds no state other than the matrix. So if the matrix stays the same (with the same row groups), then no.

src/storm/solver/helper/GuessingValueIterationHelper.h
mahdi-jfri: None. This was a mistake on my side.

mahdi-jfri:
@tquatmann Thank you so much! I believe I have answered/resolved your comments. Can you please have another look?

src/storm/solver/helper/GuessingValueIterationHelper.cpp
mahdi-jfri: About your suggestion, since the `iterationHelper` stores some information about the matrix I think it is better to have it as a class member. This allows for a faster subsequent calls of `solveEquations` on the same matrix.

mahdi-jfri:
@tquatmann Thank you so much! I believe I have resolved/answered to your comments. Can you please have another look?

src/storm/solver/helper/GuessingValueIterationHelper.cpp
tquatmann: Ahh I see. But what is the purpose of essentially storing a copy of the input matrix inside the `IterationHelper` without modifying it? Couldn't you just keep the matrix reference and use that for your computations? 

If I see this correctly, all member functions of the `IterationHelper` could be `const`, right?

mahdi-jfri: You are right. I removed the unnecessary stuff from `IterationHelper`. It now only stores the matrix reference, and a vector cache.

Because of the cache I cannot make `selectRowGroupToGuess` `const`.

mahdi-jfri:
@tquatmann @volkm Thank you for your time. Can you please tell me what the next steps are?

tquatmann:
I'll have a final look this week and will come back to you soon :)

tquatmann:
I looked through the code again and things look fine now :) 

Out of curiosity, I also tested the code on a benchmark set assembled in [this](https://sjunges.github.io/files/revised_practitioners_guide.pdf) paper (there referred to as the `practitioners` set). The implementation looks fine and has performance compared to OVI. Here's a summary of the results in two plots:

image

From my side, we can merge this. Thanks again for the contribution!

PR #728, opened 2025-06-16
Fixes for (negative) expected rewards

tquatmann:
This PR fixes a few bugs and improves behaviour based on feedback by @Chickenpowerrr (Thanks a lot!)

- Throw an error for currently unsupported (single-objective) total reward computation with negative rewards
- Warn users about potentially discarded "infinitely bad" solutions
- Properly handle multi-objective queries involving negative and mixed-sign rewards
- Mildly improve upper reward bound computation: Rewards on transitions that can only be triggered once can also only be collected once.

### DIFF
--- a/resources/examples/testfiles/mdp/tiny_rewards_negative.nm
+++ b/resources/examples/testfiles/mdp/tiny_rewards_negative.nm
@@ -1,0 +1,19 @@
+mdp
+
+module model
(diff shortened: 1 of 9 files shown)


sjunges:
LGTM. If you still have the examples that demonstrated these bugs, it may be good to add them to the tests? :-)

PR #730, opened 2025-06-20
Remove intel tbb

### LINKED ISSUE #133: Newer versions of Intel TBB are incompatible
There have been a couple of API changes in intel thread building blocks which apparently makes it incompatible with Storm right now.

For example, our `findTBB.cmake` appears to not work on macOS anymore (at least for me). However, there is a `TBBConfig.cmake` shipped with TBB nowadays...

Moreover, the file `tbb/tbb_stddef.h` included by Storm is no longer available.

Anyway... my proposal would be to (at some point) drop support for TBB entirely in favor of parallel algorithms that were added in `c++17` as these are also available on non-intel machines.

sjunges:
Fixes #133


volkm (approved):
LGTM. Thanks.

PR #731, opened 2025-06-27
Bumped Spot to version 2.14

volkm:
[Changelog](https://gitlab.lre.epita.fr/spot/spot/blob/spot-2-14/NEWS)


sjunges:
LGTM. Do we also need to update the requirements on preinstalled versions? No, right?

PR #732, opened 2025-06-27
Creating Debian packages

volkm:
Revised the existing CPack configuration to build .deb packages for Debian and Ubuntu and moved the file to `resources/cmake`.
I also added a CI workflow which builds the Debian packages, stores them as artifacts and tests them by building the starter-project.

~~Creating the packages is successful but the installation still has an issue. As seen [here](https://github.com/volkm/storm/actions/runs/15923695623/job/44916057677), the storm libraries are not found afterwards.
A couple of observations:~~
- ~~ldd on /usr/lib/storm/libstorm.so shows that all reference can be resolved, ldd on /usr/bin/storm shows that the storm libs cannot be found~~
- ~~the libraries are installed in /lib/storm/, /bin/, and /usr/lib/storm, /usr/include/storm, /usr/bin. The carl installation is different from the rest because carl is installed in /usr/local/lib/storm/resources/ and /usr/local/include/storm/resources/carl/.~~

Edit: creating packages should now be successful after integrating the following changes:
- https://github.com/stormchecker/carl-storm/pull/131
- https://github.com/stormchecker/storm/pull/901

I tested a combined version of all proposed changes in https://github.com/volkm/storm/tree/installation-test.
- Build tests [are successful](https://github.com/volkm/storm/actions/runs/23564620541)
- Creating Debian packages [is successful](https://github.com/volkm/storm/actions/runs/23564630342) as well and the resulting Debian packages are available as artifacts of the same [run](https://github.com/volkm/storm/actions/runs/23564630342).

### DIFF
--- a/.github/workflows/release_deb.yml
+++ b/.github/workflows/release_deb.yml
@@ -1,0 +1,75 @@
+# Prepares a deb package of Storm
+name: Release deb package
+
+on:
+  # needed to trigger the workflow manually
+  workflow_dispatch:
+
+env:
+  # GitHub runners currently have 4 cores
+  NR_JOBS: "4"
+
+jobs:
+  debian:
+    name: Debian package on ${{ matrix.distro.name }}-${{ matrix.platform.name }}
+    runs-on: ${{ matrix.platform.runner }}
+    strategy:
+      matrix:
+        distro:
+          - {name: "debian:12",
+             tag: "debian-12",
+             build_type: "Release"
+            }
+          - {name: "debian:13",
+             tag: "debian-13",
+             build_type: "Release"
+            }
+          - {name: "ubuntu:24.04",
+             tag: "ubuntu-24.04",
+             build_type: "Release"
+            }
+          - {name: "ubuntu:rolling",
+             tag: "latest",
+             build_type: "Release"
+            }
+        platform:
+          - {name: amd64, runner: "ubuntu-latest"}
+          - {name: arm64, runner: "ubuntu-24.04-arm"}
+      fail-fast: false
+    steps:
+      - name: Git clone
+        uses: actions/checkout@v6
+      - name: Build storm from Dockerfile
+        run: |
+          docker build -t movesrwth/storm:ci . \
+            --build-arg BASE_IMAGE=movesrwth/storm-basesystem:${{ matrix.distro.tag }} \
+            --build-arg build_type="${{ matrix.distro.build_type }}" \
+            --build-arg carl_tag="master" \
+            --build-arg cmake_args="-DSTORM_BUILD_TESTS=OFF" \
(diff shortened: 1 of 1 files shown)


resources/cmake/stormCPackConfig.cmake
```diff
@@ -1,31 +1,21 @@
...
-set(CPACK_GENERATOR "ZIP")
+set(CPACK_RESOURCE_FILE_README "${CMAKE_SOURCE_DIR}/README.md")
 set(CPACK_PACKAGE_INSTALL_DIRECTORY "${CPACK_PACKAGE_NAME}-${CPACK_PACKAGE_VERSION}")
 
-### source package settings
-set(CPACK_SOURCE_GENERATOR "ZIP")
+set(CPACK_GENERATOR "DEB")
+
+### Source package settings
+set(CPACK_SOURCE_GENERATOR "TGZ")
```
sjunges, line 20: The source generator is no longer used? But then we should probably clear what follows here and below?

volkm:
Installing from the deb package actually places the binaries in both /bin and /usr/bin, similar happens for the libs. Maybe putting them in /bin and /lib can be disabled.

Providing a separate package for carl is actually a good idea. It is a dependency and therefore should be separate package in my view.

The debian package is created from cmake via `make package`. It is based on the build tree but as far as I understand it should follow the same structure as the installation.

resources/cmake/stormCPackConfig.cmake
volkm: Do we want to have the possibility to create a source package? Or is it sufficient that any Github release already comes with an archive of the sources? I think it should be fine to remove the support for a source package here.

volkm: I did not find an option to disable source package generation. Without any settings, the source package is still possible, but includes e.g. the build directory as well. So I opted to keep the source package option but added a comment that we recommend to use `git archive` instead.

.github/workflows/release_deb.yml
```diff
@@ -0,0 +1,75 @@
...
+  # GitHub runners currently have 4 cores
+  NR_JOBS: "4"
+
+jobs:
+  debian:
+    name: Debian package on ${{ matrix.distro.name }}-${{ matrix.platform.name }}
+    runs-on: ${{ matrix.platform.runner }}
+    strategy:
+      matrix:
+        distro:
```
sjunges, line 18: Is ubuntu:rolling the same as ubuntu 25.10 right now?

volkm: [Yes](https://hub.docker.com/_/ubuntu#supported-tags-and-respective-dockerfile-links)

volkm: 26.04 will be out on April 23rd

Issue #734, opened 2025-07-04
Conditional Probabilities can not be computed / Bitvector not of same length


RobinZiemek:
Hello Storm Team,

in issue [#509](https://github.com/moves-rwth/storm/issues/509#issue-2158797232) the value for `Pmin=? [F x=2 || x=1]` is to be computed for the program 
```
mdp
module main
	x : [0..2];
	[] x=0 -> 0.7 : (x'=1) + 0.3 : (x'=2);
	[] x=1 -> 1 : (x'=1);
	[] x=2 -> 1 : (x'=1);
endmodule
```
When running this program with the current dev version of storm the following output is produced:
```
Storm 1.10.0 (dev)
DEBUG BUILD

Date: Fri Jul  4 10:32:48 2025
Command line arguments: --prism example.pm --prop 'Pmin=? [ F x=2 || F x=1 ]'
Current working directory:

Time for model input parsing: 0.014s.

Time for model construction: 0.043s.

-------------------------------------------------------------- 
Model type: 	MDP (sparse)
States: 	3
Transitions: 	4
Choices: 	3
Reward Models:  none
State Labels: 	4 labels
   * init -> 1 item(s)
   * deadlock -> 0 item(s)
   * (x = 2) -> 1 item(s)
   * (x = 1) -> 1 item(s)
Choice Labels: 	none
-------------------------------------------------------------- 

Model checking property "1": Pmin=? [(F (x = 2)) || (F (x = 1))] ...
ERROR (BitVector.cpp:482): Length of the bit vectors does not match.
Assertion failed: (bitCount == other.bitCount), function isDisjointFrom, file BitVector.cpp, line 482.
ERROR: The program received signal 6 and will be aborted in 3s.
zsh: abort      storm --prism example.pm --prop 'Pmin=? [ F x=2 || F x=1 ]'
```

Debugging the problem with a minimal program in CLion shows where the call to "isDisjointFrom" is comming from. The stack of calls is:
```
storm::storage::BitVector::isDisjointFrom(const storm::storage::BitVector &) const BitVector.cpp:482
[...]

PR #735, opened 2025-07-04
Conditional Probabilities for MDPs

### LINKED ISSUE #509: Incorrect minimal conditional probabilities.
When computing `Pmin=? [F x=2 || x=1]`  for the following PRISM Program, with Storm we get `0`, although `0.3` should be correct.
```
mdp
module main
	x : [0..2];
	[] x=0 -> 0.7 : (x'=1) + 0.3 : (x'=2);
	[] x=1 -> 1 : (x'=1);
	[] x=2 -> 1 : (x'=1);
endmodule
```

```console
Storm 1.8.2 (dev)

Date: Wed Feb 28 12:55:15 2024
Command line arguments: --prism conditional.nm -prop 'Pmin=? [F x=2 || F x=1 ]'
Current working directory: 

Time for model input parsing: 0.000s.

Time for model construction: 0.010s.

Model type: 	MDP (sparse)
States: 	3
Transitions: 	4
Choices: 	3
Reward Models:  none
State Labels: 	4 labels
   * init -> 1 item(s)
   * deadlock -> 0 item(s)
   * (x = 2) -> 1 item(s)
   * (x = 1) -> 1 item(s)
Choice Labels: 	none

Model checking property "1": Pmin=? [(F (x = 2)) || (F (x = 1))] ...
Result (for initial states): 0
```

The problem seems to be an incorrect reduction from minimizing to maximizing queries.

tquatmann:
Contains a (re-)implementation of the restart-method for computing conditional probabilities for MDPs.

Also introduces algorithms `bisection`, `bisection-advanced` and `pi` as alternatives for the default restart algorithms. These alternatives work well when the model is acyclic.

Fixes #509 and #734 .


tquatmann:
I've added some test cases, so this is now ready to be reviewed :)

src/storm/environment/modelchecker/ModelCheckerEnvironment.h
```diff
@@ -24,6 +25,9 @@ class ModelCheckerEnvironment {
     SteadyStateDistributionAlgorithm getSteadyStateDistributionAlgorithm() const;
     void setSteadyStateDistributionAlgorithm(SteadyStateDistributionAlgorithm value);
 
+    ConditionalAlgorithm getConditionalAlgorithm() const;
+    void setConditionalAlgorithm(ConditionalAlgorithm value);
```
sjunges: It doesnt matter performance-wise, but do we pass enums by value or by ref?

src/storm/modelchecker/helper/conditional/ConditionalAlgorithm.cpp
```diff
@@ -0,0 +1,36 @@
+#include "ConditionalAlgorithm.h"
+
+namespace storm {
+std::ostream& operator<<(std::ostream& stream, ConditionalAlgorithm const& algorithm) {
+    switch (algorithm) {
+        case ConditionalAlgorithm::Default:
```
sjunges: Is default really an algorithm, or is this a setting?

src/storm/modelchecker/prctl/SparseMdpPrctlModelChecker.cpp
```diff
@@ -264,6 +265,8 @@ std::unique_ptr<CheckResult> SparseMdpPrctlModelChecker<SparseMdpModelType>::com
                     "Formula needs to specify whether minimal or maximal values are to be computed on nondeterministic model.");
     STORM_LOG_THROW(this->getModel().getInitialStates().getNumberOfSetBits() == 1, storm::exceptions::InvalidPropertyException,
                     "Cannot compute conditional probabilities on MDPs with more than one initial state.");
+    STORM_LOG_THROW(checkTask.isOnlyInitialStatesRelevantSet(), storm::exceptions::InvalidPropertyException,
+                    "Conditional probabilities can only be computed for the initial states of the model.");
```
sjunges, line 269: Is this also true for the restart algorithm? (I am fine if this is implemented this way, but would be good to understand/clarify)

src/storm/settings/modules/ModelCheckerSettings.cpp
```diff
@@ -7,13 +7,15 @@
 #include "storm/settings/SettingMemento.h"
 #include "storm/settings/SettingsManager.h"
 
-namespace storm {
-namespace settings {
-namespace modules {
+#include "storm/exceptions/NotImplementedException.h"
```
sjunges: why do we need these headers now?

src/storm/modelchecker/helper/conditional/ConditionalHelper.cpp
```diff
@@ -0,0 +1,772 @@
...
+    // There are three cases of terminal states:
+    // 1. conditionStates: The condition holds, so the target value is the optimal probability to reach target from there
+    // 2. targetStates: The target is reached, so the target value is the optimal probability to reach a condition from there.
+    //                  The remaining probability mass is the probability of an observation failure
+    // 3. states that can not reach the condition under any scheduler. The target value is zero.
+
+    // TerminalStates is a superset of conditionStates and dom(nonZeroTargetStateValues).
+    // For a terminalState that is not a conditionState, it is impossible to (reach the condition and not reach the target).
+
+    ValueType targetValue(uint64_t state) const {
```
sjunges: I know this is more c++ style than what we usually do, but would expect in this code base that this is called getTargetValue?

src/storm/modelchecker/helper/conditional/ConditionalHelper.cpp
```diff
@@ -0,0 +1,772 @@
...
+        storm::utility::graph::performProb1A(transitionMatrix, transitionMatrix.getRowGroupIndices(), backwardTransitions, allStates, conditionStates);
+    auto universalObservationFailureStates = storm::utility::graph::performProb0A(backwardTransitions, allStates, extendedConditionStates);
+    std::map<uint64_t, ValueType> nonZeroTargetStateValues;
+    auto const extendedTargetStates =
+        storm::utility::graph::performProb1A(transitionMatrix, transitionMatrix.getRowGroupIndices(), backwardTransitions, allStates, targetStates);
+    computeReachabilityProbabilities(env, nonZeroTargetStateValues, dir, transitionMatrix, extendedConditionStates, allStates, extendedTargetStates);
+    auto const targetAndNotCondFailStates = extendedTargetStates & ~(extendedConditionStates | universalObservationFailureStates);
+    computeReachabilityProbabilities(env, nonZeroTargetStateValues, dir, transitionMatrix, targetAndNotCondFailStates, allStates, extendedConditionStates);
+
+    auto terminalStates = extendedConditionStates | extendedTargetStates | universalObservationFailureStates;
```
sjunges: not const?

sjunges: Ah, I see. Maybe swap order with terminStatesThatReachCondition

src/storm/modelchecker/helper/conditional/ConditionalHelper.cpp
```diff
@@ -0,0 +1,772 @@
...
+
+    if (storm::solver::minimize(dir)) {
+        // There can be target states from which (only) the *minimal* probability to reach a condition is zero.
+        // For those states, the optimal policy is to enforce observation failure.
+        // States that can only reach (target states with almost sure observation failure) or observation failure will be treated as terminal states with
+        // targetValue zero and failProbability one.
+        terminalStates |= storm::utility::graph::performProb0A(backwardTransitions, ~terminalStates, terminalStatesThatReachCondition);
+    }
+
+    auto nonTerminalStates = ~terminalStates;
```
sjunges, line 205: again, maybe swap order so that modification of nonTerminalSttes is in front of...

src/storm/modelchecker/helper/conditional/ConditionalHelper.cpp
```diff
@@ -0,0 +1,772 @@
...
+std::unique_ptr<CheckResult> computeConditionalProbabilities(Environment const& env, storm::solver::SolveGoal<ValueType, SolutionType>&& goal,
+                                                             storm::storage::SparseMatrix<ValueType> const& transitionMatrix,
+                                                             storm::storage::SparseMatrix<ValueType> const& backwardTransitions,
+                                                             storm::storage::BitVector const& targetStates, storm::storage::BitVector const& conditionStates) {
+    // We might require adapting the precision of the solver to counter error propagation (e.g. when computing the normal form).
+    auto normalFormConstructionEnv = env;
+    auto analysisEnv = env;
+    if (env.solver().isForceSoundness()) {
+        normalFormConstructionEnv.solver().minMax().setPrecision(env.solver().minMax().getPrecision() *
+                                                                 storm::utility::convertNumber<storm::RationalNumber, std::string>("1/10"));
```
sjunges: Magic constant?

src/storm/modelchecker/helper/conditional/ConditionalHelper.cpp
```diff
@@ -0,0 +1,772 @@
...
+        analysisEnv.solver().minMax().setPrecision(env.solver().minMax().getPrecision() *
+                                                   storm::utility::convertNumber<storm::RationalNumber, std::string>("9/10"));
+    }
+
+    // We first translate the problem into a normal form.
+    // @see doi.org/10.1007/978-3-642-54862-8_43
+    STORM_LOG_THROW(goal.hasRelevantValues(), storm::exceptions::NotSupportedException,
+                    "No initial state given. Conditional probabilities can only be computed for models with a single initial state.");
+    STORM_LOG_THROW(goal.relevantValues().getNumberOfSetBits() == 1, storm::exceptions::NotSupportedException,
```
sjunges: Side comment. I guess we check often whether the number of setBits is exaclty one, maybe we should have a method for that?

src/storm/environment/modelchecker/ModelCheckerEnvironment.h
tquatmann: We've been doing this mostly by ref in the past, but imo it is not reasonable to do or even enforce doing that:

- Performance-wise, there is no practical difference. Arguably, by-ref is slightly worse as it requires 8 bytes and an additional indirection whereas `enum class`es are essentially `int`s and require 4 bytes (on standard machines).
- Semantics-wise, there is always the chance that unexpected side effects might happen. For example, the reference might get invalidated in the function body because of a vector re-allocation or whatever. Of course, not the case here but generally avoiding references (whenever they are not the intended semantics) means safer code.

src/storm/modelchecker/prctl/SparseMdpPrctlModelChecker.cpp
tquatmann: To my understanding, the restart algorithm must add a backloop to *the* initial state. Is there a variant that handles multiple initial states?

src/storm/modelchecker/helper/conditional/ConditionalHelper.cpp
tquatmann: Do you think this should be a setting? For now, I just made this more explicit

```cpp
        // We intuitively have to divide the precision into two parts, one for computations when constructing the normal form and one for the actual analysis.
        // As the former is usually less numerically challenging, we use a factor of 1/10 for the normal form construction and 9/10 for the analysis.
        auto const normalFormPrecisionFactor = storm::utility::convertNumber<storm::RationalNumber, std::string>("1/10");
```

src/storm/settings/modules/ModelCheckerSettings.cpp
tquatmann: Thx, removed them.

src/storm/modelchecker/helper/conditional/ConditionalAlgorithm.cpp
tquatmann: I changed the enum to `ConditionalAlgorithmSetting`

src/storm/modelchecker/helper/conditional/ConditionalHelper.cpp
tquatmann: absolutely, yes. I changed that.

src/storm/modelchecker/helper/conditional/ConditionalHelper.cpp
tquatmann: The next line uses `nonTerminalStates`. Changing the order would change the behaviour.

src/storm/modelchecker/helper/conditional/ConditionalHelper.cpp
tquatmann: added `bool BitVector::hasUniqueSetBit() const`

tquatmann:
Thanks for the great comments! :)


I should mention that this PR right now breaks support for conditional probabilities for interval MDPs (which in general were also not handled correctly before either).

src/storm/modelchecker/prctl/SparseMdpPrctlModelChecker.cpp
RobinZiemek: According to the publications [[1](https://nbn-resolving.org/urn:nbn:de:bsz:14-qucosa2-743742), [2](https://link.springer.com/chapter/10.1007/978-3-642-54862-8_43)] the restart method crucially relies on a single initial state.
Handling multiple initial states would then (in theory) be handled by adding the typical fresh initial state which has the outgoing initial distribution (or choices).
From my POV such a feature would not be needed.

sjunges: Ah, ofcourse...  wasn't thinking properly.

src/storm/modelchecker/helper/conditional/ConditionalHelper.cpp
sjunges: If there were options, I would set this hardcoded in the options, such that people can overrride the options if they want to. Not essential though.

src/storm/environment/modelchecker/ModelCheckerEnvironment.h
sjunges: I think these are solid arguments. The main argument I know against this is that you must know whether it is an integral type or not and that this is ever changed, one must check the code base.

tquatmann: Also a good point. There's no silver bullet and I don't think we have been consistent so far.

src/storm/modelchecker/helper/conditional/ConditionalHelper.cpp
tquatmann: Yes, maybe there will be some more options introduced in the future and then I'm fine to add that.

sjunges:
TLDR regarding interval models: Interval MCs are fine (which was the use case). We can enable that in another PR.

LGTM

PR #736, opened 2025-07-07
Time limit for Gurobi LP Solver

cxlvinchau:
This PR adds the possibility to specify a time limit when using the Gurobi LP solver. This allows for stopping the solver before an optimal solution is found and retrieving sub-optimal solutions.

The following changes have been made:
- The following methods have been added to `GurobiLpSolver`: 
    * `setTimeLimit(uint64_t seconds)`: sets the parameter in the underlying Gurobi environment of the model
    * `hasTimeLimit()`: returns whether a time limit has been set
    * `getTimeLimit()`: returns the time limit if it has been set, otherwise an `InvalidAccessException` is thrown
    * `hasTimedOut()`: returns whether a time out has occurred (implemented analogously to the check of optimality)
- The methods for retrieving sub-optimal solutions have been adapted, i.e. `getContinuousValue(...)`, `getIntegerValue(...)`, `getBinaryValue(...)` and `getObjectiveValue(...)`. More precisely, the checks for infeasiblity and unboundedness have been removed. If the problem is infeasible, then the check of the solution count will be zero. If the problem is unbouded, Gurobi might still find a solution before it determines that the problem is indeed unbounded. In that case a user might still be interested in a solution.

### DIFF
--- a/src/storm/solver/GurobiLpSolver.h
+++ b/src/storm/solver/GurobiLpSolver.h
@@ -131,8 +131,9 @@ class GurobiLpSolver : public LpSolver<ValueType, RawMode> {
     bool getBinaryValue(Variable const& name, uint64_t const& solutionIndex) const;
     ValueType getObjectiveValue(uint64_t solutionIndex) const;
 
-    // Method for specifying and retrieving a time limit
+    // Method for specifying a time limit
     void setTimeLimit(uint64_t seconds);
+    // Retrieve the time limit in seconds. Requires that a time limit has been set before.
     uint64_t getTimeLimit();
     // Method for checking whether the model has a time limit
     bool hasTimeLimit();
--- a/src/storm/solver/GurobiLpSolver.h
+++ b/src/storm/solver/GurobiLpSolver.h
@@ -131,8 +131,14 @@ class GurobiLpSolver : public LpSolver<ValueType, RawMode> {
     bool getBinaryValue(Variable const& name, uint64_t const& solutionIndex) const;
     ValueType getObjectiveValue(uint64_t solutionIndex) const;
 
-    // Method for specifying a time limit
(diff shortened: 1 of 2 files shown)


sjunges:
Hey Calvin,

thanks for your contribution!

The code change looks good to me. I may want to know however whether the solution is actually optimal (which was guaranteed before, up to Gurobi imprecisions). With the current version, I can not know this if I set a timeout?  I think it would also be great if we can check whether a time limit has been set.

@tquatmann anything I am overlooking here?

Sebastian

sjunges:
You make some good points and you are absolutely right. isOptimal should do. 
I was mostly thinking that when getting a result, users *should* actually check if the result is indeed optimal if they need this. 

Do you think you'll have time to add the hasTimeLimit option? :-)

cxlvinchau:
Sure, I'd be happy to! :)

sjunges:
LGTM, many thanks. @tquatmann Do you agree with merge?

src/storm/solver/GurobiLpSolver.h
```diff
@@ -131,6 +131,14 @@ class GurobiLpSolver : public LpSolver<ValueType, RawMode> {
     bool getBinaryValue(Variable const& name, uint64_t const& solutionIndex) const;
     ValueType getObjectiveValue(uint64_t solutionIndex) const;
 
+    // Method for specifying and retrieving a time limit
+    void setTimeLimit(uint64_t seconds);
+    uint64_t getTimeLimit();
```
tquatmann:
```suggestion
    // Method for specifying a time limit
    void setTimeLimit(uint64_t seconds);
    // Retrieve the time limit in seconds. Requires that a time limit has been set before.
    uint64_t getTimeLimit();
```

Issue #469, opened 2023-12-10
Support for interval models


sjunges:
With https://github.com/moves-rwth/storm/pull/409, storm has now preliminary support for intervals. This is by no means stable. Some important feature that we are still missing are:

 - [ ] Explicit support for solving on DTMCs -- the DTMCs must be declared as MDPs
 - [ ] Support for rewards
 - [ ] Bounded reachability properties
 - [ ] Topological VI
 - [ ] Convenience functions to remove uncertainty (e.g., by taking the 'center').
 - [ ] support for robustness in formulas
 - [ ] Generating interval models from PRISM/JANI

volkm:
I think at least support for parsing interval models from Prism/JANI would be nice to have soonish.

PR #739, opened 2025-07-22
Support PRISM models with Intervals and CLI Extensions

tquatmann:
This PR adds support for building prism models with intervals as documented [here](https://www.prismmodelchecker.org/manual/PropertySpecification/UncertainModels)

It also extends/revises model-handling for CLI to incorporate intervals.

Addresses parts of #469


tquatmann:
I had to do some more extensive CLI updates in `model-handling.h` to incorporate intervals as `ValueType`.

The old code had to set the right `ValueType`s  for model building and verification  rather early.
This is becoming increasingly messy, in particular with the following things in mind:

- DRN models where we only realise during parsing whether intervals occur in the model
- Preprocessing that turn interval models into non-interval models or vice versa

We now use `ModelBase` (thus not requiring a `ValueType`) during the higher-level parts of model handling. The down-casting is done with some callback (`castAndApply`) mechanism. This required some changes in storm-pars and storm-pomdp, as well.

src/storm/adapters/RationalNumberForward.h
```diff
@@ -30,8 +30,38 @@ typedef cln::cl_I ClnIntegerNumber;
 typedef __gmp_expr<mpq_t, mpq_t> GmpRationalNumber;
 typedef __gmp_expr<mpz_t, mpz_t> GmpIntegerNumber;
 #endif
+
+/*!
+ * Interval type
+ */
 typedef carl::Interval<double> Interval;
 
+namespace detail {
```
sjunges, line 39: Do we maybe want to put this into an own interval-header? 

Advantages:

- Probably, no one will look in rationalnumber for interval based stuff
- The places that need intervals and need rationals need not coincide

Downsides:
- More headers
- With exact intervals on the way, the distinction is less clear.

sjunges: If we keep it here, maybe we should call it NumberRepresentationForward.h ? (and modify the other headers as well?)

src/storm/generator/PrismNextStateGenerator.cpp
```diff
@@ -592,6 +593,22 @@ PrismNextStateGenerator<ValueType, StateType>::getActiveCommandsByActionIndex(ui
...
+    if constexpr (storm::IsIntervalType<ValueType>) {
+        if (update.isLikelihoodInterval()) {
+            BaseValueType lower = evaluator.asRational(update.getLikelihoodExpressionInterval().first);
+            BaseValueType upper = evaluator.asRational(update.getLikelihoodExpressionInterval().second);
+            return ValueType(lower, carl::BoundType::WEAK, upper, carl::BoundType::WEAK);
+        } else {
+            return evaluator.asRational(update.getLikelihoodExpression());
+        }
+    } else {
+        STORM_LOG_ASSERT(!update.isLikelihoodInterval(), "Unexpected interval likelihood for non-interval model in update " << update << ".");
```
sjunges, line 608: We should probably also warn about this with exploration checks on.

src/storm/generator/PrismNextStateGenerator.cpp
```diff
@@ -592,6 +593,22 @@ PrismNextStateGenerator<ValueType, StateType>::getActiveCommandsByActionIndex(ui
     return result;
 }
 
+template<typename ValueType, typename BaseValueType>
+ValueType evaluateLikelihoodExpression(storm::prism::Update const& update, storm::expressions::ExpressionEvaluator<BaseValueType> const& evaluator) {
+    if constexpr (storm::IsIntervalType<ValueType>) {
+        if (update.isLikelihoodInterval()) {
+            BaseValueType lower = evaluator.asRational(update.getLikelihoodExpressionInterval().first);
+            BaseValueType upper = evaluator.asRational(update.getLikelihoodExpressionInterval().second);
+            return ValueType(lower, carl::BoundType::WEAK, upper, carl::BoundType::WEAK);
```
sjunges, line 602: An assertion/check with -ec on the lower bound being at least zero and the upper bound being at most one would probably be helpful (although one can argue where to do this?)

src/storm/models/ModelBase.h
```diff
@@ -99,6 +101,19 @@ class ModelBase : public std::enable_shared_from_this<ModelBase> {
      */
     virtual bool isSymbolicModel() const;
 
+    /*!
```
sjunges, line 104: Why do we need this for the supergeneric models? This sounds like something I would only want in DD models?

src/storm/models/symbolic/Model.h
```diff
@@ -314,6 +314,10 @@ class Model : public storm::models::Model<CValueType> {
 
     virtual bool isSymbolicModel() const override;
 
+    virtual std::optional<storm::dd::DdType> getDdType() const override;
+
+    virtual bool isExact() const override;
```
sjunges, line 319: 👍

src/storm/storage/prism/Program.cpp
```diff
@@ -707,6 +707,19 @@ storm::expressions::Expression Program::getInitialStatesExpression() const {
     }
 }
 
+bool Program::hasIntervalUpdates() const {
```
sjunges, line 710: Broader discussion: We currently do not cache such information, which may be a downside for some of the larger prism programs. Seems not urgent for me, but wanted to know your thoughts.

src/storm-pars-cli/storm-pars.cpp
```diff
@@ -400,14 +399,20 @@ void processInputWithValueTypeAndDdlib(cli::SymbolicInput& input, storm::cli::Mo
...
     }
 
     STORM_LOG_THROW(model, storm::exceptions::InvalidSettingsException, "No input model.");
     if (model) {
         model->printModelInformationToStream(std::cout);
     }
 
+    using ValueType = storm::RationalFunction;
+    auto const DdType = storm::dd::DdType::Sylvan;
+    STORM_LOG_THROW(model->supportsParameters(), storm::exceptions::UnexpectedException, "Expected a parametric model.");
```
sjunges, line 412: Why are these unexpected exceptions?

src/storm-cli-utilities/model-handling.h
```diff
@@ -606,19 +702,28 @@ std::pair<std::shared_ptr<storm::models::sparse::Model<ValueType>>, bool> prepro
     }
 
     if (mpi.applyBisimulation) {
-        result.first = preprocessSparseModelBisimulation(result.first, input, bisimulationSettings);
-        result.second = true;
+        if constexpr (storm::IsIntervalType<ValueType>) {
+            STORM_LOG_THROW(false, storm::exceptions::NotSupportedException, "Bisimulation not supported for interval models.");
```
sjunges, line 706: @linusheck did you look into this?

src/storm-cli-utilities/model-handling.h
```diff
@@ -1228,68 +1323,85 @@ void verifyWithSparseEngine(std::shared_ptr<storm::models::ModelBase> const& mod
...
-                                            "interested in building the model with state valuations using --buildstateval.");
-                        STORM_LOG_WARN_COND(
-                            sparseModel->hasChoiceLabeling() || sparseModel->hasChoiceOrigins(),
-                            "No symbolic choice information is available. The scheduler output will use internal choice ids. You might be interested in "
-                            "building the model with choice labels or choice origins using --buildchoicelab or --buildchoiceorig.");
-                        STORM_LOG_WARN_COND(sparseModel->hasChoiceLabeling() && !sparseModel->hasChoiceOrigins(),
-                                            "Only partial choice information is available. You might want to build the model with choice origins using "
-                                            "--buildchoicelab or --buildchoiceorig.");
+            if constexpr (storm::IsIntervalType<ValueType>) {
+                STORM_LOG_THROW(false, storm::exceptions::NotSupportedException, "Scheduler export for interval models is not supported.");
```
sjunges, line 1327: This relates to #741

sjunges:
I really like the changes in particular regarding the model handling :-)

src/storm/adapters/RationalNumberForward.h
tquatmann: We could have both: a `NumerRepresentation[Forward].h` that essentially just includes `RationalNumber[Forward].h` and `Interval[Forward].h`?

With exact intervals, `Interval.h` has to include `RationalNumber.h`, though.

src/storm/models/ModelBase.h
tquatmann: The problem is that I can only downcast a `ModelBase` to a `storm::models::symbolic::Model<DdType, ValueType>` once I know the `DdType`.

Alternatively, we could introduce a `getModelRepresentation()` using [this](https://github.com/moves-rwth/storm/blob/ca1a439fa00b927fff563ae16dff8e7146429eba/src/storm/models/ModelRepresentation.h#L9) enum. Would you prefer that?

src/storm/storage/prism/Program.cpp
tquatmann: We'd have to make sure that the cached information remains valid (or cleared) whenever the program is changed. 

Imo, there needs to be evidence of this being a practical performance issue before making the code more complex / unsafe by introducing a caching mechanism.

src/storm-pars-cli/storm-pars.cpp
tquatmann: Because I expect buildModel to produce a model that supports parameters and that has the right DD type (if relevant). If it is not the case, something unexpected happend. (This might as well be STORM_LOG_ASSERTS but since this is not performance critical, I don't see any harm in checking this in release mode, too).

tquatmann:
Thanks for looking through this! 

> Does this mean that one can always evaluate the lower bound and the upper bound independently of each other? I am not sure if that works well for all kinds of operations, say 1/x with x in [-1,1] is not [-1,1] ?

Yes, lower bound and upper bound are evaluated independently. As far as I understand the [PRISM specification](https://www.prismmodelchecker.org/manual/ThePRISMLanguage/UncertainModels) this should be no issue as there are no expressions of type "interval". Please correct me if I'm wrong, though :)

src/storm/generator/PrismNextStateGenerator.cpp
tquatmann: If this assertion triggers, it would really be a bug in the code logic as we already check whether there are intervals in the PRISM program during the construction (line 44-45):

```cpp
    STORM_LOG_THROW(storm::IsIntervalType<ValueType> || !this->program.hasIntervalUpdates(), storm::exceptions::WrongFormatException,
                    "The provided PRISM program contains interval updates, which are not supported by the chosen value type.");

```

In my view, this type of bug is not what exploration checks are supposed to handle.

src/storm/generator/PrismNextStateGenerator.cpp
tquatmann: I added checks for `0 <= probability <= 1` at places where `evaluateLikelihoodExpression` is called. This is also usefull for non-interval updates (although we also have checks that the sum of all probabilities is equal to 1).

However, I wonder if this actually needs to be enforced with intervals. 
One could argue that `[-1,2]` is a valid probability interval (and equivalent to `[0,1]`). Thoughts?

sjunges: I think the problem is if you have multiple parallel components, then, because [-2, 1] * [-2, 1] = [-2, 4], which is not what you would get. We don't get these problems if we ensure that we do not allow negative probabilities in our intervals.

src/storm/models/ModelBase.h
sjunges: We ran into a similar discussion with @volkm. I think it is fine for now, but we probably want to get rid of the DdType in the model. We should maybe discuss this in a meeting.

src/storm-pars-cli/storm-pars.cpp
sjunges: Ok, so unexpectedexceptions are basically programming bugs :-)

src/storm/storage/prism/Program.cpp
sjunges: I think programs are basically thought to be immutable for us, right? At least,most transformations create new objects.  

I think part of the discussion here is that the QVBS does not have large symbolic programs; but that is a choice of the benchmarks. 

Anyhow, not urgent.

src/storm/adapters/RationalNumberForward.h
sjunges: A question is whether we want DoubleIntervalForward and ExactIntervalForward or not. 
Either way, inlcuding NumberRepresentation.h sounds more understandable than including RationalNumberForward to do something with intervals :-)

tquatmann: Created a new issue #750

PR #745, opened 2025-07-31
Revised Spot configuration

volkm:
Explicitly disable debug configurations. I based the configuration arguments on the [debian build rules](https://gitlab.lre.epita.fr/spot/spot/-/blob/next/debian/rules) of Spot.

On my machine, this decreased the size of the Spot library.


sjunges:
One could argue that the debug symbols in spot are great if you are building storm in debug mode. However, that is not critical to me. 

Otherwise, thanks for making this work so fast.

volkm:
Thanks for the suggestions. We now distinguish between debug/release flags.

PR #754, opened 2025-08-13
Clean up and fix for BitVector

### LINKED ISSUE #753: Incorrect number of transitions in a one-state model
The following model generates an MDP with only one choice and one transition. Apparently, this only happens on Linux (gcc?) Systems.

```
mdp

module model
    s : [0..0] init 0;
    
    [a] s=0 -> 1 : (s'=0);
    [b] s=0 -> 1 : (s'=0);
endmodule

rewards
	[a] s=0 : 1;
	[b] s=0 : 100;
endrewards
```

```
Model type: 	MDP (sparse)
States: 	1
Transitions: 	1
Choices: 	1
Reward Models:  none
State Labels: 	2 labels
   * deadlock -> 1 item(s)
   * init -> 1 item(s)
Choice Labels: 	none
```

Note that the state also gets the label deadlock.

Changing the `s : [0..0]` to `s : [0..1]` apparently produces the right results. Replacing all four guards `s=0` with true also works.

Some edge case is probably triggered when a state is encoded with 0 bits. Likely when loading the state into the expression evaluator.

tquatmann:
- simplify code for getAsInt and getNumberOfSetBitsBeforeIndex (using c++20 std::popcount)
- correctly handle 0-sized bit vectors for getAsInt method (fixes #753)
- add a test case for 0-sized bit vectors
- replace all uint_fast64_t by uint64_t in Bitvector[Test]

### DIFF
--- a/src/storm/storage/BitVector.cpp
+++ b/src/storm/storage/BitVector.cpp
@@ -766,8 +766,10 @@ uint64_t BitVector::getNextSetIndex(uint64_t startingIndex) const {
 }
 
 uint64_t BitVector::getNextUnsetIndex(uint64_t startingIndex) const {
+#ifdef ASSERT_BITVECTOR
     STORM_LOG_ASSERT(getNextIndexWithValue<false>(buckets, startingIndex, bitCount) == (~(*this)).getNextSetIndex(startingIndex),
                      "The result is inconsistent with the next set index of the complement of this bitvector");
+#endif
     return getNextIndexWithValue<false>(buckets, startingIndex, bitCount);
 }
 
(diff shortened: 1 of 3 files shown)


sjunges:
LGTM! If you want more cleaning, think about the ifdef.. :-)

sjunges:
How about leaving the code in for the moment (but having ASSERT_BITVECTOR 0 ). Such code can be helpful in debugging issues with bitvectors?

tquatmann:
Yes, that makes sense. I re-introduced the two `#ifdef`s around `STORM_LOG_ASSERT`, because now they make sense again.

PR #756, opened 2025-08-20
add two strings into stormoptions.cmake

sjunges:
This helps for checking storm-versions when building python extension beyond stormpy, see also  https://github.com/moves-rwth/stormpy/issues/260


volkm (approved):
Thanks. LGTM

PR #757, opened 2025-08-20
Treat warnings as errors in CI

volkm:
Added CMake option `STORM_WARNING_AS_ERROR` and enabled it in the CI to treat all warnings as errors.

Currently, the flag is used for all configurations in the CI which leads to a number of warnings which should be fixed.


sjunges:
This will be very helpful. I was wondering; just like we disabled quite some warnings on mac/clang right now, maybe we should disable the massive amount of unused-parameter warnings for now, get this PR merged, and then offer someone a box of chocolates to handle the unused parameters?

volkm:
I disabled the unused parameter warnings when using `STORM_WARNING_AS_ERROR`. The CI moves a bit further now and I can take a look at some of the other warnings in due time.

How can I handle [this warning](https://github.com/moves-rwth/storm/blob/master/resources/3rdparty/sylvan/src/storm_wrapper.cpp#L21)? It is a custom warning so I am not sure how to ignore it. I tried using the CMake flag `STORM_WARNING_AS_ERROR` but I would also need to export it into the config which seems a bit overhead for one warning.

sjunges:
How about CMake's add_compile_definitions( ... )

volkm:
Needs #765

src/storm/solver/IterativeMinMaxLinearEquationSolver.cpp
```diff
@@ -221,8 +221,6 @@ bool IterativeMinMaxLinearEquationSolver<ValueType, SolutionType>::solveInducedE
         STORM_LOG_ASSERT(subB.size() == x.size(), "Sizes of subB and x do not coincide.");
         STORM_LOG_ASSERT(this->linearEquationSolverFactory != nullptr, "Wrong constructor was called.");
 
-        bool convertToEquationSystem = this->linearEquationSolverFactory->getEquationProblemFormat(env) == LinearEquationSolverProblemFormat::EquationSystem;
```
tquatmann: This seems relevant as it controls whether we have to build an equation system of the form `Ax=b` (e.g. gmm, eigen) or of the form `Ax+b=x` (e.g. topological solver, native).

As far as I can see, the code below always builds `Ax+b=x`. I suggest to change this to

```
STORM_LOG_THROW(this->linearEquationSolverFactory->getEquationProblemFormat(env) == LinearEquationSolverProblemFormat::FixedPointSystem, storm::exceptions::NotImplementedException, "Solving induced system of Interval Model not supported for the selected equation solver");
```

And create an issue for that?

volkm: Thanks for the suggestion. I added the check.

volkm:
Most of the warnings should be fixed now. There are still at least three types of warnings open, but I am currently not knowledgeable enough to fix them. If you know how to address them, feel free to let me know.

volkm:
All warnings should be fixed now. Thanks to @sjunges for some pointers on how to fix them.

There are even more warnings if we enable `STORM_ALLWARNINGS` but this is a separate task for another time.

.github/workflows/buildtest.yml
```diff
@@ -140,7 +141,8 @@ jobs:
           buildType: ${{ matrix.config.buildType }}
           docker: true
       - name: Run tests
-        run: docker exec ci bash -c "cd /opt/storm/build; ASAN_OPTIONS=detect_leaks=0,detect_odr_violation=0 ctest test --output-on-failure"
+        #run: docker exec ci bash -c "cd /opt/storm/build; ASAN_OPTIONS=detect_leaks=0,detect_odr_violation=0 ctest test --output-on-failure"
+        run: docker exec ci bash -c "cd /opt/storm/build; ctest test --output-on-failure"
```
tquatmann, line 145: Any reason to keep the commented line in? If yes, maybe add a quick comment?

tquatmann (approved):
LGTM! Many Thanks!

.github/workflows/buildtest.yml
```diff
@@ -236,8 +239,9 @@ jobs:
     name: Linux Tests (${{ matrix.distro }}, ${{ matrix.buildType }})
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
-        distro: ["debian-12", "ubuntu-24.04"]
+        distro: ["debian-12", "ubuntu-24.04", "debian-13"]
```
sjunges, line 252: What about debian-14?

resources/3rdparty/CMakeLists.txt
```diff
@@ -87,6 +87,10 @@ ExternalProject_Add(
         LOG_DOWNLOAD ON
         LOG_INSTALL ON
 )
+if(STORM_WARNING_AS_ERROR AND STORM_COMPILER_GCC)
```
sjunges, line 105: Is this compile option now used globally or only on targets in resources or only on the eigen target?

resources/3rdparty/CMakeLists.txt
```diff
@@ -535,6 +539,10 @@ FETCHCONTENT_DECLARE(
 SET(SYLVAN_GMP ON)
 SET(USE_CARL ON)
 set(SYLVAN_PORTABLE ${STORM_PORTABLE})
+if(STORM_WARNING_AS_ERROR)
```
sjunges: Same question: to what does this option now apply?

.github/workflows/buildtest.yml
volkm: Debian 14 is the testing version and will not be released for another 2 years. I think the Archlinux configurations should take care of the most recent compiler versions.

resources/3rdparty/CMakeLists.txt
volkm: I believe it is used globally. As far as I remember I could not simply disable it only for the Eigen includes. But I should maybe take another look into it.

volkm:
We opted to first merge the fixes to warnings in the source code, see #792. Afterwards, we can address the remaining CMake questions here.

resources/3rdparty/CMakeLists.txt
sjunges: But then I would really not put it here. Lets only change compile options for storm in the src folder?

resources/3rdparty/CMakeLists.txt
sjunges: Maybe use https://cmake.org/cmake/help/latest/command/target_compile_options.html if you want to change this for sylvan.

volkm: Good tip, thanks. I will try this.

resources/3rdparty/CMakeLists.txt
volkm: Setting `target_compile_options` seems to work for Clang but not for GCC

PR #765, opened 2025-08-22
Updated l3pp version

volkm:
The latest l3pp version fixed some compile warnings.

I also removed the shipped l3pp sources because they are downloaded from Github. If we prefer shipping the files, I can revert it.

### DIFF
--- a/src/storm/utility/logging.h
+++ b/src/storm/utility/logging.h
@@ -1,14 +1,10 @@
 #pragma once
 
-// Include config to know whether carl is available or not.
-#include "storm-config.h"
-#ifdef STORM_HAVE_CARL
 // Load streaming operator from CARL
 #include <carl/io/streamingOperators.h>
 namespace l3pp {
 using carl::operator<<;
 }
-#endif
 
 #include <l3pp.h>
 


src/storm/utility/initialize.cpp
```diff
@@ -66,7 +66,7 @@ void setLogLevel(l3pp::LogLevel level) {
 }
 
 void initializeFileLogging(std::string const& logfileName) {
-    l3pp::SinkPtr sink = l3pp::FileSink::create(logfileName);
+    l3pp::SinkPtr sink = l3pp::StreamSink::create(logfileName);
```
sjunges, line 69: This I didnt get? l3pp still has file sinks, dont they?

src/storm/utility/logging.h
```diff
@@ -1,7 +1,6 @@
-#ifndef STORM_UTILITY_LOGGING_H_
-#define STORM_UTILITY_LOGGING_H_
+#pragma once
 
-// Include config to know whether CARL is available or not.
+// Include config to know whether carl is available or not.
```
sjunges: I think we are not really checking this anymore anyways.

src/storm/utility/initialize.cpp
volkm: Yes, they [merged FileSink and StreamSink](https://github.com/hbruintjes/l3pp/commit/7909816adab86904992f37efdd3eea1ee0a85beb)

src/storm/utility/logging.h
volkm: True, I removed the check.

PR #768, opened 2025-08-25
Fixed ConstantComparator

volkm:
The ConstantComparator had two issues in my view.

Comparison with equal should take the maximum instead of the sum for relative precision. This also follows how python computes [isclose](https://docs.python.org/3/library/math.html#math.isclose).
Currently, the check 0.4 == 0.6 with relative precision 0.2 would yield true (0.2 <= 0.2*(0.4+0.6)) but should be false instead (0.2 <= 0.2* max(0.4,0.6)).

The less than comparison is also not right in my view.
Currently, the check 0 < 0.1 with precision 0.2 is true as expected. But the check 0.9 < 1 is false because of 0.9 < 1 - 0.2.
The proposed fix does lead to some errors though, because e.g. in the [SparseMatrix](https://github.com/moves-rwth/storm/blob/master/src/storm/storage/SparseMatrix.cpp#L2272) a check `isLess(0.1, 0)` now yields true.

Should we also adapt the check in the SparseMatrix or define IsLess in a different way?

### DIFF
--- a/src/storm/utility/ConstantsComparator.cpp
+++ b/src/storm/utility/ConstantsComparator.cpp
@@ -70,7 +70,7 @@ bool ConstantsComparator<ValueType, ConstantsComparatorEnablePrecision<ValueType
     } else {
         ValueType absDiff = storm::utility::abs<ValueType>(value1 - value2);
         if (relative) {
-            return absDiff <= precision * (storm::utility::abs(value1) + storm::utility::abs(value2));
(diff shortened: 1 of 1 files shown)


sjunges:
I am working on a larger overhaul of the comparator. I will include the changes you suggested here in that PR (to be uploaded soon).

volkm:
Sounds good. Feel free to close this PR then.

Issue #718, opened 2025-05-23
Update gmm


volkm:
The failures in #707 might be fixed when updating gmm to a newer version.

I tried updating to gmm-5.4.4, but encountered the same issue as [here](https://getfem.discourse.group/t/compilation-error-when-using-gmm-mumps-solve-is-not-a-member-of-gmm/84). Apparently, now either SuperLU or MUMPS are needed. Maybe it is possible to change it such that gmm still works without either of the libraries.

I also tried using gmm-5.4.2 where the issue is not yet present, but this version requires at least some adaptions in `GmmxxMultiplier`, for example `mult_add_by_row_bwd` was removed.

When updating gmm we should also revise its handling in CMake:
- download gmm instead of shipping it
- install it as described [here](https://getfem.org/gmm/install.html) which only exports the include files and not the rest.

sjunges:
Did we report the error to the GMM++ discussion forum?

volkm:
[Yes](https://savannah.nongnu.org/support/index.php?111246)

PR #771, opened 2025-08-28
Update gmm

tquatmann:
We no longer ship the sources of gmm and can use the latest version.

I managed to workaround the issue in #718 (about missing dependencies) by only including those files we actually need.


resources/3rdparty/CMakeLists.txt
```diff
@@ -45,14 +45,26 @@ list(APPEND STORM_DEP_TARGETS l3pp)
...
     add_library(gmm INTERFACE) # Not imported, we are in control of the sources.
+    add_dependencies(gmm gmm_src)
     target_include_directories(gmm INTERFACE
-            $<BUILD_INTERFACE:${STORM_3RDPARTY_SOURCE_DIR}/gmm-5.2/include>
+            $<BUILD_INTERFACE:${STORM_3RDPARTY_BINARY_DIR}/gmm/include>
             $<INSTALL_INTERFACE:${STORM_RESOURCE_INCLUDE_INSTALL_DIR}/gmm>
     )
     install(TARGETS gmm EXPORT storm_Targets)
-    install(DIRECTORY ${STORM_3RDPARTY_SOURCE_DIR}/gmm-5.2/include/
+    install(DIRECTORY ${STORM_3RDPARTY_BINARY_DIR}/gmm/include/
```
tquatmann, line 68: I have not implemented the "proper" installation (via `./configure --prefix ...` and `make install` as suggested in #718 . I guess I would have to add those commands to the `ExternalProject_Add` above and remove the "manual" installation here?

src/storm/adapters/gmm.h
```diff
@@ -14,8 +14,16 @@
 #pragma GCC diagnostic ignored "-Wmisleading-indentation"
 #endif
 
-#include "gmm/gmm_iter_solvers.h"
```
tquatmann, line 17: Before merging, I'd like to check if we still have to disable those warings.

resources/3rdparty/CMakeLists.txt
```diff
@@ -45,14 +45,26 @@ list(APPEND STORM_DEP_TARGETS l3pp)
 #############################################################
 
 if(NOT STORM_DISABLE_GMM)
-    message (STATUS "Storm - Including gmm 5.2.")
+    message (STATUS "Storm - Including gmm 5.4.4.")
```
volkm: Maybe set a variable GMM_VERSION for easy bumping of the version number?

volkm (approved):
LGTM. Great to see that we do not need the custom changes anymore.

Could you maybe add one line to update_resources.md?

resources/3rdparty/CMakeLists.txt
sjunges: So *IF* you would use make install, you would need to indeed install it into the binary dir and then still install it manually into whatever install directory we use during installation. (Because at build time you cannot know install directories for sure) etc. 

So, this coulde here is definitively necessary.

sjunges:
Wait wait, this PR should also fix this: https://github.com/moves-rwth/storm/blob/0641d863330902f6141850556e6793d8aa0343d8/.github/workflows/buildtest.yml#L181 (and some more occurences, right?)

volkm:
Yes, we should try to enable gmm again for the compiler tests. I am not sure how to suggest changes for these files, but lines 181, 188 and 203 should be removed.

.github/workflows/buildtest.yml
```diff
@@ -178,14 +178,14 @@ jobs:
         config:
           - {name: "GCC",
              buildType: "Debug",
-             disable_gmm: "ON", # TOOD: enable again if GMM was fixed
+             disable_gmm: "OFF",
```
volkm: Can you remove these lines and line 203, because we do not need them anymore?

tquatmann: I'm reverting this change as GMM is apparently not fixed yet :)

src/storm/adapters/gmm.h
tquatmann: GCC did not produce any warnings, so we now only disable `-Wunused-parameter` for clang

volkm: Thanks for checking

resources/3rdparty/CMakeLists.txt
tquatmann: Ahh right, makes sense! . 
It seems to me, that `configure` and `make install` would just add an additional copy of the header files (plus whatever `./configure` does, like checking for a C compiler, etc).
Overall, it seems to be simpler to keep the code as it is?

PR #776, opened 2025-09-01
Fix a number of warnings

tquatmann:
This PR fixes a number of warnings that we recently introduced.


tquatmann:
Argh! Sorry, I should have known this.
I made sure that the changes here coincide with #757 so there should not be any conflicts (I hope).

volkm (approved):
LGTM.

PR #778, opened 2025-09-02
Add clang 14 version check to cmakelist because storm does not compile for clang 14.0.6.

lukovdm:


volkm:
I think you can just increase the compiler version in line 231

CMakeLists.txt
```diff
@@ -213,23 +213,17 @@ message(VERBOSE "Storm - Assuming extension for static libraries: ${STATIC_EXT}"
...
+	if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 15.0)
+		message(FATAL_ERROR "Clang version must be at least 15.0, currently you are using version ${CMAKE_CXX_COMPILER_VERSION}.")
 	endif()
 
 	set(STORM_COMPILER_CLANG ON)
 	set(CLANG ON)
 	set(STORM_COMPILER_ID "clang")
 elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang")
 	# using AppleClang
-	if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 14.0.3)
```
volkm: Sorry, I think the checks for AppleClang should stay the same. Only "normal" Clang should be affected.

PR #781, opened 2025-09-03
Storm version 1.11.0

volkm:


CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ endif()
 # Set project name
 project (storm
 		LANGUAGES CXX C
-		VERSION 1.10.0.1)
+		VERSION 1.11.0)
```
sjunges, line 9: In the future, maybe we can add to the CI workflow that a commit that changes this back to x.y.z.1.

sjunges:
LGTM!

CMakeLists.txt
volkm: You mean immediately after a release we increase it by 1 indicate the next development version? That would indeed be a good idea.

volkm: Added by https://github.com/moves-rwth/storm/pull/782

tquatmann (approved):
LGTM!

PR #785, opened 2025-09-16
Fix conditional bisection method bugs

lukovdm:
Conditional model checking had two bugs in the bisection method:
- A log10 was used to calculate the amount of digits of a rational number. But log10 on gmp rationals converts to double internally, and thus has precision problems.
- When doing non-exact bisection, the precision of the upper and lower bound could be reached. This resulted in the `upperbound < lowerbound`. When this happens, it is now "fixed" and a warning is given to the user.

I also added a warning when using log10 on gmp rationals, that it is not exact and should be avoided.

### DIFF
--- a/src/storm/utility/constants.cpp
+++ b/src/storm/utility/constants.cpp
@@ -1,10 +1,8 @@
 #include "storm/utility/constants.h"
 
 #include <cmath>
-#include <cstdint>
 #include <type_traits>
 
-#include "storm/adapters/RationalNumberAdapter.h"
 #include "storm/storage/sparse/StateType.h"
 
 #include "storm/exceptions/InvalidArgumentException.h"
--- a/src/storm/utility/constants.h
+++ b/src/storm/utility/constants.h
@@ -1,8 +1,6 @@
 #ifndef STORM_UTILITY_CONSTANTS_H_
 #define STORM_UTILITY_CONSTANTS_H_
 
-#include <cln/types.h>
-#include <sys/types.h>
 #ifdef max
 #undef max
 #endif


lukovdm:
Let me know if the non-exact solution is wanted, as I am not entirely sure.

volkm:
Ok, thanks for the clarification. I just misinterpreted the `rationalize` function then.

src/storm/modelchecker/helper/conditional/ConditionalHelper.cpp
```diff
@@ -612,8 +621,15 @@ SolutionType computeViaBisection(Environment const& env, BisectionMethodBounds b
         if constexpr (storm::NumberTraits<SolutionType>::IsExact) {
             // find a rational number with a concise representation close to middle and within the bounds
             auto const exactMiddle = middle;
-            auto numDigits = storm::utility::convertNumber<uint64_t>(
-                storm::utility::floor(storm::utility::log10<SolutionType>(storm::utility::one<SolutionType>() / (*upperBound - *lowerBound))));
+
+            // Find number of digits - 1. Method using log10 does not work since that uses doubles internally.
+            auto numDigits = -1;
+            SolutionType remaining = storm::utility::one<SolutionType>() / (*upperBound - *lowerBound);
```
sjunges: Maybe put this into an aux function (in an anonymous namespace) called magnitude? Or even add magnitude to the constants.cpp?

src/storm/modelchecker/helper/conditional/ConditionalHelper.cpp
```diff
@@ -612,8 +621,15 @@ SolutionType computeViaBisection(Environment const& env, BisectionMethodBounds b
...
+            while (remaining >= storm::utility::one<SolutionType>()) {
+                ++numDigits;
+                remaining = storm::utility::floor<SolutionType>(remaining / storm::utility::convertNumber<SolutionType>(10));
```
sjunges: You can probably avoid doing the convert number (and creating new `one`s?) inside the loop?

PR #787, opened 2025-09-19
musl support

sjunges:
This PR aims to support linux systems that use musl rather than libc. The modifications mostly clean up the code.


src/test/CMakeLists.txt
```diff
@@ -1,8 +1,6 @@
-if (PROJECT_IS_TOP_LEVEL)
```
sjunges, line 1: Notice that I just moved this one level up to have less different places where we check for inclusions.

.github/workflows/Dockerfile.alpine
```diff
@@ -0,0 +1,98 @@
...
+#########################
+# The arguments can be set from the commandline with:
+# --build-arg <arg_name>=<value>
+
+# Specify number of threads to use for parallel compilation
+ARG no_threads=4
+# CMake build type
+ARG build_type=Release
+# Carl tag to use
+ARG carl_tag="02b75713159c881e4c0d0a3e41a48f15b90b0aa2"
```
sjunges, line 24: This can be updated after the next carl release

volkm: Can you add the Dockerfile.alpine here? https://github.com/moves-rwth/storm/blob/master/.github/workflows/update_carl.yml#L32 Then it gets automatically updated with the next carl release ;)

.github/workflows/Dockerfile.alpine
```diff
@@ -0,0 +1,98 @@
+# Dockerfile for Storm on Archlinux
```
volkm:
```suggestion
# Dockerfile for Storm on Alpine
```

.github/workflows/Dockerfile.alpine
```diff
@@ -0,0 +1,98 @@
...
+ARG disable_z3="OFF"
+ARG developer="OFF"
+ARG cln_exact="OFF"
+ARG cln_ratfunc="ON"
+ARG all_sanitizers="OFF"
+
+# Specify additional CMake arguments for Storm
+ARG cmake_args=""
+
+# Additional Archlinux packages
```
volkm:
```suggestion
# Additional Alpine packages
```

src/CMakeLists.txt
```diff
@@ -140,17 +140,23 @@ add_subdirectory(storm-pomdp)
 add_subdirectory(storm-cli-utilities)
 
 if(STORM_BUILD_EXECUTABLES)
+    message(STATUS "Storm - configuring executables..." )
```
volkm:
```suggestion
    message(STATUS "Storm - Configuring executables..." )
```

src/CMakeLists.txt
```diff
@@ -140,17 +140,23 @@ add_subdirectory(storm-pomdp)
...
     # Typically an idea to switch this off, if you want to build a library on top of Storm.
     add_subdirectory(storm-cli)
     add_subdirectory(storm-conv-cli)
     add_subdirectory(storm-dft-cli)
     add_subdirectory(storm-gspn-cli)
     add_subdirectory(storm-pars-cli)
     add_subdirectory(storm-pomdp-cli)
+else()
+    message(STATUS "Storm - not configuring executables." )
```
volkm:
```suggestion
    message(STATUS "Storm - Disabled building executables." )
```

src/CMakeLists.txt
volkm: Do we need this as output?

src/CMakeLists.txt
```diff
@@ -140,17 +140,23 @@ add_subdirectory(storm-pomdp)
...
 endif()
 
-if (STORM_BUILD_TESTS)
+if (PROJECT_IS_TOP_LEVEL AND STORM_BUILD_TESTS)
+    message(STATUS "Storm - configuring tests..." )
     add_subdirectory(test)
+else()
+    message(STATUS "Storm - not configuring tests." )
```
volkm:
```suggestion
    message(STATUS "Storm - Disabled building tests." )
```

src/storm/settings/modules/SylvanSettings.cpp
```diff
@@ -50,15 +50,15 @@ uint_fast64_t SylvanSettings::getNumberOfThreads() const {
     // Automatic detection
 #ifdef APPLE_SILICON
     // Prevents issues with multi-threaded execution on Apple Silicon
-    return 1u;
+    return 1ul;
 #else
-    return std::max(1u, storm::utility::getNumberOfThreads());
+    return std::max(UINT64_C(1), storm::utility::getNumberOfThreads());
```
volkm, line 55: I guess it does not work when simply using `1ul`?

sjunges: No, uint64_t is unsigned long on some systems and unsigned long long on others...  there is likely a question whether we should have gone for uint64_t everywhere, but I would stick with it for now.

src/CMakeLists.txt
sjunges: Likely not. I can remove it.

sjunges: Although for the sake of consistency, we mostly have positive and negative outputs, so if we say we dont do x then we typically also have output that says we do x, even if that is the default.

volkm: Sounds good. Then we can keep the output. I would maybe phrase it as "Enabled building executable", because I found the configuring part a bit confusing.

src/storm/settings/modules/SylvanSettings.cpp
volkm: Ok, thanks for the clarification.

PR #790, opened 2025-09-30
Better support for ARM

### LINKED ISSUE #788: Support for Linux ARM
Linux ARM has the same Sylvan issue as macOS Silicon (see [here](https://github.com/moves-rwth/stormpy/pull/283)) and we should only use 1 core as well. We also should add CI tests for Linux ARM.

volkm:
- added CI tests on ARM
- extended Sylvan workaround from Apple Silicon to all ARM architectures. Fixes #788 
- consistently use preprocessor macro `MACOS`

### DIFF
--- a/.github/workflows/buildtest.yml
+++ b/.github/workflows/buildtest.yml
@@ -249,7 +249,7 @@ jobs:
       matrix:
         distro: ["debian-12", "ubuntu-24.04"]
         buildType: ["Release"]
-        runner: [ubuntu-latest, ubuntu-24-04-arm]
+        runner: [ubuntu-latest, ubuntu-24.04-arm]
     steps:
       - name: Git clone
         uses: actions/checkout@v5


.github/workflows/buildtest.yml
```diff
@@ -243,12 +243,17 @@ jobs:
...
-    name: Linux Tests (${{ matrix.distro }}, ${{ matrix.buildType }})
-    runs-on: ubuntu-latest
+    name: Linux Tests (${{ matrix.distro }}, ${{ matrix.buildType }}) on ${{ matrix.runner }}
+    runs-on: ${{ matrix.runner }}
     strategy:
       matrix:
         distro: ["debian-12", "ubuntu-24.04"]
         buildType: ["Release"]
+        runner: [ubuntu-latest]
+        include:
```
sjunges: Dont you want arm and not-arm?

sjunges:
One question about the runner config from a github actions noob, otherwise, great, thanks!

.github/workflows/buildtest.yml
volkm: To limit the number of CI jobs, I only added one ARM run which I explicitly `include`. This was copied from carl where we have more configurations.
For Storm, we would only add 2 additional jobs on ARM which I think is reasonable. I will change it.

sjunges: hmm, I just didnt understand the include notation correctly :-)

volkm: Yeah, it is a bit [complex](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#expanding-or-adding-matrix-configurations). Your suggestion is clearer though and a good idea.

PR #796, opened 2025-10-02
Migrate to other Mac runners

sjunges:
Mac-13 is deprecated soon, migrating to mac-15-intel. Also update the xcode version to latest stable on the new version. I suggest to keep the old version for the older mac config to support a range of versions.


volkm:
LGTM. Thanks for the quick reaction to the deprecation warning.

sjunges:
I was afraid i would forget.

Issue #807, opened 2025-11-12
Add support for explicit scheduler-nature optimization directions in uncertain (interval) models


plindnercs:
For checking uncertain (interval) models, the [PRISM Property Specification](https://www.prismmodelchecker.org/manual/PropertySpecification/UncertainModels) allows explicitly specifying the optimization direction for both the scheduler and nature. In contrast, in STORM's current implementation for IMDPs, nature’s optimization direction is implicitly chosen as the adversarial one with respect to the scheduler’s direction. e.g., when checking `Pmax=? [F "done"]`, nature acts in a worst-case fashion and minimizes the probability, resulting in `Pmaxmin=? [F "done"]`.
At the moment, it is not possible in STORM to check properties where both optimization directions are explicity given, e.g., `Pmaxmax=? [F "done"]`.

In my opinion, it would be nice if STORM supports properties like `Pmaxmax=? [F "done"]`, as this is also possible in PRISM. Additionally, while working on my pull request adding support for IDTMCs, I noticed that checking `Pmax=? [F "done"]` yields the same result in STORM as checking `Pmin=? [F "done"]` in PRISM. This is due to the fact, that the optimization direction is interpreted as the one for the scheduler and not nature, and hence the robust value iteration minimizes the probability by a worst-case behaviour of nature.

In PRISM, if only one optimization direction is given for an IMDP, this optimization direction is chosen for both the scheduler and nature, e.g., `Pmax=?` is interpreted as `Pmaxmax=?`. However, when checking `Pmaxmax=?` on an IDTMC, PRISM currently returns an error.

volkm:
Is this an issue with the parsing of the property? As far as I can see from [this test](https://github.com/moves-rwth/storm/blob/master/src/test/storm/modelchecker/prctl/mdp/RobustMdpPrctlModelCheckerTest.cpp#L56) Storm in principle already has support for specifying both optimization directions.

tquatmann:
I edited to cooperative as it fits better, indeed.

I'm absolutely in favour of setting the interval resolution more explicitly. 
Your suggestion also means that we don't have to tinker with the `OperatorFormula`s and the `FormulaParser` which is a nice side effect :)

I would then be in favour of saying (in CLI)  `--interval-resolution <val>` must be present iff the input is an interval model. API users also have to set that explicitly (e.g. in the checktask). 

Only downsides are that

- PRISM users can't use their properties out-of-the-box.
[...]

PR #409, opened 2023-08-25
Support for interval-based transitions in storm

sjunges:
- Support interval-based models
- Support import/export to DRN
- Added Robust VI for MinMaxLinearSolver 
- Various changes that allow doubles as result whereas the model consists of intervals.

 In particular, it requires 
- [x] models can have interval-valued transitions 
- [x] models can be parsed and exported to DRN  
- [x] an update of carl, (done, see https://github.com/moves-rwth/carl-storm/pull/14)
- [x] control robustness via checktask
- [x] check scheduler extraction
- [x] checks to ensure that intervals are non-zero
- [x] implementation of canHandle
- [x] merging https://github.com/moves-rwth/carl-storm/pull/21
- [x] some convenience functions to add uncertainty
- [x] finished tests, 

Things that I will currently not support. I track these in this issue: https://github.com/moves-rwth/storm/issues/469
- Explicit support for solving on DTMCs -- the DTMCs must be declared as MDPs
- Support for rewards 
- Bounded reachability properties
- Topological VI
- Convenience functions to remove uncertainty (e.g., by taking the 'center').
- support for robustness in formulas
- Generating interval models from PRISM/JANI

Other things that I do not support and that I do not see happening right now
- DD-based representations
- Modified Policy iteration / LP-based methods
- Exact intervals

Issue #264, opened 2022-07-29
Default selected method not supported (dd's)


sjunges:
When using storm with dd and default settings, I get a warning that the selected method is not supported, which is confusing.

```
Storm 1.6.5 (dev)

Command line arguments: --prism examples/grid.prism -const N=800 --prop 'Tmin=? [F x=3 & y > 4]' -e dd

Model checking property "1": T[exp]min=? [F ((x = 3) & (y > 4))] ...
 WARN (SymbolicMinMaxLinearEquationSolver.cpp:64): Selected method is not supported for this solver, switching to value iteration.
```

tquatmann:
I see where this is coming from: the default solver is `topological` which is not available for `dd`.

One option would be to change the command line interface so that `topological` is no longer seen as a *type* of solver, but an option for solving that can be enabled/disabled (independent of the solver type). IMO this would be  a bit cleaner and also solves this issue.

sjunges:
Agree! So the option --topological should have an auto, disabled and enabled option, with auto being default and auto translating to enabled on sparse and disabled on dd?

PR #808, opened 2025-11-12
Implemented support for verifying IDTMCs

plindnercs:
This feature extension includes IDTMC support to verify the following properties:
- Reachability Probabilities
- Reach-Reward
- Propositional

Note:
- As discussed in issue #807, the results for `Pmin`, `Pmax`, `Rmin` and `Rmax` are flipped compared to PRISM.
- The expected results in the unit tests were double-checked with PRISM and manually verified by computing them by hand (except for the BRP-32-2 test case).
- If the `minmax:method` is not manually set to `vi`, the current implementation prints a warning and defaults to robust value iteration. This is somewhat related to issue #264. Note that the warning currently appears twice as the `minmax:method` is checked in two places.
- The implementation heavily relies on the IMDP implementation (see #409)
- This pull request should solve the first item and second item, at least for IDTMCs, in #762

I'm happy to receive any kind of feedback!

### DIFF
--- a/src/storm-cli-utilities/model-handling.h
+++ b/src/storm-cli-utilities/model-handling.h
@@ -1296,17 +1296,17 @@ void verifyModel(std::shared_ptr<storm::models::sparse::Model<ValueType>> const&
     auto verificationCallback = [&sparseModel, &ioSettings, &mpi](std::shared_ptr<storm::logic::Formula const> const& formula,
                                                                   std::shared_ptr<storm::logic::Formula const> const& states) {
         auto createTask = [&ioSettings](auto const& f, bool onlyInitialStates) {
-            (void)ioSettings;  // suppress unused lambda capture warning. [[maybe_unused]] doesn't work for lambda captures.
             if constexpr (storm::IsIntervalType<ValueType>) {
                 STORM_LOG_THROW(ioSettings.isUncertaintyResolutionModeSet(), storm::exceptions::InvalidSettingsException,
                                 "Uncertainty resolution mode required for uncertain (interval) models.");
                 return storm::api::createTask<ValueType>(f, storm::solver::convert(ioSettings.getUncertaintyResolutionMode()), onlyInitialStates);
             } else {
+                (void)ioSettings;  // suppress unused lambda capture warning. [[maybe_unused]] doesn't work for lambda captures.
                 return storm::api::createTask<ValueType>(f, onlyInitialStates);
             }
         };
-        bool filterForInitialStates = states->isInitialFormula();
-        auto task = createTask(formula, states->isInitialFormula());
+        bool const filterForInitialStates = states->isInitialFormula();
+        auto task = createTask(formula, filterForInitialStates);
         if (ioSettings.isExportSchedulerSet()) {
             task.setProduceSchedulers(true);
         }


src/storm/modelchecker/prctl/helper/SparseDtmcPrctlHelper.cpp
```diff
@@ -53,103 +54,130 @@ std::map<storm::storage::sparse::state_type, storm::RationalFunction> SparseDtmc
...
+    // Set up the solver.
+    storm::solver::GeneralMinMaxLinearEquationSolverFactory<ValueType, SolutionType> minMaxLinearEquationSolverFactory;
+    std::unique_ptr<storm::solver::MinMaxLinearEquationSolver<ValueType, SolutionType>> solver =
+        storm::solver::configureMinMaxLinearEquationSolver(env, std::move(goal), minMaxLinearEquationSolverFactory, std::move(submatrix));
+    solver->setRequirementsChecked();
+    solver->setUncertaintyIsRobust(goal.isRobust());
+    solver->setHasUniqueSolution(false);
+    solver->setHasNoEndComponents(false);
+    solver->setLowerBound(storm::utility::zero<SolutionType>());
+    solver->setUpperBound(storm::utility::one<SolutionType>());
```
tquatmann: The upper bound of 1 does not apply if we compute expected rewareds

src/storm/modelchecker/prctl/SparseDtmcPrctlModelChecker.cpp
```diff
@@ -74,130 +84,166 @@ bool SparseDtmcPrctlModelChecker<SparseDtmcModelType>::canHandle(CheckTask<storm
...
+    }
 }
 
 template<typename SparseDtmcModelType>
 std::unique_ptr<CheckResult> SparseDtmcPrctlModelChecker<SparseDtmcModelType>::computeUntilProbabilities(
-    Environment const& env, CheckTask<storm::logic::UntilFormula, ValueType> const& checkTask) {
+    Environment const& env, CheckTask<storm::logic::UntilFormula, SolutionType> const& checkTask) {
     storm::logic::UntilFormula const& pathFormula = checkTask.getFormula();
+    if (storm::IsIntervalType<ValueType>) {
+        STORM_LOG_THROW(checkTask.isOptimizationDirectionSet(), storm::exceptions::InvalidPropertyException,
```
tquatmann: It's a bit odd that the optimization direction controls nature for IDTMCs (but not for IMDPs, it seems).
Suggestion: the interval optimization direction should be stored as an `std::optional<storm::OptimizationDirection>` in the `Checktask` (and then checked e.g. here). That forces an API user to set it explicitly. 
When creating a checktask from a formula, the interval optimization direction can be derived automatically (following PRISM behaviour).

src/storm/modelchecker/prctl/SparseDtmcPrctlModelChecker.cpp
```diff
@@ -209,218 +255,278 @@ std::unique_ptr<CheckResult> SparseDtmcPrctlModelChecker<storm::models::sparse::
...
+    }
 }
 
 template<typename SparseDtmcModelType>
 std::unique_ptr<CheckResult> SparseDtmcPrctlModelChecker<SparseDtmcModelType>::computeReachabilityRewards(
-    Environment const& env, CheckTask<storm::logic::EventuallyFormula, ValueType> const& checkTask) {
+    Environment const& env, CheckTask<storm::logic::EventuallyFormula, SolutionType> const& checkTask) {
     storm::logic::EventuallyFormula const& eventuallyFormula = checkTask.getFormula();
+    if (storm::IsIntervalType<ValueType>) {
+        STORM_LOG_THROW(checkTask.isOptimizationDirectionSet(), storm::exceptions::InvalidPropertyException,
```
tquatmann: See above

src/storm/modelchecker/prctl/SparseDtmcPrctlModelChecker.cpp
tquatmann: Of course, this also means that we have to read the IntervalOptimizationDirection from the SolveGoal (somewhere in the Helper).

tquatmann (commented):
Good progress! My comment regarding the interval optimization direction is related to https://github.com/moves-rwth/storm/issues/807 .

src/storm-pars/modelchecker/region/SparseDtmcParameterLiftingModelChecker.cpp
```diff
@@ -440,7 +440,7 @@ std::vector<ConstantType> SparseDtmcParameterLiftingModelChecker<SparseModelType
         solver->setHasUniqueSolution();
         solver->setHasNoEndComponents();
         // Uncertainty is not robust (=adversarial)
-        solver->setUncertaintyIsRobust(false);
+        solver->setUncertaintyResolutionMode(UncertaintyResolutionMode::Cooperative);
```
plindnercs, line 441: In case the `uncertaintyIsRobust`-flag has been set explicitly, I used the corresponding enum value.

src/test/storm/modelchecker/prctl/mdp/RobustMdpPrctlModelCheckerTest.cpp
```diff
@@ -62,25 +62,76 @@ void checkModel(std::string const& path, std::string const& formulaString, doubl
...
+    auto resultMaxMax = checker.check(env, taskMaxMax);
+    EXPECT_NEAR(maxmax, getQuantitativeResultAtInitialState(mdp, resultMaxMax), 0.0001);
+}
+
+TEST(RobustMdpModelCheckerTest, RobotRewardsMinMaxTest) {
+#ifndef STORM_HAVE_Z3
+    GTEST_SKIP() << "Z3 not available.";
+#endif
+    // Maxima reachability rewards using PRISM format.
+    checkPrismModelForQuantitativeResult(STORM_TEST_RESOURCES_DIR "/imdp/robot-delta-0.1.prism",
```
tquatmann: The file `robot-delta-0.1.prism` is just a copy of `robot.prism` with `delta=0.1` set, correct? If yes, I'd suggest to avoid this extra file and provide the constants string `delta=0.1` here.

src/storm/solver/UncertaintyResolutionMode.h
```diff
@@ -0,0 +1,28 @@
+#ifndef STORM_UNCERTAINTYRESOLUTIONMODE_H
+#define STORM_UNCERTAINTYRESOLUTIONMODE_H
```
tquatmann:
```suggestion
#pragma once
```

Nowadays, we try to avoid preprocessor include guards (older files still have them)

src/storm/utility/ConstantsComparator.h
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "storm/adapters/IntervalAdapter.h"
```
volkm: Is it possible to keep the include only in the cpp file? This avoids costly includes of carl in header files.

plindnercs: With the current design (`ValueType precision`; stored by value), this is not possible, because `ValueType` needs to be complete at the point where `ConstantsComparator<ValueType>` is instantiated. However, as an alternative, we could change it to `std::shared_ptr<const ValueType> precision`, and use the `precision` variable by dereferencing the pointer in the implementation (`ConstantsComparator.cpp`). This would avoid requiring `ValueType` to be complete in the header. If that is acceptable, then I can change the code accordingly.

src/storm-pars/modelchecker/region/SparseDtmcParameterLiftingModelChecker.h
```diff
@@ -11,6 +11,7 @@
 #include "storm-pars/transformer/ParameterLifter.h"
 #include "storm-pars/transformer/RobustParameterLifter.h"
 #include "storm/solver/MinMaxLinearEquationSolver.h"
+#include "storm/solver/UncertaintyResolutionMode.h"
```
tquatmann: Please move the `#include  ..` to the SparseDtmcParameterLiftingModelChecker.cpp file.

src/storm/utility/ConstantsComparator.h
tquatmann: let's change that to 

```suggestion
#include "storm/adapters/IntervalForward"
```

The pointer solution would yield a runtime overhead

src/storm/settings/modules/MinMaxEquationSolverSettings.cpp
```diff
@@ -97,6 +98,14 @@ MinMaxEquationSolverSettings::MinMaxEquationSolverSettings() : ModuleSettings(mo
                              .setDefaultValueString("all")
                              .build())
             .build());
+
+    std::vector<std::string> uncertaintyResolutionModes = {"minimize", "maximize", "robust", "cooperative", "min", "max"};
+    this->addOption(storm::settings::OptionBuilder(moduleName, uncertaintyResolutionModeName, false, "Mode to resolve the uncertainty (intervals)")
+                        .setIsAdvanced()
```
tquatmann:
```suggestion
```

I feel like this setting shouldn't be advanced

src/storm/settings/modules/MinMaxEquationSolverSettings.cpp
```diff
@@ -22,6 +22,7 @@ const std::string forceUniqueSolutionRequirementOptionName = "force-require-uniq
 const std::string lpEqualityForUniqueActionsOptionName = "lp-eq-unique-actions";
 const std::string lpUseNonTrivialBoundsOptionName = "lp-use-nontrivial-bounds";
 const std::string lpOptimizeOnlyInitialStateOptionName = "lp-objective-type";
+const std::string uncertaintyResolutionModeName = "uncertainty-resolution";
```
tquatmann: I suggest to move this to `IOSettings` because it's part of the query.

src/storm/solver/helper/ValueIterationHelper.h
```diff
@@ -25,16 +26,20 @@ class ValueIterationHelper {
     template<storm::OptimizationDirection Dir, bool Relative>
     SolverStatus VI(std::vector<SolutionType>& operand, std::vector<ValueType> const& offsets, uint64_t& numIterations, SolutionType const& precision,
                     std::function<SolverStatus(SolverStatus const&)> const& iterationCallback = {}, MultiplicationStyle mult = MultiplicationStyle::GaussSeidel,
-                    bool robust = true) const;
+                    std::optional<UncertaintyResolutionMode> const& uncertaintyResolutionMode =
+                        std::make_optional<UncertaintyResolutionMode>(UncertaintyResolutionMode::Unset)) const;
```
tquatmann: Why do we need an std::optional when there already is "Unset"? I.e., right now there are two ways of not setting an uncertainty resolution mode

src/storm/solver/SolveGoal.cpp
```diff
@@ -86,6 +86,11 @@ bool SolveGoal<ValueType, SolutionType>::minimize() const {
 
 template<typename ValueType, typename SolutionType>
 OptimizationDirection SolveGoal<ValueType, SolutionType>::direction() const {
+    if (!optimizationDirection.has_value()) {
+        // required for IDTMCs with explicit uncertainty resolution flag
+        STORM_LOG_WARN("Optimization direction not set, defaulting to maximize.");
```
tquatmann: IMO, this should just THROW or assert that this never happens. The calling code should handle this properly.

src/storm/solver/UncertaintyResolutionMode.cpp
```diff
@@ -0,0 +1,60 @@
...
+    return uncertaintyResolutionMode != UncertaintyResolutionMode::Unset;
+}
+
+bool isUncertaintyResolvedRobust(UncertaintyResolutionMode uncertaintyResolutionMode, OptimizationDirection optimizationDirection) {
+    if (uncertaintyResolutionMode == UncertaintyResolutionMode::Minimize) {
+        if (optimizationDirection != OptimizationDirection::Minimize) {
+            return true;
+        } else {
+            return false;
+        }
```
tquatmann:
```suggestion
        return optimizationDirection != OptimizationDirection::Minimize;
```

src/storm/solver/UncertaintyResolutionMode.cpp
```diff
@@ -0,0 +1,60 @@
...
+        return out << "unset";
+    }
+}
+
+bool isSet(UncertaintyResolutionMode uncertaintyResolutionMode) {
+    return uncertaintyResolutionMode != UncertaintyResolutionMode::Unset;
+}
+
+bool isUncertaintyResolvedRobust(UncertaintyResolutionMode uncertaintyResolutionMode, OptimizationDirection optimizationDirection) {
+    if (uncertaintyResolutionMode == UncertaintyResolutionMode::Minimize) {
```
tquatmann: `switch` statement?

src/storm/solver/UncertaintyResolutionMode.cpp
```diff
@@ -0,0 +1,60 @@
+#include "UncertaintyResolutionMode.h"
+
+namespace storm {
+namespace solver {
+std::ostream& operator<<(std::ostream& out, UncertaintyResolutionMode mode) {
+    if (mode == UncertaintyResolutionMode::Minimize) {
```
tquatmann: Use a `switch` statement?

src/storm/solver/UncertaintyResolutionMode.cpp
```diff
@@ -0,0 +1,60 @@
...
+        return out << "minimize";
+    } else if (mode == UncertaintyResolutionMode::Maximize) {
+        return out << "maximize";
+    } else if (mode == UncertaintyResolutionMode::Robust) {
+        return out << "robust";
+    } else if (mode == UncertaintyResolutionMode::Cooperative) {
+        return out << "cooperative";
+    } else {
+        STORM_LOG_WARN("Uncertainty resolution mode not set.");
```
tquatmann:
```suggestion
```

src/storm/modelchecker/CheckTask.h
```diff
@@ -7,10 +7,14 @@
 #include "storm/logic/Formulas.h"
 #include "storm/utility/constants.h"
 
+#include "storm/adapters/IntervalForward.h"
 #include "storm/logic/ComparisonType.h"
 #include "storm/logic/PlayerCoalition.h"
 #include "storm/modelchecker/hints/ModelCheckerHint.h"
+#include "storm/settings/SettingsManager.h"
+#include "storm/settings/modules/MinMaxEquationSolverSettings.h"
```
tquatmann: Outside of CLI-code, we try to avoid reading from Settings as much as possible, because this causes all kinds of troubles, e.g., with stormpy.

I suggest to extend the `createTask` method in `src/storm/api/verification.h` so that it also reads an UncertaintyResolutionMode. Then, in `model-handling.h` we can pass the settings to the checktask.

src/storm/modelchecker/CheckTask.h
```diff
@@ -299,18 +314,31 @@ class CheckTask {
     }
 
     /*!
-     * Conversion operator that strips the type of the formula.
+     * Retrieves the mode which decides how the uncertainty will be resolved.
      */
-    operator CheckTask<storm::logic::Formula, ValueType>() const {
-        return this->substituteFormula<storm::logic::Formula>(this->getFormula());
```
tquatmann: Was this accidently removed?

src/storm/modelchecker/prctl/helper/SparseDtmcPrctlHelper.cpp
```diff
@@ -41,103 +41,134 @@ std::map<storm::storage::sparse::state_type, storm::RationalFunction> SparseDtmc
...
+    solver->setHasUniqueSolution(false);
+    solver->setHasNoEndComponents(false);
+
+    if (!computeReward) {
+        solver->setLowerBound(storm::utility::zero<SolutionType>());
+        solver->setUpperBound(storm::utility::one<SolutionType>());
     }
 
-    return result;
+    // Solve the corresponding system of equations.
```
tquatmann, line 167: We actually need to check the requirements. Something like

```
auto req = solver->getRequirements(env);
if (!computeReward) {
    req.clearbounds();
}
STORM_LOG_THROW(!req.hasEnabledCriticalRequirement(), storm::exceptions::UncheckedRequirementException,
                        "Solver requirements " + req.getEnabledRequirementsAsString() + " not checked.");
```

src/storm/modelchecker/prctl/helper/SparseDtmcPrctlHelper.cpp
```diff
@@ -185,186 +216,227 @@ std::vector<ValueType> SparseDtmcPrctlHelper<ValueType, RewardModelType>::comput
...
+        }
+    }
+    return result;
+}
+
+template<typename ValueType, typename RewardModelType, typename SolutionType>
+std::vector<SolutionType> SparseDtmcPrctlHelper<ValueType, RewardModelType, SolutionType>::computeAllUntilProbabilities(
+    Environment const& env, storm::solver::SolveGoal<ValueType, SolutionType>&& goal, storm::storage::SparseMatrix<ValueType> const& transitionMatrix,
+    storm::storage::BitVector const& initialStates, storm::storage::BitVector const& phiStates, storm::storage::BitVector const& psiStates) {
+    // Is this method even needed anymore?
```
tquatmann:
```suggestion
```

I believe this is exposed to stormpy? In any case, the comment doesn't contribute anything

src/storm/modelchecker/prctl/helper/SparseDtmcPrctlHelper.cpp
```diff
@@ -438,41 +518,49 @@ std::vector<ValueType> SparseDtmcPrctlHelper<ValueType, RewardModelType>::comput
...
+template<typename ValueType, typename RewardModelType, typename SolutionType>
+std::vector<SolutionType> SparseDtmcPrctlHelper<ValueType, RewardModelType, SolutionType>::computeReachabilityRewards(
+    Environment const& env, storm::solver::SolveGoal<ValueType, SolutionType>&& goal, storm::storage::SparseMatrix<ValueType> const& transitionMatrix,
     storm::storage::SparseMatrix<ValueType> const& backwardTransitions, std::vector<ValueType> const& totalStateRewardVector,
     storm::storage::BitVector const& targetStates, bool qualitative, ModelCheckerHint const& hint) {
     return computeReachabilityRewards(
         env, std::move(goal), transitionMatrix, backwardTransitions,
         [&](uint_fast64_t numberOfRows, storm::storage::SparseMatrix<ValueType> const&, storm::storage::BitVector const& maybeStates) {
-            std::vector<ValueType> result(numberOfRows);
+            std::vector<ValueType> result(numberOfRows, storm::utility::one<ValueType>());
```
tquatmann:
```suggestion
            std::vector<ValueType> result(numberOfRows, storm::utility::zero<ValueType>());
```
non-maybe states should have an expected reward of zero, right?

src/storm/modelchecker/prctl/helper/SparseDtmcPrctlHelper.cpp
```diff
@@ -527,121 +615,171 @@ std::vector<ValueType> SparseDtmcPrctlHelper<ValueType, RewardModelType>::comput
...
     } else {
         if (!maybeStates.empty()) {
-            // Check whether we need to convert the input to equation system format.
-            storm::solver::GeneralLinearEquationSolverFactory<ValueType> linearEquationSolverFactory;
-            bool convertToEquationSystem =
-                linearEquationSolverFactory.getEquationProblemFormat(env) == storm::solver::LinearEquationSolverProblemFormat::EquationSystem;
+            if constexpr (storm::IsIntervalType<ValueType>) {
+                // In this case we have to compute the reward values for the remaining states.
+                // We can eliminate the rows and columns from the original transition probability matrix.
+                storm::storage::SparseMatrix<ValueType> submatrix = transitionMatrix.getSubmatrix(true, maybeStates, maybeStates, false);
```
tquatmann: `submatrix` computed twice
```suggestion
```

tquatmann (changes requested):
Thanks for the contribution! Looks good, I just have some nitpicks :)

src/storm/modelchecker/CheckTask.h
plindnercs: It wasn't removed, it just moved further down (see lines 340-341).

src/storm/utility/ConstantsComparator.h
plindnercs: Incorporating this change results in the following error while executing the tests locally on my machine (same as before adding the include for `IntervalAdapter.h`):
`error: ‘storm::utility::ConstantsComparator<ValueType>::precision’ has incomplete type`.

src/storm/modelchecker/prctl/helper/SparseDtmcPrctlHelper.cpp
plindnercs: Fixed in dee5cad048806fcb12abe6f06f494173561728ba

src/storm/modelchecker/prctl/helper/SparseDtmcPrctlHelper.cpp
plindnercs: Fixed in dee5cad048806fcb12abe6f06f494173561728ba

src/storm/modelchecker/prctl/helper/SparseDtmcPrctlHelper.cpp
plindnercs: Fixed in dee5cad048806fcb12abe6f06f494173561728ba

src/storm/modelchecker/prctl/helper/SparseDtmcPrctlHelper.cpp
plindnercs: Temporarily fixed in dee5cad048806fcb12abe6f06f494173561728ba. I hope this is okay for now.

src/storm/settings/modules/MinMaxEquationSolverSettings.cpp
plindnercs: Fixed in dee5cad048806fcb12abe6f06f494173561728ba

src/storm/modelchecker/CheckTask.h
plindnercs: Fixed in dee5cad048806fcb12abe6f06f494173561728ba

src/storm/solver/UncertaintyResolutionMode.cpp
plindnercs: Fixed in dee5cad048806fcb12abe6f06f494173561728ba

src/storm/solver/UncertaintyResolutionMode.cpp
plindnercs: Fixed in dee5cad048806fcb12abe6f06f494173561728ba

src/storm/solver/UncertaintyResolutionMode.cpp
plindnercs: Fixed in dee5cad048806fcb12abe6f06f494173561728ba

src/storm/solver/UncertaintyResolutionMode.cpp
plindnercs: Fixed in dee5cad048806fcb12abe6f06f494173561728ba

src/storm/solver/helper/ValueIterationHelper.h
plindnercs: Fixed in dee5cad048806fcb12abe6f06f494173561728ba

src/storm/settings/modules/MinMaxEquationSolverSettings.cpp
plindnercs: Fixed in dee5cad048806fcb12abe6f06f494173561728ba

src/storm-pars/modelchecker/region/SparseDtmcParameterLiftingModelChecker.h
plindnercs: Fixed in dee5cad048806fcb12abe6f06f494173561728ba

src/storm/solver/UncertaintyResolutionMode.h
plindnercs: Fixed in dee5cad048806fcb12abe6f06f494173561728ba

src/test/storm/modelchecker/prctl/mdp/RobustMdpPrctlModelCheckerTest.cpp
plindnercs: Fixed in dee5cad048806fcb12abe6f06f494173561728ba

src/storm/solver/SolveGoal.cpp
plindnercs: fixed in 4e0ccd8d1a1e9e033171b193bc963eadf6c8a95d

sjunges:
What does it even mean that we use an interval as precision in the constants comparator?

volkm:
The following patch should handle the includes for ConstantsComparator.
[patch.txt](https://github.com/user-attachments/files/24430823/patch.txt)

src/storm/modelchecker/prctl/helper/SparseDtmcPrctlHelper.cpp
tquatmann: I think we should throw already at this point. If the solver requires an upper bound and we ignore that requirement, it will fail or (even worse) terminate with a wrong result.

On another note: We should have negative rewards in mind. This means that 0 is not necessarily a lower bound. (I know there are still some parts that assume non-negative rewards).

Is there an issue with  my suggestion?

src/storm/settings/modules/IOSettings.cpp
```diff
@@ -520,6 +529,28 @@ bool IOSettings::isPropertiesAsMultiSet() const {
...
+    if (uncertaintyResolutionModeString == "minimize" || uncertaintyResolutionModeString == "min") {
+        return UncertaintyResolutionModeSetting::Minimize;
+    } else if (uncertaintyResolutionModeString == "maximize" || uncertaintyResolutionModeString == "max") {
+        return UncertaintyResolutionModeSetting::Maximize;
+    } else if (uncertaintyResolutionModeString == "robust") {
+        return UncertaintyResolutionModeSetting::Robust;
+    } else if (uncertaintyResolutionModeString == "cooperative") {
+        return UncertaintyResolutionModeSetting::Cooperative;
+    } else if (uncertaintyResolutionModeString == "both") {
+        STORM_LOG_ERROR("Uncertainty resolution mode 'both' not yet supported.");
```
tquatmann: ```STORM_LOG_ASSERT(false ...```

with the validator above, this should not be reachable. It's better to just fail for now.

src/storm/solver/helper/ValueIterationHelper.cpp
```diff
@@ -102,14 +102,14 @@ SolverStatus ValueIterationHelper<ValueType, TrivialRowGrouping, SolutionType>::
...
 template<storm::OptimizationDirection Dir, bool Relative>
-SolverStatus ValueIterationHelper<ValueType, TrivialRowGrouping, SolutionType>::VI(
-    std::vector<SolutionType>& operand, std::vector<ValueType> const& offsets, uint64_t& numIterations, SolutionType const& precision,
-    std::function<SolverStatus(SolverStatus const&)> const& iterationCallback, MultiplicationStyle mult,
-    std::optional<UncertaintyResolutionMode> const& uncertaintyResolutionMode) const {
+SolverStatus ValueIterationHelper<ValueType, TrivialRowGrouping, SolutionType>::VI(std::vector<SolutionType>& operand, std::vector<ValueType> const& offsets,
+                                                                                   uint64_t& numIterations, SolutionType const& precision,
+                                                                                   const std::function<SolverStatus(const SolverStatus&)>& iterationCallback,
+                                                                                   MultiplicationStyle mult,
+                                                                                   const UncertaintyResolutionMode& uncertaintyResolutionMode) const {
```
tquatmann: For consistency: `UncertaintyResolutionMode const&`

src/storm/solver/UncertaintyResolutionMode.cpp
```diff
@@ -3,50 +3,45 @@
...
+        case UncertaintyResolutionMode::Minimize:
+            return optimizationDirection != OptimizationDirection::Minimize;
+        case UncertaintyResolutionMode::Robust:
             return true;
-        } else {
+        case UncertaintyResolutionMode::Cooperative:
             return false;
-        }
+        case UncertaintyResolutionMode::Unset:
+            STORM_LOG_WARN("Uncertainty resolution mode not set properly, assuming to resolve uncertainty in a robust fashion.");
```
tquatmann: `STORM_LOG_THROW(false, ...`

volkm:
> **Edit**: I realise that this, again, requires the `IntervalForward` include in the ConstantsComparator.
> I don't know how to prevent that. @volkm any idea?

I think the `IntervalForward` is okay to include. Up until now, I managed to avoid including it, because ValueType is only later instantiated in the cpp and it was somehow okay to delay the actual include to the class which later uses it, for example in the tests. Maybe that is still possible.

src/storm/settings/modules/IOSettings.cpp
plindnercs: fixed in 84e552b7a49f37aab9e9b2445728a6d6be960763.

src/storm/solver/helper/ValueIterationHelper.cpp
plindnercs: fixed in 84e552b7a49f37aab9e9b2445728a6d6be960763.

src/storm/solver/UncertaintyResolutionMode.cpp
plindnercs: fixed in 84e552b7a49f37aab9e9b2445728a6d6be960763.

tquatmann:
> > > > @plindnercs so you use two precisions, one for the lower bound and one for the upper bound? Is this a hypothetical use case, or a real one?
> > > 
> > > 
> > > Yes, you can see it that way. Right now, it is only a hypothetical use case as far as I know. At least the implementation that I contributed does not make use of this directly.
> > 
> > 
> > I'd be in favour of using the most obvious semantics for now, until something else is required. To me, "most obvious semantics" would be to have `IntervalBaseType<ValueType>` as type of `precision` and then apply that precision to both, lower and upper interval boundaries
> 
> I fully second this. In particular, I would not even allow instantiating the constants comparator with intervals...

I created issue #860 for this discussion.  That shouldn't be a blocker for this PR as it no longer touches the constants comparator.

PR #812, opened 2025-11-17
Updated parallel-hashmap to commit 4e410c7

volkm:
- Renamed directory `parallel_hashmap` to `parallel-hashmap` to be consistent with original repository
- Removed `.github` directory


sjunges:
I am fine with merging, but maybe let @tquatmann chime in? :-)

volkm:
By the way: we could also consider moving to gtl as suggested [by the author](https://github.com/greg7mdp/parallel-hashmap?tab=readme-ov-file#parallel-hashmap-or-gtl)

PR #819, opened 2025-11-23
Cleanup of some includes

volkm:
- Consistent order of includes for tests: First include `storm-config.h` and `test/storm_gtest.h`, then the rest. Added remarks in doc/developers.
- Use `RationalFunctionForward` instead of `RationalFunctionAdapter` in more header files
- Replaced `#include <storm...>` by `#include "storm..."`

### DIFF
--- a/src/storm/analysis/GraphConditions.h
+++ b/src/storm/analysis/GraphConditions.h
@@ -3,7 +3,10 @@
 #include <type_traits>
 #include <unordered_set>
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wmacro-redefined"  // clash for likely() macro between Ginac and Sylvan
 #include <carl/formula/Formula.h>
+#pragma clang diagnostic pop
 
 #include "storm/adapters/RationalFunctionAdapter.h"
 #include "storm/models/sparse/Dtmc.h"


src/storm-dft/modelchecker/DFTASFChecker.cpp
```diff
@@ -659,13 +660,13 @@ storm::solver::SmtSolver::CheckResult DFTASFChecker::checkDependencyConflict(uin
     STORM_LOG_DEBUG("Check " << dft.getElement(dep1Index)->name() << " and " << dft.getElement(dep2Index)->name());
     andConstr.clear();
     // AND FDEP1 is triggered before FDEP2 is resolved
-    andConstr.push_back(std::make_shared<IsGreaterEqual>(timePointVariables.at(dep1Index), timePointVariables.at(dep2Index)));
+    andConstr.push_back(std::make_shared<IsLessEqual>(timePointVariables.at(dep2Index), timePointVariables.at(dep1Index)));
```
sjunges: Semantics are changing here, that is potentially unexpected in this PR?

src/test/storm_gtest.cpp
```diff
@@ -1,14 +1,51 @@
 #include "test/storm_gtest.h"
 
+#ifdef STORM_HAVE_Z3
```
sjunges, line 3: nice to move more here!

src/storm-dft/modelchecker/DFTASFChecker.cpp
volkm: The semantics should not change through this step because it is just a rewrite. But it might indeed be a bit confusing to change it here.
Let me make a separate PR for this.

volkm: Separate PR #821

PR #826, opened 2025-11-28
Preparation for using probability matrix in CTMC bisimulation

volkm:
As preparation for the CTMC change from rate matrix to probability matrix, I saw that the bisimulation needs a rate matrix. This PR uses adds a members `transitionMatrix` such that both probability and rate matrix could be used.

Also cleaned up the pragmas and includes in `storm/storage/bisimulation`.

### DIFF
--- a/.github/workflows/Dockerfile.alpine
+++ b/.github/workflows/Dockerfile.alpine
@@ -6,7 +6,7 @@
 # --build-arg BASE_IMAGE=<new_base_image>
 
 # Set base image
-ARG BASE_IMAGE=alpine:latest
+ARG BASE_IMAGE=alpine:edge
 FROM $BASE_IMAGE
 LABEL org.opencontainers.image.authors="dev@stormchecker.org"
 
@@ -44,8 +44,6 @@ ARG cmake_args=""
 # Additional Alpine packages
 ARG packages=""
 
-RUN echo "@testing https://dl-cdn.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories
-RUN echo "@community https://dl-cdn.alpinelinux.org/alpine/edge/community" >> /etc/apk/repositories
 RUN apk add --no-cache \
     bash \
     cmake \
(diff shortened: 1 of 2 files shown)


volkm:
Using const references did not work, so I will try to make specialized versions for CTMC.

src/storm/storage/bisimulation/DeterministicModelBisimulationDecomposition.cpp
```diff
@@ -722,18 +716,54 @@ void DeterministicModelBisimulationDecomposition<ModelType>::buildQuotient() {
...
 
     // Finally construct the quotient model.
-    this->quotient = std::shared_ptr<ModelType>(new ModelType(builder.build(), std::move(newLabeling), std::move(rewardModels)));
+    this->quotient = std::make_shared<ModelType>(builder.build(), std::move(newLabeling), std::move(rewardModels));
+}
+
+template<typename ModelType>
+DeterministicModelBisimulationDecomposition<ModelType>::ValueType DeterministicModelBisimulationDecomposition<ModelType>::getTransitionValue(
+    storm::storage::MatrixEntry<storm::storage::sparse::state_type, ValueType> const& matrixEntry, storm::storage::sparse::state_type state) const {
+    return matrixEntry.getValue();
```
tquatmann: I'd suggest something like this to avoid the specialisations for all ValueTypes

```suggestion
    if constexpr (std::is_same_v<ModelType, storm::models::sparse::Ctmc<typename ModelType::ValueType>) {
    auto transitionValue = matrixEntry.getValue()
    // TODO: enable when removing CTMC rate matrix
    // transitionValue *= this->model.getExitRateVector().at(state);
    return transitionValue;
} else {
STORM_LOG_ASSERT(!this->model.isDiscreteTimeModel(), "Unhandled model type");
    return matrixEntry.getValue();
```

tquatmann (approved):
LGTM! Thanks :)

src/storm/storage/bisimulation/DeterministicModelBisimulationDecomposition.cpp
volkm: Thanks for the suggestion. I have integrated it.

volkm:
I will merge it for now. The issue with the CTMC bisimulation quotient is tracked in https://github.com/moves-rwth/storm/issues/833.

PR #830, opened 2025-11-29
ConstantsComparator no longer depends on global state

sjunges:
The constant comparator was one of the deepest code parts that depends on the settings. This PR reduces this dependency and requires to set the precision explicitly. 

The consequence is that some parts now (explicitly) depend on this setting, but I also didnt want to escalate the PR. 

Contributes towards #764

### DIFF
--- a/src/storm/utility/ConstantsComparator.cpp
+++ b/src/storm/utility/ConstantsComparator.cpp
@@ -45,7 +45,7 @@ bool ConstantsComparator<ValueType>::isEqual(ValueType const& value1, ValueType 
 template<typename ValueType>
 bool ConstantsComparator<ValueType>::isLess(ValueType const& value1, ValueType const& value2) const {
     STORM_LOG_ASSERT(!relative, "Relative precision and constants comparator is currently not supported.");
-    return value1 - precision < value2;
+    return value1 < value2 - precision;
 }
 
 // Explicit instantiations.


src/storm-dft/builder/DFTBuilder.cpp
```diff
@@ -179,7 +179,7 @@ void DFTBuilder<ValueType>::addBasicElementErlang(std::string const& name, Value
         // shape=1 reduces to exponential distribution
         addBasicElementExponential(name, rate, dormancyFactor);
     } else {
-        STORM_LOG_THROW(this->comparator.isLess(storm::utility::zero<ValueType>(), rate), storm::exceptions::WrongFormatException,
+        STORM_LOG_THROW(storm::utility::isPositive(rate), storm::exceptions::WrongFormatException,
```
volkm:
```suggestion
        STORM_LOG_THROW(!storm::utility::isConstant(rate) || storm::utility::isPositive(rate), storm::exceptions::WrongFormatException,
```

src/storm-dft/builder/DFTBuilder.cpp
```diff
@@ -190,9 +190,9 @@ void DFTBuilder<ValueType>::addBasicElementErlang(std::string const& name, Value
 
 template<typename ValueType>
 void DFTBuilder<ValueType>::addBasicElementWeibull(std::string const& name, ValueType shape, ValueType rate) {
-    STORM_LOG_THROW(this->comparator.isLess(storm::utility::zero<ValueType>(), rate), storm::exceptions::WrongFormatException,
+    STORM_LOG_THROW(storm::utility::isPositive(rate), storm::exceptions::WrongFormatException,
```
volkm:
```suggestion
    STORM_LOG_THROW(!storm::utility::isConstant(rate) || storm::utility::isPositive(rate), storm::exceptions::WrongFormatException,
```

src/storm-dft/builder/DFTBuilder.cpp
```diff
@@ -190,9 +190,9 @@ void DFTBuilder<ValueType>::addBasicElementErlang(std::string const& name, Value
...
                     "Weibull distribution of BE " << name << " requires a positive scale.");
-    STORM_LOG_THROW(this->comparator.isLess(storm::utility::zero<ValueType>(), shape), storm::exceptions::WrongFormatException,
+    STORM_LOG_THROW(storm::utility::isPositive(shape), storm::exceptions::WrongFormatException,
```
volkm:
```suggestion
    STORM_LOG_THROW(!storm::utility::isConstant(shape) || storm::utility::isPositive(shape), storm::exceptions::WrongFormatException,
```

src/storm-dft/builder/DFTBuilder.cpp
```diff
@@ -206,7 +206,7 @@ void DFTBuilder<ValueType>::addBasicElementWeibull(std::string const& name, Valu
 
 template<typename ValueType>
 void DFTBuilder<ValueType>::addBasicElementLogNormal(std::string const& name, ValueType mean, ValueType standardDeviation) {
-    STORM_LOG_THROW(this->comparator.isLess(storm::utility::zero<ValueType>(), standardDeviation), storm::exceptions::WrongFormatException,
+    STORM_LOG_THROW(storm::utility::isPositive(standardDeviation), storm::exceptions::WrongFormatException,
```
volkm:
```suggestion
    STORM_LOG_THROW(!storm::utility::isConstant(standardDeviation) || storm::utility::isPositive(standardDeviation), storm::exceptions::WrongFormatException,
```

src/storm/utility/constants.h
```diff
@@ -80,6 +79,18 @@ ValueType zero();
...
+bool isApproxEqual(ValueType const& a, ValueType const& b, ValueType const& precision, bool relative);
+
+template<typename ValueType>
+bool isPositive(ValueType const& a);
+
+template<typename ValueType>
+bool isNonNegative(ValueType const& a);
+
+template<typename ValueType>
+bool isBetween(ValueType const& a, ValueType const& b, ValueType const& c, bool strict = false);
```
volkm, line 95: Can we add some documentation?

```suggestion
/*!
 * Check whether a <= b <= c holds (if not strict, the default) or a < b < c (if strict).
 */
template<typename ValueType>
bool isBetween(ValueType const& a, ValueType const& b, ValueType const& c, bool strict = false);
```

src/storm/utility/constants.cpp
```diff
@@ -53,6 +49,37 @@ bool isNan(double const& value) {
     return std::isnan(value);
 }
 
+template<typename ValueType>
+bool isApproxEqual(ValueType const& a, ValueType const& b, ValueType const& precision, bool relative) {
+    ValueType const absDiff = storm::utility::abs<ValueType>(a - b);
+    if (relative) {
+        return absDiff <= precision * (storm::utility::abs(a) + storm::utility::abs(b));
```
volkm: I suggest to incorporate the change from https://github.com/moves-rwth/storm/pull/768:
```suggestion
        return absDiff <= precision * storm::utility::max(storm::utility::abs(a), storm::utility::abs(b));
```

src/storm/utility/constants.cpp
```diff
@@ -1033,6 +1092,9 @@ template bool isOne(int const& value);
 template bool isZero(int const& value);
 template bool isConstant(int const& value);
 template bool isInfinity(int const& value);
+template bool isNonNegative(int const& value);
+template bool isPositive(int const& value);
+template bool isApproxEqual(int const& a, int const& b, int const& precision, bool relative);
```
volkm, line 1105: IsBetween missing?

src/storm/utility/constants.cpp
```diff
@@ -1041,15 +1103,21 @@ template uint32_t infinity();
 template bool isOne(uint32_t const& value);
 template bool isZero(uint32_t const& value);
 template bool isConstant(uint32_t const& value);
+template bool isNonNegative(uint32_t const& value);
+template bool isPositive(uint32_t const& value);
```
volkm, line 1116: IsBetween missing?

src/storm/utility/constants.cpp
```diff
@@ -1041,15 +1103,21 @@ template uint32_t infinity();
...
 template storm::storage::sparse::state_type one();
 template storm::storage::sparse::state_type zero();
 template storm::storage::sparse::state_type infinity();
+template bool isApproxEqual(storm::storage::sparse::state_type const& a, storm::storage::sparse::state_type const& b,
+                            storm::storage::sparse::state_type const& precision, bool relative);
 template bool isOne(storm::storage::sparse::state_type const& value);
 template bool isZero(storm::storage::sparse::state_type const& value);
 template bool isConstant(storm::storage::sparse::state_type const& value);
+template bool isPositive(storm::storage::sparse::state_type const& value);
+template bool isNonNegative(storm::storage::sparse::state_type const& value);
```
volkm, line 1130: IsBetween missing?

src/storm/utility/constants.cpp
```diff
@@ -1128,9 +1200,9 @@ template bool isOne(Interval const& value);
 template bool isZero(Interval const& value);
 template bool isInfinity(Interval const& value);
 template bool isAlmostZero(Interval const& value);
-
+template bool isNonNegative(Interval const& value);
+template bool isPositive(Interval const& value);
```
volkm, line 1216: IsBetween missing?

src/storm/utility/ConstantsComparator.cpp
```diff
@@ -1,94 +1,50 @@
...
-template<typename ValueType>
-bool ConstantsComparator<ValueType, ConstantsComparatorEnablePrecision<ValueType>>::isInfinity(ValueType const& value) const {
-    return storm::utility::isInfinity(value);
-}
-
-template<typename ValueType>
-bool ConstantsComparator<ValueType, ConstantsComparatorEnablePrecision<ValueType>>::isLess(ValueType const& value1, ValueType const& value2) const {
+bool ConstantsComparator<ValueType>::isLess(ValueType const& value1, ValueType const& value2) const {
+    STORM_LOG_ASSERT(!relative, "Relative precision and constants comparator is currently not supported.");
     return value1 < value2 - precision;
```
volkm, line 48: Might not always be correct: `isLess(0.9, 1)` with precision 0.2 yields False because of 0.9 < 1 - 0.2
Using `value1 - precision < value2` as suggested in https://github.com/moves-rwth/storm/pull/768 would yield True because of 0.9-0.2 < 1. This might have some other problems though, for example `isLess(0, 0.1)` with precision 0.2 yields True.

src/storm/storage/SparseMatrix.cpp
```diff
@@ -2259,17 +2259,17 @@ typename SparseMatrix<ValueType>::index_type SparseMatrix<ValueType>::getNoncons
...
         auto rowSum = getRowSum(row);
         if (!comparator.isOne(rowSum)) {
             return false;
         }
     }
     for (auto const& entry : *this) {
-        if (comparator.isConstant(entry.getValue())) {
-            if (comparator.isLess(entry.getValue(), storm::utility::zero<ValueType>())) {
+        if (storm::utility::isConstant(entry.getValue())) {
+            if (!storm::utility::isNonNegative(entry.getValue())) {
```
volkm, line 2272: Should we not check for in between 0 and 1 here?
```suggestion
            if (!storm::utility::isBetween(storm::utility::zero<ValueType>(), entry.getValue(), storm::utility::one<ValueType>())) {
```

src/storm/models/sparse/Model.cpp
```diff
@@ -105,7 +111,7 @@ void Model<ValueType, RewardModelType>::assertValidityOfComponents(
     } else if (this->isOfType(ModelType::S2pg)) {
         STORM_LOG_THROW(components.player1Matrix.is_initialized(), storm::exceptions::IllegalArgumentException,
                         "No player 1 matrix given for stochastic game.");
-        STORM_LOG_ASSERT(components.player1Matrix->isProbabilistic(),
+        STORM_LOG_ASSERT(components.player1Matrix->isProbabilistic(0),
```
volkm, line 114: Why do we not use stochasticTolerance here?

src/storm/models/sparse/MarkovAutomaton.cpp
```diff
@@ -164,17 +162,12 @@ void MarkovAutomaton<ValueType, RewardModelType>::turnRatesToProbabilities() {
...
         } else {
             if (assertRates) {
-                STORM_LOG_THROW(comparator.isZero(this->exitRates[state]), storm::exceptions::InvalidArgumentException,
+                STORM_LOG_THROW(storm::utility::isZero(this->exitRates[state]), storm::exceptions::InvalidArgumentException,
                                 "The specified exit rate for (non-Markovian) choice should be 0.");
             } else {
                 this->exitRates.push_back(storm::utility::zero<ValueType>());
             }
         }
-        for (; row < this->getTransitionMatrix().getRowGroupIndices()[state + 1]; ++row) {
```
volkm, line 173: It would be good to keep it as an assertion, because turning rates to components is done after the model is created from components. Thus, the assertValidityOfComponents check was already done, but we should still check whether the transition are probabilistic.

src/storm/modelchecker/helper/indefinitehorizon/visitingtimes/SparseDeterministicVisitingTimesHelper.cpp
```diff
@@ -40,8 +40,7 @@ SparseDeterministicVisitingTimesHelper<ValueType>::SparseDeterministicVisitingTi
       backwardTransitions(storm::NullRef),
       sccDecomposition(storm::NullRef),
       nonBsccStates(transitionMatrix.getRowCount(), false) {
-    // For the CTMC case we assert that the caller actually provided the probabilistic transitions
-    STORM_LOG_ASSERT(this->transitionMatrix.isProbabilistic(), "Non-probabilistic transitions");
+    // Note that we are not checking assumptions regarding well-formedness of the input, even in debug mode.
```
volkm, line 43: Why not? This could be useful when going from rate matrix to probability matrix for CTMCs.

src/storm/modelchecker/helper/infinitehorizon/SparseDeterministicInfiniteHorizonHelper.cpp
```diff
@@ -38,10 +38,7 @@ SparseDeterministicInfiniteHorizonHelper<ValueType>::SparseDeterministicInfinite
 template<typename ValueType>
 SparseDeterministicInfiniteHorizonHelper<ValueType>::SparseDeterministicInfiniteHorizonHelper(storm::storage::SparseMatrix<ValueType> const& transitionMatrix,
                                                                                               std::vector<ValueType> const& exitRates)
-    : SparseInfiniteHorizonHelper<ValueType, false>(transitionMatrix, exitRates) {
-    // For the CTMC case we assert that the caller actually provided the probabilistic transitions
-    STORM_LOG_ASSERT(this->_transitionMatrix.isProbabilistic(), "Non-probabilistic transitions");
```
volkm, line 43: See previous comment why the assertion is removed.

src/storm/modelchecker/prctl/helper/DsMpiUpperRewardBoundsComputer.cpp
```diff
@@ -105,9 +105,11 @@ ValueType DsMpiDtmcUpperRewardBoundsComputer<ValueType>::computeLambdaForChoice(
         for (auto const& e : transitionMatrix.getRow(choice)) {
             rewardSum += e.getValue() * w[e.getColumn()];
         }
-        STORM_LOG_WARN_COND(w[state] >= rewardSum || storm::utility::ConstantsComparator<ValueType>().isEqual(w[state], rewardSum),
+        // The following is a bit of a hack but I'd prefer not getting the settings into this part of the code for such a simple check.
+        storm::utility::ConstantsComparator<ValueType> cc(storm::utility::convertNumber<ValueType>(0.0001));
```
volkm, line 109: Maybe we could use an environment here in the future to get the desired precision?

src/storm/models/sparse/Model.cpp
sjunges: Because this is a 0-1 matrix

volkm: Okay, that makes sense.

src/storm/utility/constants.cpp
sjunges: Thanks for spotting this!

src/storm/utility/constants.cpp
sjunges: Yes... isBetween is only used for values so I didnt need it. I will add them.

src/storm/utility/ConstantsComparator.cpp
sjunges: I think the value1 < value2 + precision is indeed better. Sorry I missed the changes from #768. I expected that a merge would retain this, it didnt.

volkm: No worries. We never merged #768, so I just wanted to make sure that some of the ideas are incorporated here.

src/storm/models/sparse/MarkovAutomaton.cpp
sjunges: I found this check (as an exception) a bit much. In the same routine, a few lines earlier, we explicitly compute these values that should by construction now sum to one. We then check once more if they do. I can see that they were useful during debugging and I am not against additional asserts. 

But also here, there is then the question: What precision do we use. I am happy to include environments or magic constants, but maybe less so for assertions that I would consider rather arbitrarily located. (note that we already check whether the rates sum to the exit rate, so it really is just checking whether \sum (e_i / r ) = 1, given that \sum e_i = r ...

src/storm/modelchecker/prctl/helper/DsMpiUpperRewardBoundsComputer.cpp
sjunges: Yes, but ultimately, this is similar to the other cases (only that the way the check is implemented is more targeted towards specific debugging)

src/storm/utility/ConstantsComparator.cpp
sjunges: If we change this, we break the code on a lot of places (see tests). I therefore do not think that this change / fix should be part of this PR.

Issue #831, opened 2025-12-02
storm::utility::isLess is broken for nonzero precision/tolerance. 


sjunges:
This came up in #830 regarding the implementation of storm::utility::isLess.

Might not always be correct: `isLess(0.9, 1)` with precision 0.2 yields False because of 0.9 < 1 - 0.2
Using `value1 - precision < value2` as suggested in https://github.com/moves-rwth/storm/pull/768 would yield True because of 0.9-0.2 < 1. This might have some other problems though, for example `isLess(0, 0.1)` with precision 0.2 yields True.

_Originally posted by @volkm in https://github.com/moves-rwth/storm/pull/830#discussion_r2576450417_

However, fixing this causes a plethora of tests to fail as we've likely worked around the buggy behavior preivously.

PR #837, opened 2025-12-03
Replaced isLess in DFTModelChecker

volkm:
Addresses some parts of https://github.com/moves-rwth/storm/issues/831


volkm:
My main reason was that `isLess` requires a `ConstantsComparator`. But I agree that it would be more clear to use `isLess` or even better `isLessEqual`.
Maybe a more general question: when should we use `constants.h` and when `ConstantsComparator`? It seems that both overlap in some functionality.

sjunges:
I think we haven't used < because the code has to also work for polynomials etc and the intended semantics of the < operator is not  clear?

PR #847, opened 2025-12-24
Improvements for conditional bisection method

tquatmann:
* Use acyclic min max solver when possible
* enable caching of solver data to avoid some re-allocations
* More careful treatment of exact computations

### DIFF
--- a/src/storm/modelchecker/helper/conditional/ConditionalHelper.cpp
+++ b/src/storm/modelchecker/helper/conditional/ConditionalHelper.cpp
@@ -441,13 +441,13 @@ class WeightedReachabilityHelper {
             }
             cachedSolver = storm::solver::GeneralMinMaxLinearEquationSolverFactory<ValueType, SolutionType>().create(solverEnv, submatrix);
             cachedSolver->setCachingEnabled(true);
-            cachedSolver->setOptimizationDirection(dir);
             cachedSolver->setRequirementsChecked();
             cachedSolver->setHasUniqueSolution(true);
             cachedSolver->setHasNoEndComponents(true);
             cachedSolver->setLowerBound(-storm::utility::one<ValueType>());
             cachedSolver->setUpperBound(storm::utility::one<ValueType>());
         }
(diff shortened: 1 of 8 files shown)


src/storm/modelchecker/helper/conditional/ConditionalHelper.cpp
```diff
@@ -617,22 +632,36 @@ SolutionType computeViaBisection(Environment const& env, BisectionMethodBounds b
...
-            auto const exactMiddle = middle;
-
-            // Find number of digits - 1. Method using log10 does not work since that uses doubles internally.
-            auto numDigits = storm::utility::numDigits<SolutionType>(*upperBound - *lowerBound) - 1;
-
-            do {
-                ++numDigits;
-                middle = storm::utility::kwek_mehlhorn::sharpen<SolutionType, SolutionType>(numDigits, exactMiddle);
-            } while (middle <= *lowerBound || middle >= *upperBound);
+            // Check if the rationalCandidate has been within the bounds for four iterations.
```
sjunges, line 684: Do we also do this if we are just searching to bound  the value, rather than computing an exact number?

tquatmann: Good question, we don't seem to have a dedicated code path for thresholded properties. 
@lukovdm do you already have something for that? If not, I can add it here.

sjunges: For stuff like VI, we do have terminationconditions, right? Would they be applicable here?

tquatmann: With a bound value, we just need to invoke the WeightedReachabilityHelper once. I would probably write a dedicated method for that. Basically a stripped down variant of `computeViaBisection`.

We can and should use terminationconditions for the restart method, though.

Fortunately, we already pass the SolveGoal (which includes the bound) to the ConditionalHelper. So it is a relatively small change.

PR #852, opened 2026-01-05
Updated Dockerfile for Alpine

volkm:
Fixed Dockerfile for Alpine after recent failures in CI.

Use Docker tag `edge` instead of `latest`.
@sjunges Is this tag correct or should we use the latest release?

Packages `cln-dev` and `ginac-dev` are part of `community` now. Added `patch` packages into Dockerfile.


volkm:
Good point. I have added a [PR](https://github.com/moves-rwth/carl-storm/pull/110) for carl as well. Stormpy is not using alpine.

PR #853, opened 2026-01-09
IMDPs: Improved support for reachability rewards and reachability probabilities

tquatmann:
Work in Progress. Merge #808 first.

IMDPs: Improved support for reachability rewards and reachability probabilities

- The graph algorithms currently do not handle non-graph-preserving intervals of the form [0,x] for x>0 correctly. This PR adds a check and rejects such inputs.
- SparseMdpPrctlModelChecker::canHandle now also accepts reachability rewards for IMDPs
- Enable end component elimination for IMDP models (which works as for MDPs as long as we can assume graph-preservation).
- Allow scheduler export via command line


sjunges:
How does the EC elimination without assuming graph-preservation @linusheck  compare to this?

PR #857, opened 2026-01-28
Only require MultiObjectiveSettings if needed

volkm:
Running `storm-pars` lead to an error `Cannot retrieve unknown module 'multiobjective'.`. This is fixed now by only requiring the setting if needed. There might be a more elegant solution though.

Was introduced with https://github.com/moves-rwth/storm/pull/805

### DIFF
--- a/src/storm-cli-utilities/model-handling.h
+++ b/src/storm-cli-utilities/model-handling.h
@@ -430,6 +430,7 @@ inline std::pair<SymbolicInput, ModelProcessingInformation> preprocessSymbolicIn
     if (ioSettings.isPropertiesAsMultiSet()) {
         STORM_LOG_THROW(!input.properties.empty(), storm::exceptions::InvalidArgumentException,
                         "Can not translate properties to multi-objective formula because no properties were specified.");
+        // If we come from storm-pars, the following fails as multiObjectiveSettings are not loaded
         auto multiObjSettings = storm::settings::getModule<storm::settings::modules::MultiObjectiveSettings>();
         output.properties = {storm::api::createMultiObjectiveProperty(output.properties, multiObjSettings.isLexicographicModelCheckingSet())};
     }


sjunges:
I don't understand how this fixes the problem. Wouldnt the fix be that storm-pars needs to initialize multi-objective settings?

volkm:
Let's call it a workaround ;). The proper fix would indeed be to initialize the multi-objective setting in storm-pars as well. However, this settings module should not be relevant for storm-pars because we cannot do multi-objective queries on parametric models (as far as I know). So adding this could introduce some confusion.

Maybe the best way for now would be to indeed add the setting to storm-pars.
Then long-term, we could try to separate some functionality in `model-handling.h` which can be used in general (storm, storm-pars, storm-pomdp) and specific functions which are only relevant for some binaries (such as functions for multi-objective, counter-examples, game-based abstraction, etc.)

src/storm-cli-utilities/model-handling.h
```diff
@@ -423,14 +423,14 @@ inline void ensureNoUndefinedPropertyConstants(std::vector<storm::jani::Property
...
     auto ioSettings = storm::settings::getModule<storm::settings::modules::IOSettings>();
-    auto multiObjSettings = storm::settings::getModule<storm::settings::modules::MultiObjectiveSettings>();
 
     SymbolicInput output = input;
 
     // Preprocess properties (if requested)
     if (ioSettings.isPropertiesAsMultiSet()) {
         STORM_LOG_THROW(!input.properties.empty(), storm::exceptions::InvalidArgumentException,
                         "Can not translate properties to multi-objective formula because no properties were specified.");
+        auto multiObjSettings = storm::settings::getModule<storm::settings::modules::MultiObjectiveSettings>();
```
tquatmann, line 434:
```suggestion
        // If we come from storm-pars, the following fails as multiObjectiveSettings are not loaded
        auto multiObjSettings = storm::settings::getModule<storm::settings::modules::MultiObjectiveSettings>();
```

tquatmann: Right now, if a user sets --propsasmulti, there will be again an error saying that multiObjectiveSettings are missing. I was about to suggest a `STORM_LOG_THROW` here for a more readable error ("Multi objective is not supported in storm-pars ..."), but apparently its not known that we are loading a parametric model in this method and changing the method signature seems overkill.

sjunges:
Given the discusson above, lets merge this but maybe track this in an issue

PR #864, opened 2026-02-25
Upgrade to Eigen 5.0.1

tquatmann:
Fixes #791
Fixes #354

PR #865, opened 2026-02-25
Removed eigen package from Dockerfiles

volkm:
Following discussion in https://github.com/moves-rwth/storm/pull/864


volkm:
Was a duplicate of #866 and the result is an [empty commit](https://github.com/moves-rwth/storm/commit/143ce7224a4eae73f66fb13447952d8d6f1fb354) 😄. Interesting that Github does not prevent such merges...

PR #866, opened 2026-02-25
Remove eigen from Dockerfile.archlinux

sjunges:
Removed 'eigen' from the list of installed packages on arch, relates to #865


volkm:
LGTM.

PR #868, opened 2026-02-27
Improvements for multi-objective model checking

tquatmann:
* multi-objective model checking in --sound mode now handles approximation errors correctly.
* source code: simplified Pareto curve approximation algorithm (PCAA) implementation by unifying the different query types
* added `--multiobjective:approxtradeoff` setting to control Pareto curve approximation heuristic.
* source code: simplified weight vector checker interface

Note: This might affect performance, in particular in `--sound` mode which previously just ignored any approximation errors in the underlying single-objective calls


src/storm/environment/modelchecker/MultiObjectiveModelCheckerEnvironment.h
```diff
@@ -56,6 +56,11 @@ class MultiObjectiveModelCheckerEnvironment {
     bool getUseRedundantBsccConstraints() const;
     void setUseRedundantBsccConstraints(bool value);
 
+    bool isApproximationTradeoffSet() const;
+    storm::RationalNumber const& getApproximationTradeoff() const;
+    void setApproximationTradeoff(storm::RationalNumber const& value);
```
sjunges, line 67: Can you write a short comment somehwere what an approximationTradeoff is? What does it mean to be set not set, and what does it mean to be 1/2, say?

src/storm/modelchecker/multiobjective/pcaa/PcaaWeightVectorChecker.cpp
```diff
@@ -55,45 +61,34 @@ boost::optional<typename SparseModelType::ValueType> PcaaWeightVectorChecker<Spa
...
-std::unique_ptr<PcaaWeightVectorChecker<ModelType>> WeightVectorCheckerFactory<ModelType>::create(
+std::unique_ptr<PcaaWeightVectorChecker<ModelType>> createWeightVectorChecker(
     preprocessing::SparseMultiObjectivePreprocessorResult<ModelType> const& preprocessorResult) {
-    if (preprocessorResult.containsOnlyTrivialObjectives()) {
-        return std::make_unique<StandardMdpPcaaWeightVectorChecker<ModelType>>(preprocessorResult);
+    if constexpr (std::is_same_v<ModelType, storm::models::sparse::Mdp<typename ModelType::ValueType>>) {
+        if (preprocessorResult.containsOnlyTrivialObjectives()) {
+            return std::make_unique<StandardMdpPcaaWeightVectorChecker<ModelType>>(preprocessorResult);
+        } else {
+            STORM_LOG_DEBUG("Query contains reward bounded formula");
```
sjunges: This debug message was puzzling me a bit. In particular, I was also considering whether the logical negation of onlyTrivialObjectives should be "has reward-bounded formula". Sounds like something that may easily be violated in the future.

src/storm/modelchecker/multiobjective/pcaa/PcaaWeightVectorChecker.cpp
```diff
@@ -55,45 +61,34 @@ boost::optional<typename SparseModelType::ValueType> PcaaWeightVectorChecker<Spa
...
+            return std::make_unique<RewardBoundedMdpPcaaWeightVectorChecker<ModelType>>(preprocessorResult);
+        }
     } else {
-        STORM_LOG_DEBUG("Query contains reward bounded formula");
-        return std::make_unique<RewardBoundedMdpPcaaWeightVectorChecker<ModelType>>(preprocessorResult);
+        static_assert(std::is_same_v<ModelType, storm::models::sparse::MarkovAutomaton<typename ModelType::ValueType>>);
+        return std::make_unique<StandardMaPcaaWeightVectorChecker<ModelType>>(preprocessorResult);
```
sjunges, line 75: Should you here assume/assert that there is onlyTrivialObjectives??

src/storm/modelchecker/multiobjective/pcaa/PcaaWeightVectorChecker.h
```diff
@@ -37,37 +37,65 @@ class PcaaWeightVectorChecker {
...
+     * Solves the Weighted Sum Optimization Problem for the given weight vector.
+     * setWeightedPrecision(..) can be used to control the accuracy of the results.
+     * After calling this, getAchievablePoint(), getOptimalWeightedSum(), and computeScheduler() can be invoked to retrieve the result of this call.
+     * @note before calling this for the first time, a weighted precision needs to be set. Otherwise, an exception is thrown.
+     * @note Minimizing objectives (Pmin=? [...], Rmin=? [...]) are *implicitly* negated, i.e.,
+     *       we optimize sum_i p_i * (isMinimizing[i] ? -weightVector[i] : weightVector[i]), where p_i is the value of objective i
+     *       For instance, a uniform weight vector like weightVector = (1, 1, ..., 1) will give equal weight to all (minimizing and maximizing) objectives.
+     * @param env
+     * @param weightVector
+     */
```
sjunges, line 50: Are there any assumptions on weightvector?

src/storm/modelchecker/multiobjective/pcaa/PcaaWeightVectorChecker.h
```diff
@@ -37,37 +37,65 @@ class PcaaWeightVectorChecker {
...
 
     /*!
-     * Retrieves a scheduler that induces the current values (if such a scheduler was generated).
-     * Note that check(..) has to be called before retrieving the scheduler. Otherwise, an exception is thrown.
+     * Returns whether achieving precise values (i.e. low weighted precisions) might be challenging due to the type of objectives.
+     * Currently, only time-bounded Markov automaton objectives are considered challenging.
+     * This information is used to guide approximation heuristics, e.g. the choice of the approximation tradeoff in Pareto curve exploration.
      */
-    virtual storm::storage::Scheduler<ValueType> computeScheduler() const;
+    virtual bool smallPrecisionsAreChallenging() const;
```
sjunges, line 98: Is the convergence rate the challenge? Or the fact that we cannot compute *the* weighted value?

src/storm/modelchecker/multiobjective/pcaa/PcaaWeightVectorChecker.h
```diff
@@ -80,23 +108,15 @@ class PcaaWeightVectorChecker {
 
     // The (preprocessed) objectives
     std::vector<Objective<ValueType>> objectives;
+
+   private:
     // The precision of this weight vector checker.
-    ValueType weightedPrecision;
+    std::optional<ValueType> weightedPrecision;
```
sjunges, line 114: Is this optional if you write elsewhere that it has to be set?

src/storm/modelchecker/multiobjective/pcaa/RewardBoundedMdpPcaaWeightVectorChecker.cpp
```diff
@@ -362,16 +375,16 @@ void RewardBoundedMdpPcaaWeightVectorChecker<SparseMdpModelType>::updateCachedDa
 
 template<class SparseMdpModelType>
 std::vector<typename RewardBoundedMdpPcaaWeightVectorChecker<SparseMdpModelType>::ValueType>
-RewardBoundedMdpPcaaWeightVectorChecker<SparseMdpModelType>::getUnderApproximationOfInitialStateResults() const {
-    STORM_LOG_THROW(underApproxResult, storm::exceptions::InvalidOperationException, "Tried to retrieve results but check(..) has not been called before.");
-    return underApproxResult.get();
+RewardBoundedMdpPcaaWeightVectorChecker<SparseMdpModelType>::getAchievablePoint() const {
+    STORM_LOG_THROW(achievablePoint, storm::exceptions::InvalidOperationException, "Tried to retrieve results but check(..) has not been called before.");
```
sjunges, line 388: Earlier you use IllegalFunctionCall; what is the difference? Do we need both?

src/storm/modelchecker/multiobjective/pcaa/PcaaWeightVectorChecker.h
tquatmann: The former. Precisions like 1e-6 are practically infeasible for the implemented algorithm for time-bounded Markov automata. On the other hand, 10^-6 would be a good default when computing reachability probabilities via unsound VI.

src/storm/modelchecker/multiobjective/pcaa/PcaaWeightVectorChecker.h
tquatmann: This is only for internal representation. We'd want to distinguish "not set" from any other value.

src/storm/modelchecker/multiobjective/pcaa/PcaaWeightVectorChecker.h
tquatmann: Great point. It should not be the zero vector. The rest is fine (we now normalize it inside the check.)

tquatmann:
Normalising weight vectors inside the WeightVectorChecker revealed some issue that I still have to investigate.

Apparently, `carl::sqrt(RationalNumber ..)` is *very* imprecise, which might be related.

volkm:
Just having a short look, maybe you try what happens if you use [sqrt_safe](https://github.com/moves-rwth/carl-storm/blob/eba0606d945945e0ce9e99d228d38536d8a7e79b/src/carl/numbers/adaption_cln/operations.h#L434) which returns the interval for a sqrt.

tquatmann:
> In the PCAAWeightVectorChecker, computeScheduler seems to return a scheduler on the preprocessed model (after goal state extension?). Is that the intended behavior? How does one get the original scheduler back ? :-)

`src/storm/modelchecker/multiobjective/MultiObjectivePostprocessing.h` has a "transformObjectiveSchedulersToOriginal" for this. It requires `reverseData` which is part of the PreprocessorResult.

tquatmann:
> > > Also based on the python bindings:
> > > The achievable points have this renormalisation, but the weighted sum can be negative if one of the objectives is a minization value. Is this something that can be made more consistent? I would naively expect that the optimal_weighted_sum is positive, if the achievable point contains of only positive points
> > 
> > 
> > This intuition becomes problematic as soon as we mix minimizing and maximizing objectives. Then, Just taking the sum of all (positive) objective values is somewhat meaningless (e.g. points `(0,1)` and (1,0)` with a uniform weight vector would have the same weighted sum even though one of them clearly dominates the other).
> 
> > > Also based on the python bindings:
> > > The achievable points have this renormalisation, but the weighted sum can be negative if one of the objectives is a minization value. Is this something that can be made more consistent? I would naively expect that the optimal_weighted_sum is positive, if the achievable point contains of only positive points
> > 
> > 
> > This intuition becomes problematic as soon as we mix minimizing and maximizing objectives. Then, Just taking the sum of all (positive) objective values is somewhat meaningless (e.g. points `(0,1)` and (1,0)` with a uniform weight vector would have the same weighted sum even though one of them clearly dominates the other).
> 
> Do you want to expose the weighted sum to the outside then or should this value not be exposed at all? Because the information that this value basically requires reversing to have some meaning?

In the Pareto curve approximation algorithm, it is helpful to properly capture approximation errors. 
I don't see an immediate need to expose this in stormpy, though.

PR #869, opened 2026-02-27
Removed unnecessary Z3 CMake defines

volkm:
- The defines `STORM_HAVE_Z3_OPTIMIZE` and `STORM_Z3_API_USES_STANDARD_INTEGERS` are not needed anymore and removed through this PR.
- Removed `STORM_Z3_VERSION` as it was not used. If some 3rdparty library is using it, let me know.
- Bumped the minimal version of z3 to 4.8.12 (the one from Debian 12)
- Also enabled some tests again which were [buggy for Z3 v4.8.8](https://github.com/moves-rwth/storm/commit/f4a3ceb60e4c140eacb2e1bd084d3557b847efb0). They work again for 4.13.3 (the version on Debian 13 and Ubuntu 25.10), so I enabled them again for these versions. I did not check what the earliest version is where it is fixed.

### DIFF
--- a/src/test/storm/modelchecker/multiobjective/SparseMaPcaaMultiObjectiveModelCheckerTest.cpp
+++ b/src/test/storm/modelchecker/multiobjective/SparseMaPcaaMultiObjectiveModelCheckerTest.cpp
@@ -96,10 +96,13 @@ TEST_F(SparseMaPcaaMultiObjectiveModelCheckerTest, server) {
     std::vector<storm::RationalNumber> lb(2, -eps), ub(2, eps);
     auto bloatingBox = storm::storage::geometry::Hyperrectangle<storm::RationalNumber>(lb, ub).asPolytope();
 
-    if (storm::test::z3AtLeastVersion(4, 8, 8)) {
-        // TODO: z3 v4.8.8 is known to be broken here. Check if this is fixed in future versions >4.8.8
+#ifdef STORM_HAVE_Z3
+    if (storm::test::z3AtLeastVersion(4, 8, 8) && !storm::test::z3AtLeastVersion(4, 13, 3)) {
+        // z3 v4.8.8 is known to be broken here. It is working for v4.13.3.
         GTEST_SKIP() << "Test disabled since it triggers a bug in the installed version of z3.";
     }
+#endif
+
     EXPECT_TRUE(
         expectedAchievableValues->minkowskiSum(bloatingBox)
             ->contains(result->asExplicitParetoCurveCheckResult<double>().getUnderApproximation()->convertNumberRepresentation<storm::RationalNumber>()));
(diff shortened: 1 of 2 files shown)


resources/3rdparty/CMakeLists.txt
```diff
@@ -275,14 +275,13 @@ if(NOT STORM_DISABLE_Z3)
...
-            message(STATUS "Storm - Linking with Z3 (version ${Z3_VERSION}). (Z3 version supports optimization)")
+            message(STATUS "Storm - Linking with Z3 (version ${Z3_VERSION}).")
             list(APPEND STORM_DEP_IMP_TARGETS z3::libz3)
         endif()
     else()
-        # Debian and Ubuntu packages are as of Debian 14 and Ubuntu 24.04 very old,
+        # Debian and Ubuntu packages are as of Debian 12 and Ubuntu 24.04 very old,
         # such that no Cmake config files are provided. For the time being, we use a FindZ3.cmake script.
-        find_package(Z3 QUIET 4.8.7)
+        find_package(Z3 QUIET 4.8.12)
```
sjunges, line 284: I think this was after where the config part was introduced, so I think we can now simplify the cmakelist and can assume that we find the z3config (not urgent, but maybe worht trying)

src/test/storm/modelchecker/multiobjective/SparseMaPcaaMultiObjectiveModelCheckerTest.cpp
```diff
@@ -18,7 +17,16 @@
 #include "storm/storage/geometry/Polytope.h"
 #include "storm/storage/jani/Property.h"
 
-TEST(SparseMaPcaaMultiObjectiveModelCheckerTest, serverRationalNumbers) {
+class SparseMaPcaaMultiObjectiveModelCheckerTest : public ::testing::Test {
+   protected:
+    void SetUp() override {
+#ifndef STORM_HAVE_Z3
```
sjunges, line 24: @tquatmann does this require z3 or an exact LP solver? (As we also have Soplex...)

volkm: One thing to keep in mind here is that even if z3 is not required for the solving, it is still needed for parsing the Prism files.

resources/3rdparty/CMakeLists.txt
volkm: Thanks for the tip. I will look into it.

src/test/storm/modelchecker/multiobjective/SparseMaPcaaMultiObjectiveModelCheckerTest.cpp
tquatmann: The multi-objective part currently (and also with #868 merged) requires z3. 
An exact LP solver would actually suffice but it's not implemented that way.
We could add something like `createExact` to the LpSolverFactory but that should imo be a separate PR

resources/3rdparty/CMakeLists.txt
volkm: I tried it and at least on Debian 12, the CONFIG mode does not work yet. So we still need the fallback for now.

PR #875, opened 2026-03-04
import/export of explicit models in (binary) UMB format

tquatmann:
Import and export of [UMB](https://pmc-tools.github.io/umb/spec) models.

- Provides `UmbModel` class that serves as an immediate layer between the raw files and a `storm::models::sparse::Model`.
- Supports almost all of UMB (Valuations are not fully integrated yet)
- Uses `for_each_field` from [Boost.PFR](https://www.boost.org/doc/libs/latest/doc/html/boost_pfr/short_examples_for_the_impatient.html) for cleaner JSON (de-)serialization and to import/export the different `.bin` files from and to the right location.


src/storm/storage/umb/export/ExportOptions.h
```diff
@@ -0,0 +1,40 @@
...
+    /*!
+     * Whether export of choice origins is enabled.
+     * @note If both, choice origins and choice labelings are present and enabled, choice origins will be used.
+     */
+    bool allowChoiceLabelingAsActions{true};
+
+    /*!
+     * The type of compression used for the exported UMB model.
+     */
+    storm::io::CompressionMode compression{storm::io::CompressionMode::Xz};
```
tquatmann: ~~The default compression is still subject to discussion.~~

src/storm/storage/umb/export/SparseModelToUmb.cpp
```diff
@@ -0,0 +1,494 @@
...
+        uint64_t const numActions = choiceOriginsToUmb(*model.getChoiceOrigins(), umbModel);
+        umbModel.index.transitionSystem.numChoiceActions = numActions;
+    } else if (options.allowChoiceLabelingAsActions && model.hasChoiceLabeling()) {
+        uint64_t const numActions = choiceLabelingToUmb(model.getChoiceLabeling(), umbModel);
+        umbModel.index.transitionSystem.numChoiceActions = numActions;
+    }
+
+    // State valuations
+    if (model.hasStateValuations()) {
+        STORM_LOG_THROW(false, storm::exceptions::NotSupportedException, "State valuations are not yet supported for UMB export.");
```
tquatmann, line 422: State valuations are not (fully) supported by this PR yet.

I'd like to replace our existing state valuation representation as it is somewhat incompatible with the UMB state valuations. I fear this will require touching another ~20 files, so maybe we should merge the current PR first.

src/storm/storage/umb/export/ExportOptions.h
tquatmann: Switched to gzip as default (aligning with e.g. PRISM implementation)

tquatmann:
This is now ready for review @sjunges @volkm  🙂

resources/3rdparty/modernjson/include/nlohmann/storm_utility.hpp
```diff
@@ -107,6 +107,11 @@ constexpr ToType convert(FromType const& v)
     {
         return static_cast<ToType>(v);
     }
+	else if constexpr (std::is_same_v<ToType, int>)
+	{
```
sjunges, line 111: Ok. Is this something we should try to get into main modernjson?

src/storm/settings/modules/IOSettings.cpp
```diff
@@ -156,6 +158,10 @@ IOSettings::IOSettings() : ModuleSettings(moduleName) {
                                          .addValidatorString(ArgumentValidatorFactory::createExistingFileValidator())
                                          .build())
                         .build());
+    this->addOption(storm::settings::OptionBuilder(moduleName, explicitUmbOptionName, false, "Parses the model given in the UMB format.")
```
sjunges: Should we mention that this can be compressed/not compressed.

src/storm/storage/expressions/ExpressionManager.h
```diff
@@ -349,6 +355,11 @@ class ExpressionManager : public std::enable_shared_from_this<ExpressionManager>
      */
     uint_fast64_t getNumberOfArrayVariables() const;
 
+    /*!
+     * Retrieves the number of array variables.
```
sjunges: string variables?

src/storm/storage/umb/export/SparseModelToUmb.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <memory>
+#include "storm/models/sparse/Model.h"
```
sjunges: Not sure where this is all included, but maybe model can be forward declared (really depends on whether this header is used a lot)

src/storm/storage/umb/import/SparseModelFromUmb.h
```diff
@@ -0,0 +1,37 @@
+#pragma once
+
+#include <memory>
+#include "storm/models/sparse/Model.h"
```
sjunges: Here, also potential for for forward declaration...

src/storm/storage/umb/model/Type.h
```diff
@@ -0,0 +1,68 @@
...
+ * Returns the default size (in bits) of a type, if available
+ */
+uint64_t defaultBitSize(Type const type);
+
+/*!
+ * Returns a string representation of the type.
+ */
+std::string toString(Type const type);
+
+struct SizedType {
```
sjunges, line 58: Should this be part of umb headers? Is a umb representation something that is intimately bound to this json serialization (maybe it is, after all, it is currently foremost a file format...)

src/storm/storage/umb/Umb.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+#include <filesystem>
+#include <memory>
+
+#include "storm/storage/umb/export/ExportOptions.h"
+#include "storm/storage/umb/import/ImportOptions.h"
+
+#include "storm/models/sparse/Model.h"
```
sjunges: One more potential forward

sjunges:
whoop whoop. LGTM. I didnt go over the complete umb code....

resources/3rdparty/modernjson/include/nlohmann/storm_utility.hpp
tquatmann: The `convert` function is mainly needed to handle conversions from and to `storm::RationalNumber`.

Iirc, we made some more intrusive changes, because a json "float" has to represented as a pointer to a rational number for us.

In any case, we should update our shipped modernjson soon. (currently shipped version is apparently from Nov 2023)

src/storm/storage/umb/model/Type.h
tquatmann: What specifically do you mean by "umb headers"?
Generally, most of the things in storm/storage/umb/model/ are closely tied to the UMB standard (by intention).

On the one side, `SizedType` and `Type` could into `ModelIndex.h` , i.e., the main file describing almost all of the JSON of a UMB model.
However, we also need `Type` and `SizedType` in other places where it would be odd to include `ModelIndex.h` (like in `ValueEncoding.h` where we do low-level bit stuff).

src/storm/storage/umb/import/SparseModelFromUmb.cpp
```diff
@@ -0,0 +1,373 @@
...
+
+    STORM_LOG_ASSERT(useDefault || useDouble || useRational, "Unexpected value type option: " << static_cast<int>(options.valueType) << ".");
+
+    if constexpr (std::is_same_v<ValueType, double>) {
+        return (useDefault && haveDouble) || (useDouble && !haveInterval);
+    } else if constexpr (std::is_same_v<ValueType, storm::RationalNumber>) {
+        return (useDefault && haveRational) || (useRational && !haveInterval);
+    } else {
+        static_assert(std::is_same_v<ValueType, storm::Interval>, "Unhandled value type");
+        // Rational intervals currently not supported.
```
lukovdm, line 343: I currently have #885 working which adds rational intervals. Would it be an idea to implement rational intervals at some point?

src/storm/storage/umb/export/SparseModelToUmb.h
tquatmann: Its not included super often, but still sounds like a good idea. The new `ModelForward.h` could probably be used at other places, too.

src/storm/storage/umb/import/SparseModelFromUmb.cpp
tquatmann: Yes, the UMB code is kind of ready for that. Once both PRs are merged, we can probably add rational intervals easily.

src/storm/storage/umb/model/Type.h
sjunges: My point was mostly whether it makes sense to separate umb from the json representation, but I guess it doesnt make sense, currently).

sjunges:
Merge?! :-)

PR #876, opened 2026-03-05
Revised check for probabilistic transition matrix

tquatmann:
Intervals like [0,0.4] should not be disallowed on a global level.

### DIFF
--- a/src/storm/models/sparse/Model.cpp
+++ b/src/storm/models/sparse/Model.cpp
@@ -66,9 +66,6 @@ void Model<ValueType, RewardModelType>::assertValidityOfComponents(
     STORM_LOG_ASSERT(
         components.rateTransitions || this->hasParameters() || this->hasUncertainty() || this->getTransitionMatrix().isProbabilistic(stochasticTolerance),
         "The matrix is not probabilistic.");
-    if (this->hasUncertainty()) {
-        STORM_LOG_ASSERT(this->getTransitionMatrix().hasOnlyPositiveEntries(), "Not all entries are (strictly) positive.");
-    }
     STORM_LOG_THROW(this->getStateLabeling().getNumberOfItems() == stateCount, storm::exceptions::IllegalArgumentException,
                     "Invalid item count (" << this->getStateLabeling().getNumberOfItems() << ") of state labeling (states: " << stateCount << ").");
     for (auto const& rewardModel : this->getRewardModels()) {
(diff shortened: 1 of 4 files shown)


src/storm/models/sparse/Model.cpp
```diff
@@ -63,11 +63,11 @@ void Model<ValueType, RewardModelType>::assertValidityOfComponents(
...
                     "Invalid column count of transition matrix.");
-    STORM_LOG_ASSERT(
-        components.rateTransitions || this->hasParameters() || this->hasUncertainty() || this->getTransitionMatrix().isProbabilistic(stochasticTolerance),
-        "The matrix is not probabilistic.");
-    if (this->hasUncertainty()) {
-        STORM_LOG_ASSERT(this->getTransitionMatrix().hasOnlyPositiveEntries(), "Not all entries are (strictly) positive.");
+    {
+        std::string reasonNotProbabilistic [[maybe_unused]];
+        STORM_LOG_ASSERT(
+            components.rateTransitions || this->hasParameters() || this->getTransitionMatrix().isProbabilistic(stochasticTolerance, reasonNotProbabilistic),
```
volkm:
```suggestion
            components.rateTransitions || this->getTransitionMatrix().isProbabilistic(stochasticTolerance, reasonNotProbabilistic),
```

src/storm/storage/SparseMatrix.cpp
```diff
@@ -2251,19 +2251,64 @@ typename SparseMatrix<ValueType>::index_type SparseMatrix<ValueType>::getNoncons
...
+            return value;
+        } else {
+            return storm::utility::convertNumber<BaseType>(value);
+        }
+    };
+    STORM_LOG_ASSERT(storm::utility::isConstant(tolerance), "Expected constant tolerance. Got " << tolerance);
+    BaseType const zeroMinusTolerance = storm::utility::zero<BaseType>() - toBaseType(tolerance);
+    BaseType const onePlusTolerance = storm::utility::one<BaseType>() + toBaseType(tolerance);
+    BaseType const oneMinusTolerance = storm::utility::one<BaseType>() - toBaseType(tolerance);
+
```
volkm, line 2268: Do we need a tolerance for rational functions or rational numbers?

volkm (approved):
LGTM. Thanks for the changes. Just two comments.

src/storm/storage/SparseMatrix.cpp
tquatmann: The hypothetical use case I have in mind is that we have a model given in `double` precision and convert that to an exact representation.
Right now, the tolerance would just be set to zero.

volkm: Okay, sounds reasonable.

PR #878, opened 2026-03-08
added state-valuation transformer with test

sjunges:
Necessary to build state valuations with extra variables (for tutorial).

The code is not used right now, but will be used from stormpy. I think that the code is general enough that it should be part of storm and not of stormpy.

### DIFF
--- a/src/storm/storage/sparse/StateValuationTransformer.h
+++ b/src/storm/storage/sparse/StateValuationTransformer.h
@@ -10,8 +10,8 @@ namespace storm::storage::sparse {
  * The freshly introduced variables may either replace or extend the existing variable set.
  */
 class StateValuationTransformer {
-    public:
-     StateValuationTransformer(StateValuations const& oldStateValuations);
+   public:
+    StateValuationTransformer(StateValuations const& oldStateValuations);
     /*!
      * Add a Boolean variable defined by the given expression. Note that these should all be over the same expression manager.
      * @param var A variable with type Bool


src/storm/storage/sparse/StateValuations.cpp
```diff
@@ -319,7 +319,7 @@ bool StateValuations::assertValuation(StateValuation const& valuation) const {
             }
             bv->set(varIndex.second, true);
         } else {
-            STORM_LOG_ERROR("Valuation does not provide a value for all variables");
+            STORM_LOG_ERROR("Valuation does not provide a value for for variable: " << varIndex.first.getName());
```
tquatmann:
```suggestion
            STORM_LOG_ERROR("Valuation does not provide a value for variable: " << varIndex.first.getName());
```

src/storm/storage/sparse/StateValuationTransformer.cpp
```diff
@@ -0,0 +1,58 @@
...
+                    }
+                } else {
+                    STORM_LOG_ASSERT(sv.isRational(), "Must be RationalVariable");
+                    evaluator.setRationalValue(var, sv.getRationalValue());
+                    if (extend) {
+                        STORM_LOG_ASSERT(false, "Currently not supported.");
+                    }
+                }
+            }
+            // Label assignments can be safely skipped.
```
tquatmann: Why? It seems to me that they are unsupported, i.e., we would have to assert something like `!extend` in the `else` case of the if assignment from line 29.

src/storm/storage/sparse/StateValuationTransformer.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+#include "storm/storage/expressions/Expression.h"
+#include "storm/storage/sparse/StateValuations.h"
+
+namespace storm::storage::sparse {
+
```
tquatmann, line 6: I'd suggest to add a little more explanation what this actually does. Something like

```suggestion

/*!
 * Transforms the given state valuations to a new state valuations over a new variable set.
 * The values of the new variables are determined by evaluating the provided expressions w.r.t. the old variable valuation.
 * The freshly introduced variables may either replace or extend the existing variable set.
```

src/storm/storage/sparse/StateValuationTransformer.cpp
```diff
@@ -0,0 +1,58 @@
+#include "storm/storage/sparse/StateValuationTransformer.h"
+#include "storm/adapters/RationalNumberAdapter.h"
+#include "storm/storage/expressions/ExpressionEvaluator.h"
+#include "storm/utility/constants.h"
+
+namespace storm::storage::sparse {
+
+void StateValuationTransform::addBooleanExpression(storm::expressions::Variable const& var, storm::expressions::Expression const& expr) {
+    booleanVariables.push_back(var);
```
tquatmann, line 20: Might want to assert that `var`, `expr` and `oldValuations` all use the same expression manager.

src/storm/storage/sparse/StateValuationTransformer.cpp
```diff
@@ -0,0 +1,58 @@
...
+                    evaluator.setIntegerValue(var, sv.getIntegerValue());
+                    if (extend) {
+                        integerValues.push_back(sv.getIntegerValue());
+                    }
+                } else {
+                    STORM_LOG_ASSERT(sv.isRational(), "Must be RationalVariable");
+                    evaluator.setRationalValue(var, sv.getRationalValue());
+                    if (extend) {
+                        STORM_LOG_ASSERT(false, "Currently not supported.");
+                    }
```
tquatmann: This will fail in debug mode but ignore such variables in release mode. I'll be in favour of aligning the behaviours because it might be a limitation that is easy to miss (in particular from stormpy perspective)
```suggestion
                    STORM_LOG_THROW(!extend, storm::exceptions::NotSupportedException, "Extending state valuations with rational values is currently not supported.");
```

src/storm/storage/sparse/StateValuationTransformer.h
```diff
@@ -0,0 +1,41 @@
...
+/*!
+ * Transforms the given state valuations to a new state valuations over a new variable set.
+ * The values of the new variables are determined by evaluating the provided expressions w.r.t. the old variable valuation.
+ * The freshly introduced variables may either replace or extend the existing variable set.
+ */
+class StateValuationTransform {
+    // TODO: Also support adding integer variables.
```
tquatmann: This is now obsolete, right?
```suggestion
```

src/test/storm/storage/StateValuationTest.cpp
```diff
@@ -0,0 +1,80 @@
...
+    val.operator++();
+    ASSERT_EQ(val.getName(), "alwaysFalse");
+    val.operator++();
+    ASSERT_EQ(val.getName(), "s");
+    val.operator++();
+    ASSERT_EQ(val.getName(), "d");
+    val.operator++();
+    ASSERT_TRUE(val == newsv.at(0).end());
+    for (uint64_t state = 0; state < newsv.getNumberOfStates(); ++state) {
+        std::cout << newsv.getStateInfo(state) << "\n";
```
tquatmann:
```suggestion
```

PR #883, opened 2026-03-09
Add link options for AddressSanitizer

sjunges:
Necessary on my mac (seems a general thing though).


sjunges:
CI Is not actually testing ASAN currently, is it? (I think we still have memleaks etc)

volkm:
You are correct, ASAN is currently disabled. So the CI should be successful.

PR #886, opened 2026-03-10
Monitor verification

lukovdm:
Add ATVA25 and AAMAS 26 monitor verification code.

PR #887, opened 2026-03-10
Conditional model checking

lukovdm:


lukovdm:
This PR builds on top of #886.

lukovdm:
The failing tests have to deal with some irritating floating point errors leading to incorrect values for pMin and pMax in bisection advanced. The failing test specifically has a pMax that is about 10^-9 below the true value. This is also outside the standard 10^-12 bound we keep for floating point comparisons. The result of the floating point error is a lower bound > 1 by about 10^-6 (i.e. 1.0000001). I am not entirely sure how to tackle this. Using the conditional tolerance does not seem right.

sjunges:
Without sound methods in the backend, this can indeed happen. I would mark the tests as skip/expect fail maybe. We should probably open an issue to ensure that we have the ability to run this with a sound backend and make sure that we test the logic with exact arithmetic as well.

src/storm/environment/modelchecker/ModelCheckerEnvironment.cpp
```diff
@@ -43,6 +45,22 @@ void ModelCheckerEnvironment::setConditionalAlgorithmSetting(ConditionalAlgorith
...
 
+void ModelCheckerEnvironment::setConditionalTolerance(storm::RationalNumber const& value) {
+    conditionalToleranceSetting = value;
+}
+
+storm::RationalNumber ModelCheckerEnvironment::getConditionalTolerance() const {
+    return conditionalToleranceSetting;
+}
+
+bool ModelCheckerEnvironment::isAllowOptimizationForBoundedPropertiesSet() const {
```
sjunges: What does this flag do?

src/storm/environment/modelchecker/ModelCheckerEnvironment.h
```diff
@@ -28,6 +29,13 @@ class ModelCheckerEnvironment {
     ConditionalAlgorithmSetting getConditionalAlgorithmSetting() const;
     void setConditionalAlgorithmSetting(ConditionalAlgorithmSetting value);
 
+    // TODO: This should probably be moved.
```
sjunges: We could open a subenvironment for conditional properties? That would probably group things nicer together.

src/storm/modelchecker/helper/conditional/ConditionalHelper.cpp
```diff
@@ -701,20 +1259,23 @@ std::optional<SolutionType> handleTrivialCases(uint64_t const initialState, Norm
...
-        auto const normalFormPrecisionFactor = storm::utility::convertNumber<storm::RationalNumber, std::string>("1/10");
-        normalFormConstructionEnv.solver().minMax().setPrecision(env.solver().minMax().getPrecision() * normalFormPrecisionFactor);
-        analysisEnv.solver().minMax().setPrecision(env.solver().minMax().getPrecision() *
-                                                   (storm::utility::one<storm::RationalNumber>() - normalFormPrecisionFactor));
-    }
+
+    //    if (env.solver().isForceSoundness()) {
+    //        // We intuitively have to divide the precision into two parts, one for computations when constructing the normal form and one for the actual
+    //        analysis.
+    //        // As the former is usually less numerically challenging, we use a factor of 1/10 for the normal form construction and 9/10 for the analysis.
```
sjunges: This commented out code shoudl be checked/removed/etc

src/storm/modelchecker/helper/conditional/ConditionalHelper.cpp
```diff
@@ -725,58 +1286,186 @@ std::unique_ptr<CheckResult> computeConditionalProbabilities(Environment const&
...
+
         // sw.stop();
         // STORM_PRINT_AND_LOG("Time for analyzing the normal form: " << sw << ".\n");
     }
-    return std::unique_ptr<CheckResult>(new ExplicitQuantitativeCheckResult<ValueType>(initialState, initialStateValue));
+    std::unique_ptr<CheckResult> result(new ExplicitQuantitativeCheckResult<SolutionType>(initialState, initialStateValue));
+
+    // if produce schedulers was set, we have to construct a scheduler with memory
+    if (checkTask.isProduceSchedulersSet() && scheduler) {
+        // not sure about this
```
sjunges: double check this

src/storm/modelchecker/helper/conditional/ConditionalHelper.cpp
```diff
@@ -725,58 +1286,186 @@ std::unique_ptr<CheckResult> computeConditionalProbabilities(Environment const&
...
+                }
+            } else {
+                finalScheduler->setChoice(scheduler->getChoice(state), state, 0);
+            }
+
+            // set choices for memory 1, these are the choices after condition was reached
+            if (normalFormData.schedulerChoicesForReachingTargetStates->isChoiceSelected(state)) {
+                finalScheduler->setChoice(normalFormData.schedulerChoicesForReachingTargetStates->getChoice(state), state, 1);
+            } else {
+                finalScheduler->setChoice(0, state, 1);  // arbitrary choice if no choice was recorded, TODO: this could be problematic for paynt?
```
sjunges: Figure out this TODO and track in an issue or remove.

sjunges: Also in other places.

src/storm/utility/constants.cpp
```diff
@@ -37,6 +37,11 @@ bool isOne(ValueType const& a) {
 
 template<typename ValueType>
 bool isZero(ValueType const& a) {
+    static_assert(
```
sjunges: This assertion is a bit strange?

src/test/storm/modelchecker/prctl/mdp/ConditionalMdpPrctlModelCheckerTest.cpp
```diff
@@ -94,6 +119,32 @@ class SparseRationalNumberBisectionAdvancedEnvironment {
...
+
+class SparseRationalNumberBisectionAdvancedPtEnvironment {
+   public:
+    static const bool isExact = true;
+    typedef storm::RationalNumber ValueType;
+    typedef storm::models::sparse::Mdp<ValueType> ModelType;
+    static storm::Environment createEnvironment() {
+        storm::Environment env;
+        env.modelchecker().setConditionalAlgorithmSetting(storm::ConditionalAlgorithmSetting::BisectionAdvancedPolicyTracking);
+        env.solver().minMax().setPrecision(storm::utility::convertNumber<storm::RationalNumber>(0));  // TODO: this should not be necessary
```
sjunges: Another TODO that we should fix (there is also storm::utility::zero<> if that helps?)

src/storm/modelchecker/prctl/helper/SparseMdpPrctlHelper.cpp
```diff
@@ -1476,6 +1475,179 @@ MDPSparseModelCheckingHelperReturnType<SolutionType> SparseMdpPrctlHelper<ValueT
     }
 }
 
+template<typename ValueType, typename SolutionType>
+std::unique_ptr<CheckResult> SparseMdpPrctlHelper<ValueType, SolutionType>::computeConditionalProbabilities(
```
tquatmann: If I see this correctly, this re-introduces the old conditional probability implementation. Probably not intended?

src/storm/modelchecker/helper/conditional/ConditionalHelper.h
```diff
@@ -22,6 +24,7 @@ class BackwardTransitionCache;
 
 template<typename ValueType, typename SolutionType = ValueType>
 std::unique_ptr<CheckResult> computeConditionalProbabilities(Environment const& env, storm::solver::SolveGoal<ValueType, SolutionType>&& goal,
+                                                             storm::modelchecker::CheckTask<storm::logic::ConditionalFormula, SolutionType> const& checkTask,
```
tquatmann: Why specifically do we need to pass the `checkTask` here?

I'd be in favour of keeping the Checktask inside the `ModelChecker` classes because it's such a general purpose structure.
If it is only for the `TopLevelTolerance`, we could just pass that.

src/storm/modelchecker/CheckTask.h
```diff
@@ -321,6 +322,20 @@ class CheckTask {
...
+     */
+    ValueType const& getTopLevelTolerance() const {
+        return *TopLevelTolerance;
+    }
+
+    /*!
+     * Sets the tolerance for the top-level model checking.
+     */
+    void setTopLevelTolerance(ValueType topLevelTolerance) {
+        this->TopLevelTolerance = std::make_shared<ValueType>(std::move(topLevelTolerance));
```
tquatmann: This seems problematic.

The top level tolerance that we introduce here would (for now) only apply to conditional probabilities. All other properties take their tolerances from the `storm::Environment`.

Can't we also use the environment for that?

src/storm/modelchecker/CheckTask.h
```diff
@@ -391,6 +410,9 @@ class CheckTask {
 
     // Whether uncertainty should be resolved be minimizing, maximizing, acting robust or cooperative.
     UncertaintyResolutionMode uncertaintyResolutionMode;
+
+    // The tolerance for the top-level model checking.
+    std::shared_ptr<ValueType> TopLevelTolerance;
```
tquatmann: Why a shared_ptr ?

src/storm/settings/modules/MinMaxEquationSolverSettings.cpp
```diff
@@ -54,7 +54,7 @@ MinMaxEquationSolverSettings::MinMaxEquationSolverSettings() : ModuleSettings(mo
                         .setIsAdvanced()
                         .addArgument(storm::settings::ArgumentBuilder::createDoubleArgument("value", "The precision to achieve.")
                                          .setDefaultValueDouble(1e-06)
-                                         .addValidatorDouble(ArgumentValidatorFactory::createDoubleRangeValidatorExcluding(0.0, 1.0))
+                                         .addValidatorDouble(ArgumentValidatorFactory::createDoubleRangeValidatorIncluding(0.0, 1.0))
```
tquatmann: Is this necessary? I'm not sure if there are other parts in the code that assume `precision > 0`.

src/storm/settings/modules/ModelCheckerSettings.cpp
```diff
@@ -5,6 +5,7 @@
 #include "storm/settings/Option.h"
 #include "storm/settings/OptionBuilder.h"
 #include "storm/settings/SettingsManager.h"
+#include "storm/utility/constants.h"
```
tquatmann, line 8: Why does this include become necessary?

src/storm/settings/modules/ModelCheckerSettings.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "storm-config.h"
+#include "storm/adapters/RationalNumberAdapter.h"
```
tquatmann, line 4: Necessary?

src/storm/utility/constants.h
```diff
@@ -76,6 +76,9 @@ ValueType one();
 template<typename ValueType>
 ValueType zero();
 
+template<typename ValueType>
+ValueType defaultTolerance();
```
tquatmann: I'm not entirely sure where such a default tolerance would apply? Can you elaborate?

src/test/storm/modelchecker/prctl/mdp/ConditionalMdpPrctlModelCheckerTest.cpp
```diff
@@ -93,7 +119,33 @@ class SparseRationalNumberBisectionAdvancedEnvironment {
     typedef storm::models::sparse::Mdp<ValueType> ModelType;
     static storm::Environment createEnvironment() {
         storm::Environment env;
-        env.modelchecker().setConditionalAlgorithmSetting(storm::ConditionalAlgorithmSetting::BisectionAdvanced);
+        env.modelchecker().conditional().setAlgorithm(storm::ConditionalAlgorithmSetting::BisectionAdvanced);
+        env.solver().minMax().setPrecision(storm::utility::zero<storm::RationalNumber>());  // TODO: this should not be necessary
```
tquatmann: Why is it necessary?

src/storm/modelchecker/helper/conditional/ConditionalHelper.cpp
```diff
@@ -511,80 +888,158 @@ class WeightedReachabilityHelper {
...
-                                 NormalFormData<ValueType> const& normalForm) {
+typename internal::ResultReturnType<ValueType> computeViaBisection(Environment const& env, bool const useAdvancedBounds, bool const usePolicyTracking,
+                                                                   SolutionType const precision, uint64_t const initialState,
+                                                                   storm::solver::SolveGoal<ValueType, SolutionType> const& goal, bool computeScheduler,
+                                                                   storm::storage::SparseMatrix<ValueType> const& transitionMatrix,
+                                                                   storm::storage::SparseMatrix<ValueType> const& backwardTransitions,
+                                                                   NormalFormData<ValueType> const& normalForm) {
     // We currently handle sound model checking incorrectly: we would need the actual lower/upper bounds of the weightedReachabilityHelper
-    STORM_LOG_WARN_COND(!env.solver().isForceSoundness(),
+    STORM_LOG_WARN_COND(!(env.solver().isForceSoundness() && storm::utility::isZero(precision)),
```
tquatmann: Why only issue this warning if precision is zero?

tquatmann (commented):
Thank you for the PR! Great Work :)

I have some of the usual nitpickings.

Most importantly, I'm not a big fan of introducing the top level tolerance in this PR because it would (for now) only apply to conditional probabilities.

Also, `--minmax:precision 0` might have some weird side-effects that we are currently unaware of.

src/test/storm/modelchecker/prctl/mdp/ConditionalMdpPrctlModelCheckerTest.cpp
tquatmann: Imo, setting `--minmax:precision 0` is never reasonable. Either we are in exact mode (in which case the precision is ignored) or we are in inexact mode, in which case this likely leads to non-termination somewhere unexpected.

sjunges:
precision zero should not be necessary anymore if the precision does not control the tolerance... I think tolerance was introduced exactly to also allow non-zero precision in the exact case.

tquatmann:
Ahh, I see. Exact arithmetics with non-exact solutions is something that we can't really specify right now.

Imo, the current solution is a bit unclear (what is the difference between tolerance vs. precision?). But it's also not obvious to me how to solve this. Do we have to introduce support for that in this PR?

src/storm/settings/modules/ModelCheckerSettings.cpp
lukovdm: These were left over after removing conditional tolerance. But since they are now back this is necessary again.

src/storm/settings/modules/ModelCheckerSettings.h
lukovdm: Same here.

volkm: (Without having any knowledge of the rest) Is it possible to use RationalNumberForward here? I think this include spreads through some parts of the code base and it would be best if we could avoid including the header here.

src/storm/settings/modules/ModelCheckerSettings.cpp
```diff
@@ -26,14 +28,22 @@ ModelCheckerSettings::ModelCheckerSettings() : ModuleSettings(moduleName) {
...
-    std::vector<std::string> const conditionalAlgs = {"default", "restart", "bisection", "bisection-advanced", "pi"};
+    std::vector<std::string> const conditionalAlgs = {"default", "restart", "bisection", "bisection-advanced", "bisection-pt", "bisection-advanced-pt", "pi"};
     this->addOption(storm::settings::OptionBuilder(moduleName, conditionalAlgorithmOptionName, false, "The used algorithm for conditional probabilities.")
-                        .setIsAdvanced()
                         .addArgument(storm::settings::ArgumentBuilder::createStringArgument("name", "The name of the method to use.")
                                          .addValidatorString(ArgumentValidatorFactory::createMultipleChoiceValidator(conditionalAlgs))
                                          .setDefaultValueString("default")
                                          .build())
                         .build());
+    // Would be better if there was a createRationalArgument .
```
tquatmann:
```suggestion
```
I agree, but this should not be a comment imo :)

src/storm/modelchecker/helper/conditional/ConditionalHelper.cpp
```diff
@@ -598,52 +1053,158 @@ SolutionType computeViaBisection(Environment const& env, BisectionMethodBounds b
...
+            STORM_LOG_INFO("Bisection method determined result after " << iterationCount << " iterations. Found bounds are ["
+                                                                       << storm::utility::convertNumber<double>(*lowerBound) << ", "
+                                                                       << storm::utility::convertNumber<double>(*upperBound) << "], threshold is "
+                                                                       << storm::utility::convertNumber<double>(goal.thresholdValue()) << ".");
+            break;
         }
         // check for early termination
         if (storm::utility::resources::isTerminate()) {
-            STORM_LOG_WARN("Bisection solver aborted after " << iterationCount << "iterations. Bound difference is "
+            STORM_LOG_INFO("Bisection solver aborted after " << iterationCount << "iterations. Bound difference is "
```
tquatmann:
```suggestion
            STORM_LOG_WARN("Bisection solver aborted after " << iterationCount << "iterations. Bound difference is "
```

This is printed when Storm receives a termination signal (e.g. ctrl+c). Imo a warning is justified then.

src/storm/modelchecker/helper/conditional/ConditionalHelper.cpp
```diff
@@ -701,20 +1262,12 @@ std::optional<SolutionType> handleTrivialCases(uint64_t const initialState, Norm
...
     auto normalFormConstructionEnv = env;
     auto analysisEnv = env;
-    if (env.solver().isForceSoundness()) {
-        // We intuitively have to divide the precision into two parts, one for computations when constructing the normal form and one for the actual analysis.
-        // As the former is usually less numerically challenging, we use a factor of 1/10 for the normal form construction and 9/10 for the analysis.
-        auto const normalFormPrecisionFactor = storm::utility::convertNumber<storm::RationalNumber, std::string>("1/10");
-        normalFormConstructionEnv.solver().minMax().setPrecision(env.solver().minMax().getPrecision() * normalFormPrecisionFactor);
-        analysisEnv.solver().minMax().setPrecision(env.solver().minMax().getPrecision() *
-                                                   (storm::utility::one<storm::RationalNumber>() - normalFormPrecisionFactor));
-    }
```
tquatmann, line 717: Why is this removed?
Shouldn't we also set the minmax precision in the `analysisEnv` from the conditional tolerance? (regardless of soundness settings)

src/storm/modelchecker/helper/conditional/ConditionalHelper.cpp
```diff
@@ -724,56 +1277,175 @@ std::unique_ptr<CheckResult> computeConditionalProbabilities(Environment const&
...
+            scheduler =
+                std::unique_ptr<storm::storage::Scheduler<SolutionType>>(new storm::storage::Scheduler<SolutionType>(transitionMatrix.getRowGroupCount()));
+        }
         STORM_LOG_DEBUG("Initial state has trivial value " << initialStateValue);
     } else {
         STORM_LOG_ASSERT(normalFormData.maybeStates.get(initialState), "Initial state must be a maybe state if it is not a terminal state");
-        auto alg = analysisEnv.modelchecker().getConditionalAlgorithmSetting();
+        auto alg = analysisEnv.modelchecker().conditional().getAlgorithm();
         if (alg == ConditionalAlgorithmSetting::Default) {
             alg = ConditionalAlgorithmSetting::Restart;
```
tquatmann: I suggest to change that default?

tquatmann (commented):
I believe the precision configuration should be streamlined with other algorithms. See `LongRunAverageSolver[Environment|Settings.h` for example.

- For the sake of consistency, should we call it "precision" and not "tolerance"?
- The ConditionalModelCheckerEnvironment needs a `bool relative` flag. Right now this is apparently always read from `minmax` settings.
- The tolerance/precision setting in the ConditionalModelCheckerEnvironment should also apply for the restart method and for the (unconditional) reachability analysis used to obtain the Normal form (we can configure the `analysisEnv` in the `ConditionalHelper` accordingly
- Conditional *DTMC* model checking should now also read the precisions from the ConditionalModelCheckerEnvironment now.
- In the CLI, we can currently set e.g. `--precision 1e-10` which then applies to all kinds of precisions. Setting `--exact` (without specifying a precision) always yields the exact rational number result. I'm very much in favour of keeping it this way with conditionals. This would require a new settings module, though. See e.g. `LongRunAverageSolverSettings.h` for example. To handle cases where we want to have an approximate solution with exact arithmetics, I suggest something like `--exact --conditional:precision 1e-4`. In settings modules, we have been using `bool is[...]SetFromDefaultValue()` functions to implement this kind of behaviour.

Once that is clarified, we're ready to merge (as far as I'm concerned).

src/storm/settings/modules/ModelCheckerSettings.h
tquatmann: For other settings, we just use `double` here and convert to rational in the `Environment`. I don't mind having rationals here, though.

lukovdm: I went for double as there is no good rational argument yet as far as I could find.

lukovdm:
I only have not found a solution to let DTMC conditional queries use the conditional precision. This is because many different precisions could be used depending on the linear equation solver type. It is not clear to me how I could set the precision used for the conditional queries. (Looking at the tests they set many different precisions for different possible environments)

src/storm/settings/modules/ConditionalSettings.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "storm/settings/Option.h"
+#include "storm/settings/OptionBuilder.h"
+#include "storm/settings/SettingsManager.h"
+
+namespace storm::settings::modules {
+
+const std::string ConditionalSettings::moduleName = "conditional";
+const std::string ConditionalSettings::conditionalAlgorithmOptionName = "algorithm";
+const std::string ConditionalSettings::conditionalPrecisionOptionName = "precision";
+const std::string ConditionalSettings::conditionalPrecisionRelativeOptionName = "relative";
```
tquatmann: In other cases we allow to set `--absolute` (and have relative as default).

tquatmann:
> I only have not found a solution to let DTMC conditional queries use the conditional precision. This is because many different precisions could be used depending on the linear equation solver type. It is not clear to me how I could set the precision used for the conditional queries. (Looking at the tests they set many different precisions for different possible environments)

You can use `setLinearEquationSolverPrecision` in `storm::environment::SolverEnvironment` which will do this automatically.
This is not critical, though.

tquatmann (approved):
LGTM!

PR #892, opened 2026-03-19
Add ccache support to Dockerfiles and build workflow

lukovdm:
The PR CI actions where slowing me down quite a bit thus I investigated a bit into how it could be sped up. I added caching of the homebrew downloads and added ccache in all builds with a github cached ccache.


lukovdm:
Thanks for the questions.
- The includes are necessary I think because https://github.com/lukovdm/storm/blob/fc5220f3e1920dab08030f4d47c00111e1883250/CMakeLists.txt#L315 includes several headers for many files.
- About the PCH, looking at these docs, using PCH and pragma once might be the issue: https://ccache.dev/manual/latest.html#_precompiled_headers. But I did not want to change pragma once in all files. I tested a several things to get PCH and ccache to properly work but could not do it.
- The flag and branch things I can implement.
- I have not noticed anything with the dependencies, but they are the slowest part of a warm cache clean build folder build now.

Also, some GitHub workflows are not properly caching yet, but the once that are, are running in 4-10 minutes.

Dockerfile
```diff
@@ -45,12 +45,29 @@ ARG cmake_args=""
 
 # Build Storm
 #############
+RUN if ! command -v ccache >/dev/null 2>&1; then \
```
sjunges, line 48: Can you write some comments here to clarify what this does (and why) for other users.

sjunges:
Code/build changes LGTM; @volkm can you go over the github action changes?

CMakeLists.txt
```diff
@@ -157,6 +159,8 @@ if (STORM_DEVELOPER)
     set(STORM_DEBUG_SPOT ON)
     set(STORM_DEBUG_SYLVAN ON)
     set(STORM_WARNING_AS_ERROR ON)
+	# Turn off PCH for faster ccache usage on warm caches.
```
volkm, line 162:
```suggestion
	# Turn off PCH for faster ccache usage on warm caches. PCH compiled files cannot be properly cached currently.
```

CMakeLists.txt
```diff
@@ -167,13 +171,23 @@ set(CMAKE_COMPILE_WARNING_AS_ERROR ${STORM_WARNING_AS_ERROR})
...
 	find_program(CCACHE_FOUND ccache)
 	mark_as_advanced(CCACHE_FOUND)
 	if(CCACHE_FOUND)
 		message(STATUS "Storm - Using ccache")
+		set(ENV{CCACHE_SLOPPINESS} "pch_defines,time_macros,include_file_mtime,include_file_ctime")
 		set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ccache)
-		set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK ccache)
+
+		if(STORM_COMPILE_WITH_PCH)
+			message(STATUS "STORM - PCH is enabled which will lead to significantly lower ccache hit rates. Consider disabling PCH if you plan to rebuild storm multiple times.")
```
volkm, line 189:
```suggestion
			message(STATUS "STORM - PCH is enabled which will lead to significantly lower ccache hit rates. Consider disabling PCH if you plan to rebuild Storm multiple times.")
```

CMakeLists.txt
```diff
@@ -167,13 +171,23 @@ set(CMAKE_COMPILE_WARNING_AS_ERROR ${STORM_WARNING_AS_ERROR})
...
+	set(CMAKE_DISABLE_PRECOMPILE_HEADERS ON)
+	message(STATUS "Storm - Disabling precompiled headers because of ccache compatibility.")
+endif()
+
 if(STORM_COMPILE_WITH_CCACHE)
 	find_program(CCACHE_FOUND ccache)
 	mark_as_advanced(CCACHE_FOUND)
 	if(CCACHE_FOUND)
 		message(STATUS "Storm - Using ccache")
+		set(ENV{CCACHE_SLOPPINESS} "pch_defines,time_macros,include_file_mtime,include_file_ctime")
```
volkm, line 185: Can you add an explanations for this selection here?

CMakeLists.txt
```diff
@@ -317,6 +331,12 @@ SET(STORM_PRECOMPILED_HEADERS "<any>" "<map>" "<vector>" "<unordered_set>" "<for
 		"<optional>" "<ostream>" "<istream>" "<list>" "<set>" "<fstream>" "<string>" "<boost/optional.hpp>" "<boost/variant.hpp>"
 		"<boost/container/flat_set.hpp>" "<boost/dynamic_bitset.hpp>" "<boost/range/irange.hpp>")
 
+function(storm_target_precompile_headers target)
```
volkm, line 334: Maybe move to resources/cmake/macros or to src/CMakeLists.txt?

Dockerfile
```diff
@@ -78,6 +95,6 @@ RUN cmake -DCMAKE_BUILD_TYPE=$build_type \
 
 # Build Storm
 # (This can be adapted to only build 'storm' or 'binaries' depending on custom needs)
-RUN make -j $no_threads
+RUN ccache --max-size=3G && ccache --zero-stats && make -j $no_threads && ccache --show-stats --verbose
```
volkm, line 98: Can we make the magic number 3G configurable?
Is showing the ccache stats necessary? Users who simply build Storm probably do not care about it.

Dockerfile
```diff
@@ -45,12 +45,29 @@ ARG cmake_args=""
 
 # Build Storm
 #############
+RUN if ! command -v ccache >/dev/null 2>&1; then \
+        if command -v apt-get >/dev/null 2>&1; then \
+            apt-get update && apt-get install -y --no-install-recommends ccache && rm -rf /var/lib/apt/lists/*; \
+        elif command -v apk >/dev/null 2>&1; then \
```
volkm, line 51: We currently (silently) assume that this Dockerfile only works for Ubuntu/Debian images. We have a separate Dockerfile for archlinux.
There is nothing OS-specific here, so we could indeed use a generic Dockerfile. But then I would propose to already install ccache in the base images for the the different OS.

Dockerfile
volkm: Also can we make the usage of ccache configurable?

Dockerfile
volkm: Why do we actually need a call to ccache? Should `STORM_COMPILE_WITH_CCACHE` not automatically take care of it?

src/storm/utility/SignalHandler.h
```diff
@@ -3,6 +3,8 @@
 #include <cstdint>
 #include <cstdlib>
 
+#include <unistd.h>
+
```
volkm, line 7:
```suggestion
#include <cstdint>
#include <cstdlib>
#include <unistd.h>
```

src/storm/solver/helper/SchedulerTrackingHelper.h
```diff
@@ -2,6 +2,8 @@
 #include <memory>
 #include <vector>
 
+#include <boost/optional.hpp>
+
```
volkm, line 6:
```suggestion
#include <vector>
#include <boost/optional.hpp>
```

resources/3rdparty/include_spot.cmake
```diff
@@ -60,10 +60,13 @@ if(NOT STORM_DISABLE_SPOT)
         set(SPOT_SHIPPED_VERSION 2.14.3)
         set(STORM_SPOT_FLAGS "--disable-python;--enable-shared;--disable-static")
         if (NOT STORM_DEBUG_SPOT)
-                set(STORM_SPOT_FLAGS "${STORM_SPOT_FLAGS};--disable-devel;--disable-debug;--enable-optimizations")
+            set(STORM_SPOT_FLAGS "${STORM_SPOT_FLAGS};--disable-devel;--disable-debug;--enable-optimizations")
         else()
-                message(WARNING "Storm - Building Spot in DEBUG mode.")
-                set(STORM_SPOT_FLAGS "${STORM_SPOT_FLAGS};--enable-devel;--enable-debug;--disable-optimizations")
+            message(WARNING "Storm - Building Spot in DEBUG mode.")
+            set(STORM_SPOT_FLAGS "${STORM_SPOT_FLAGS};--disable-devel;--enable-debug;--disable-optimizations")
```
volkm, line 66: Why is devel disabled in Debug mode?

Dockerfile
```diff
@@ -45,12 +45,29 @@ ARG cmake_args=""
...
+            apk add --no-cache ccache; \
+        elif command -v pacman >/dev/null 2>&1; then \
+            pacman -S --noconfirm ccache; \
+        else \
+            echo "No supported package manager found for ccache installation"; \
+        fi; \
+    fi
+
+ENV CCACHE_DIR=/root/.ccache
```
volkm, line 60: Is there a variable for the dir we could use in case of a user different from root?

.github/workflows/Dockerfile.archlinux
```diff
@@ -67,39 +68,44 @@ RUN pacman -S --noconfirm \
...
+RUN mkdir -p "$CCACHE_DIR" && \
+        if [ -d "/opt/storm/.ci-ccache" ]; then cp -a /opt/storm/.ci-ccache/. "$CCACHE_DIR"/; fi
+
 # Switch to build directory
 RUN mkdir -p /opt/storm/build
 WORKDIR /opt/storm/build
 
 # Configure Storm
 RUN cmake -DCMAKE_BUILD_TYPE=$build_type \
-          -DSTORM_PORTABLE=ON \
```
volkm, line 82: Can we maybe revert this? The indentation was deliberate such that the CMake configs align ;)

.github/workflows/buildtest.yml
```diff
@@ -1,494 +1,101 @@
...
-      - name: Run installed storm
-        run: |
-          docker exec ci bash -c "/usr/local/bin/storm --version"
-      - name: Build starter-project on installed Storm
-        uses: ./.github/actions/starter-project-action
-        with:
-          docker: true
-          stormDir: ""
-          command: "--help" # TODO using --help as running the starter-project requires Z3.
+          - {name: "Debug; All dependencies",          buildType: "Debug",   baseImg: "movesrwth/storm-dependencies:latest",            disable_all: "OFF", ClnExact: "OFF", ClnRatfunc: "OFF"}
```
volkm, line 15: Can you align and indent them again as before?

volkm: Also why are we not using the default values for ClnExact and ClnRatfunc here?

.github/workflows/buildtest.yml
```diff
@@ -1,494 +1,101 @@
...
+          - {name: "Release; All dependencies",        buildType: "Release", baseImg: "movesrwth/storm-dependencies:latest",            disable_all: "OFF", ClnExact: "OFF", ClnRatfunc: "OFF"}
+          - {name: "Release; GMP; GMP; Minimal dependencies", buildType: "Release", baseImg: "movesrwth/storm-basesystem:minimal_dependencies", disable_all: "ON",  ClnExact: "OFF", ClnRatfunc: "OFF"}
+    uses: ./.github/workflows/test-build.yml
+    with:
+      NR_JOBS: "4"
```
volkm, line 20:
```suggestion
      nr_jobs: "4"
```

.github/workflows/buildtest.yml
```diff
@@ -1,494 +1,101 @@
...
-  # GitHub runners currently have 4 cores
-  NR_JOBS: "4"
-
 jobs:
-  # Perform in-depth tests with different configurations
-  indepthTests:
-    name: Indepth Tests (${{ matrix.config.name }}, ${{ matrix.config.buildType }})
-    runs-on: ubuntu-latest
+  buildTests:
+    if: "! contains(github.event.pull_request.labels.*.name, 'ready for review')"
```
volkm, line 10: Can this check be done for the complete workflow?

.github/workflows/buildtest.yml
```diff
@@ -1,494 +1,101 @@
-name: Build Test
-# Builds and tests storm on various platforms
-# also deploys images to DockerHub
+name: CI PR
```
volkm, line 1: Better rename the workflow file now to something like `test-pr`

.github/workflows/ci-weekly.yml
```diff
@@ -0,0 +1,145 @@
...
+  workflow_dispatch:
+
+jobs:
+  buildTests:
+    strategy:
+      fail-fast: false
+      matrix:
+        buildType: ["Debug", "Release"]
+        config:
+          # TODO: Replace movesrwth/storm-basesystem:latest with the correct image for "required dependencies"
```
volkm: storm-basesystem:latest is the correct Docker image for required dependencies.

.github/workflows/ci-weekly.yml
```diff
@@ -0,0 +1,145 @@
...
+          - {name: "GMP; GMP; Required dependencies", baseImg: "movesrwth/storm-basesystem:latest",               disable_all: "OFF", ClnExact: "OFF", ClnRatfunc: "OFF"}
+          - {name: "CLN; GMP; Required dependencies", baseImg: "movesrwth/storm-basesystem:latest",               disable_all: "OFF", ClnExact: "ON",  ClnRatfunc: "OFF"}
+          - {name: "GMP; CLN; Required dependencies", baseImg: "movesrwth/storm-basesystem:latest",               disable_all: "OFF", ClnExact: "OFF", ClnRatfunc: "ON"}
+          - {name: "CLN; CLN; Required dependencies", baseImg: "movesrwth/storm-basesystem:latest",               disable_all: "OFF", ClnExact: "ON",  ClnRatfunc: "ON"}
+          - {name: "All dependencies (disabled)",      baseImg: "movesrwth/storm-dependencies:latest",             disable_all: "ON",  ClnExact: "OFF", ClnRatfunc: "OFF"}
+          - {name: "Minimal dependencies",             baseImg: "movesrwth/storm-basesystem:minimal_dependencies", disable_all: "ON",  ClnExact: "OFF", ClnRatfunc: "OFF"}
+    uses: ./.github/workflows/test-build.yml
+    with:
+      NR_JOBS: "4"
```
volkm, line 25:
```suggestion
      nr_jobs: "4"
```

.github/workflows/buildtest.yml
```diff
@@ -1,494 +1,101 @@
...
+      disable_soplex: ${{ matrix.config.disable_all }}
+      disable_spot: ${{ matrix.config.disable_all }}
+      disable_sylvan: ${{ matrix.config.disable_all }}
+      disable_xerces: ${{ matrix.config.disable_all }}
+      disable_z3: ${{ matrix.config.disable_all }}
+      Developer: ${{ matrix.config.buildType == 'Debug' && 'ON' || 'OFF' }}
+      ClnExact: ${{ matrix.config.ClnExact }}
+      ClnRatfunc: ${{ matrix.config.ClnRatfunc }}
+      AllSanitizers: "OFF"
+      cmakeArgs: "-DSTORM_WARNING_AS_ERROR=ON"
```
volkm, line 39: I think this should be enabled by default for all tests. So set it in `test-build`.

src/storm/CMakeLists.txt
```diff
@@ -29,6 +29,16 @@ target_link_libraries(storm PUBLIC ${STORM_DEP_TARGETS} ${STORM_DEP_IMP_TARGETS}
 list(APPEND STORM_TARGETS storm)
 set(STORM_TARGETS ${STORM_TARGETS} PARENT_SCOPE)
 
+# GCC 15 raises a false positive -Wfree-nonheap-object when compiling SparseDtmcEliminationModelChecker
```
volkm, line 32: Can we maybe disable the compiler flag directly in the file?
```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wno-free-nonheap-object"
....
#pragma GCC diagnostic pop
```
We do similar ignores in other files, for example [here](https://github.com/stormchecker/storm/blob/master/src/storm/storage/dd/DdManager.cpp#L177)

src/storm-parsers/CMakeLists.txt
```diff
@@ -11,14 +11,19 @@ target_sources(storm-parsers
...
+storm_target_precompile_headers(storm-parsers REUSE_FROM storm)
 set_target_properties(storm-parsers PROPERTIES VERSION ${STORM_VERSION} SOVERSION ${STORM_VERSION})
 set_target_properties(storm-parsers PROPERTIES DEFINE_SYMBOL "")  # to avoid problems with pch on linux.
 target_link_libraries(storm-parsers PUBLIC storm)
 
 list(APPEND STORM_TARGETS storm-parsers)
 set(STORM_TARGETS ${STORM_TARGETS} PARENT_SCOPE)
 
+# Boost Spirit's utf8.hpp uses char_traits<ucs4_char> which Apple libc++ (Xcode 26+) deprecated
+if (STORM_COMPILER_APPLECLANG)
```
volkm, line 23: Can we maybe disable the compiler flag directly in the files which include boost, similar to [here](https://github.com/stormchecker/storm/blob/master/src/storm/adapters/SpotAdapter.h#L5)

.github/workflows/ci-weekly.yml
```diff
@@ -0,0 +1,144 @@
...
+      fail-fast: false
+      matrix:
+        buildType: ["Debug", "Release"]
+        config:
+          - {name: "GMP; GMP; Required dependencies", baseImg: "movesrwth/storm-basesystem:latest",               disable_all: "OFF", ClnExact: "OFF", ClnRatfunc: "OFF"}
+          - {name: "CLN; GMP; Required dependencies", baseImg: "movesrwth/storm-basesystem:latest",               disable_all: "OFF", ClnExact: "ON",  ClnRatfunc: "OFF"}
+          - {name: "GMP; CLN; Required dependencies", baseImg: "movesrwth/storm-basesystem:latest",               disable_all: "OFF", ClnExact: "OFF", ClnRatfunc: "ON"}
+          - {name: "CLN; CLN; Required dependencies", baseImg: "movesrwth/storm-basesystem:latest",               disable_all: "OFF", ClnExact: "ON",  ClnRatfunc: "ON"}
+          - {name: "All dependencies (disabled)",      baseImg: "movesrwth/storm-dependencies:latest",             disable_all: "ON",  ClnExact: "OFF", ClnRatfunc: "OFF"}
+          - {name: "Minimal dependencies",             baseImg: "movesrwth/storm-basesystem:minimal_dependencies", disable_all: "ON",  ClnExact: "OFF", ClnRatfunc: "OFF"}
```
volkm, line 22: Align with others

.github/workflows/selfrun.yml
```diff
@@ -0,0 +1,122 @@
...
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    types: [opened, synchronize, reopened, labeled]
+
+jobs:
+  buildTests:
+    if: contains(github.event.pull_request.labels.*.name, 'ready for review') || github.event_name == 'push'
```
volkm, line 14: Will this also trigger for every push on a PR?

.github/workflows/buildtest.yml
volkm: Can we set this for the whole workflow?

.github/workflows/selfrun.yml
```diff
@@ -0,0 +1,122 @@
...
+      fail-fast: true
+      matrix:
+        config:
+          - {name: "Debug; All dependencies",              buildType: "Debug",   baseImg: "movesrwth/storm-dependencies:latest", disable_all: "OFF", ClnExact: "OFF", ClnRatfunc: "OFF"}
+          - {name: "Release; All dependencies",            buildType: "Release", baseImg: "movesrwth/storm-dependencies:latest", disable_all: "OFF", ClnExact: "OFF", ClnRatfunc: "OFF"}
+          - {name: "Release; GMP; GMP; Required dependencies", buildType: "Release", baseImg: "movesrwth/storm-basesystem:latest", disable_all: "OFF", ClnExact: "OFF", ClnRatfunc: "OFF"}
+          - {name: "Release; CLN; CLN; Required dependencies", buildType: "Release", baseImg: "movesrwth/storm-basesystem:latest", disable_all: "OFF", ClnExact: "ON",  ClnRatfunc: "ON"}
+    uses: ./.github/workflows/test-build.yml
+    with:
+      NR_JOBS: "4"
```
volkm, line 25: Can we set this for the whole workflow?

.github/workflows/selfrun.yml
```diff
@@ -0,0 +1,122 @@
...
+    if: contains(github.event.pull_request.labels.*.name, 'ready for review') || github.event_name == 'push'
+    strategy:
+      fail-fast: true
+      matrix:
+        distro: ["debian-12", "debian-13", "ubuntu-24.04"]
+    uses: ./.github/workflows/test-linux.yml
+    with:
+      NR_JOBS: "4"
+      distro: ${{ matrix.distro }}
+      buildType: Release
```
volkm, line 82: From the latest [anecdotal evidence](https://github.com/stormchecker/storm/issues/912) we might want to build in Debug mode instead.

.github/workflows/selfrun.yml
```diff
@@ -0,0 +1,122 @@
...
+      save_cache: false
+      install: true
+
+  macTests:
+    if: contains(github.event.pull_request.labels.*.name, 'ready for review') || github.event_name == 'push'
+    strategy:
+      fail-fast: true
+      matrix:
+        config:
+          - {name: "Mac 14, ARM",   distro: "macos-14",       xcode: "15.4",          buildType: "Release"}
```
volkm, line 94: Also use at least one config in Debug mode here?

.github/workflows/selfrun.yml
```diff
@@ -0,0 +1,122 @@
...
+      name: ${{ matrix.config.name }}
+      distro: ${{ matrix.config.distro }}
+      xcode: ${{ matrix.config.xcode }}
+      buildType: ${{ matrix.config.buildType }}
+      restore_cache: ${{ github.event_name != 'push' }}
+      save_cache: false
+      install: true
+
+  deploy:
+    if: contains(github.event.pull_request.labels.*.name, 'ready for review') || github.event_name == 'push'
```
volkm, line 110: Should not happen for PRs but only for commits to master branch.

.github/workflows/test-build.yml
```diff
@@ -0,0 +1,193 @@
+name: Build Tests
+
+on:
+  workflow_call:
+    inputs:
+      NR_JOBS:
```
volkm, line 6:
```suggestion
      nr_jobs:
```

.github/workflows/test-build.yml
```diff
@@ -0,0 +1,193 @@
...
+            --build-arg all_sanitizers="${{ inputs.AllSanitizers }}" \
+            --build-arg cmake_args="${{ inputs.cmakeArgs }}" \
+            --build-arg packages="${{ inputs.packages }}" \
+            --build-arg no_threads=${NR_JOBS}
+      - name: Export ccache from Docker image
+        run: |
+          rm -rf .ci-ccache
+          mkdir -p .ci-ccache
+          docker create --name ccache-extract movesrwth/storm:ci
+          docker cp ccache-extract:/root/.ccache/. .ci-ccache/ || true
```
volkm, line 148: Is it always the root directory or could it be a custom user directory?

.github/workflows/test-build.yml
```diff
@@ -0,0 +1,193 @@
...
+      name:
+        description: "Display name for the test configuration"
+        type: string
+        required: true
+      buildType:
+        description: "CMake build type (Debug or Release)"
+        type: string
+        default: "Debug"
+      Developer:
+        description: "Enable developer mode (ON/OFF)"
```
volkm, line 19: In the interest of nicer handling and especially avoiding invalid input, there could be two options for improvement:
1. Use custom options `"ON"/"OFF"` is inputs using [input type choice](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onworkflow_dispatchinputs)
2. Use boolean as input and convert to ON/OFF via something like `${{ input && 'ON' || 'OFF' }}`

.github/workflows/test-mac.yml
```diff
@@ -0,0 +1,136 @@
...
+      - name: Cache Homebrew downloads
+        if: inputs.restore_cache
+        uses: actions/cache@v4
+        with:
+          path: ~/Library/Caches/Homebrew
+          key: buildtest-brew-${{ inputs.distro }}-${{ hashFiles('.github/workflows/buildtest.yml') }}
+          restore-keys: |
+            buildtest-brew-${{ inputs.distro }}-
+      - name: Ensure ccache directory exists (macOS)
+        run: mkdir -p /Users/runner/.ccache
```
volkm, line 61:
```suggestion
        if: inputs.restore_cache
        run: mkdir -p /Users/runner/.ccache
```

.github/workflows/test-mac.yml
```diff
@@ -0,0 +1,136 @@
...
+            buildtest-ccache-macos-${{ inputs.distro }}-${{ inputs.buildType }}-
+            buildtest-ccache-macos-${{ inputs.distro }}-
+            buildtest-ccache-macos-
+      - name: Install dependencies
+        # cmake and gmp are already installed
+        run: |
+          brew update
+          brew install automake boost ccache cln ginac glpk hwloc libarchive xerces-c z3
+      - name: Prepare ccache
+        run: |
```
volkm, line 79: Also guarded by `save_cache` or `restore_cache`? Also in more places later on.

.github/workflows/test-linux.yml
```diff
@@ -0,0 +1,111 @@
...
+        run: |
+          mkdir -p .ci-ccache
+          docker build -t movesrwth/storm:ci . \
+            --build-arg BASE_IMAGE=movesrwth/storm-basesystem:${{ inputs.distro }} \
+            --build-arg build_type="${{ inputs.buildType }}" \
+            --build-arg no_threads=${NR_JOBS} \
+            --build-arg cmake_args="-DSTORM_WARNING_AS_ERROR=ON -DSTORM_COMPILE_WITH_PCH=OFF"
+            # Omitting arguments developer, disable_*, cln_exact, cln_ratfunc, all_sanitizers
+      - name: Export ccache from Docker image
+        run: |
```
volkm, line 64: Should be guarded by `save_cache` or `restore_cache`? Also in more places.

resources/scripts/benchmark-pch-ccache.sh
```diff
@@ -0,0 +1,295 @@
...
+    --cmake-arg <arg>       Extra CMake argument (can be repeated)
+    --trace                 Generate Chrome trace files via ninjatracing
+    --ninjatracing <path>   Path to local ninjatracing script/binary
+    --help                  Show this help
+
+Output:
+  - Summary table in stdout
+  - CSV at <build-root>/results.csv
+  - Per-run ccache stats in <build-root>/results/
+    - Optional traces at <build-root>/results/trace-<pch>-<cache>.json
```
volkm, line 45:
```suggestion
  - Optional traces at <build-root>/results/trace-<pch>-<cache>.json
```

volkm (commented):
Thanks for tackling this undertaking of revising our CI. I like the new structure and it should decrease the time for CI and make maintenance a bit easier. Great work!

I have a number of comments, I hope most are just small ones.

Dockerfile
lukovdm: The last two ccache calls are just to get proper stats. This was originally done to verify ccache was actually working. They can also be removed.

Dockerfile
lukovdm: Yes, I think installing ccache in the base images is a good idea.

Dockerfile
lukovdm: Does the docker not always build storm as root?

Dockerfile
lukovdm: I think if we have ccache in the base IMG, setting the CMAKE arg that disables ccache would do the trick.

src/storm/utility/SignalHandler.h
lukovdm: Storm now builds without these, was there a specific reason you wanted them?

resources/3rdparty/include_spot.cmake
lukovdm: I just tried to enable it and spot no longer builds on macos: https://github.com/stormchecker/storm/actions/runs/27282971720/job/80582285081?pr=892. I don't know why this happens.

lukovdm:
One storm parser test fails only on macos intel. It is quite a weird bug, this is what claude thinks about it:

> macOS Intel (Release) test crash: PrismParser.IllegalInputTest

> The test fails with libc++abi: terminating due to uncaught exception of type WrongFormatException, even though the parse is wrapped in EXPECT_THROW(..., WrongFormatException) (which has a catch(...)). So the exception isn't reaching any handler — it's an unwinding failure, not a parser or assertion bug.
> What's going wrong: SpiritErrorHandler::operator() throws the WrongFormatException from inside the handler, which Boost.Spirit invokes deep within its own -O3-inlined Qi/Phoenix template machinery (via on_error<qi::fail>). On the macOS x86_64 Release toolchain, the exception can't be unwound out through those optimized frames and hits std::terminate. It works everywhere else because Linux uses a more forgiving DWARF unwinder and the mac Debug builds (what master used) don't inline the fragile frames. This PR switched mac from Debug → Release, which is what surfaced it. Only Intel fails; ARM Release generates unwind info the unwinder copes with.
> This matches a known macOS/clang bug family where exceptions escape try/catch and terminate — see [LLVM #92121](https://github.com/llvm/llvm-project/issues/92121) (custom std::exception subclasses, optimized builds) and [OCaml #9026](https://github.com/ocaml/ocaml/issues/9026)/[#10423](https://github.com/ocaml/ocaml/issues/10423) (compact-unwind vs DWARF; the x86_64 fix was -Wl,-keep_dwarf_unwind).
> Note: it's not exception slicing — STORM_NEW_EXCEPTION's operator<< correctly returns the derived type.

> Directions to investigate:

> Proper fix: stop throwing from SpiritErrorHandler. on_error is meant to return a qi::error_handler_result — record the error message and return qi::fail, then throw the WrongFormatException from normal control flow after phrase_parse returns. This removes the throw-through-Spirit entirely. (Touches the Prism/Formula/Expression/IMCA grammars that share the handler.)

> Quick validation: try linking the macOS build with -Wl,-keep_dwarf_unwind (and/or -fno-omit-frame-pointer). If CI goes green, it confirms the compact-unwind root cause.

> Fallback: keep mac on Debug (as on master) if the above aren't pursued now.

I would suggest for now working around it and possibly making an issue.

.github/workflows/buildtest.yml
lukovdm: sadly not

.github/workflows/ci-pr.yml
```diff
@@ -1,494 +1,101 @@
...
-  # GitHub runners currently have 4 cores
-  NR_JOBS: "4"
-
 jobs:
-  # Perform in-depth tests with different configurations
-  indepthTests:
-    name: Indepth Tests (${{ matrix.config.name }}, ${{ matrix.config.buildType }})
-    runs-on: ubuntu-latest
+  buildTests:
+    if: "! contains(github.event.pull_request.labels.*.name, 'ready for review')"
```
lukovdm: Sadly not

.github/workflows/ci-ready-for-review.yml
```diff
@@ -0,0 +1,122 @@
...
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    types: [opened, synchronize, reopened, labeled]
+
+jobs:
+  buildTests:
+    if: contains(github.event.pull_request.labels.*.name, 'ready for review') || github.event_name == 'push'
```
lukovdm: Yes that is working now

.github/workflows/test-build.yml
lukovdm: I thought the docker always ran as the root user

volkm:
The addition of `<cstdint>` touches a lot of files. Can you shortly say why we need it?

volkm:
> One storm parser test fails only on macos intel. It is quite a weird bug, this is what claude thinks about it:
> 
> > macOS Intel (Release) test crash: PrismParser.IllegalInputTest
> 
> > The test fails with libc++abi: terminating due to uncaught exception of type WrongFormatException, even though the parse is wrapped in EXPECT_THROW(..., WrongFormatException) (which has a catch(...)). So the exception isn't reaching any handler — it's an unwinding failure, not a parser or assertion bug.
> > What's going wrong: SpiritErrorHandler::operator() throws the WrongFormatException from inside the handler, which Boost.Spirit invokes deep within its own -O3-inlined Qi/Phoenix template machinery (via on_errorqi::fail). On the macOS x86_64 Release toolchain, the exception can't be unwound out through those optimized frames and hits std::terminate. It works everywhere else because Linux uses a more forgiving DWARF unwinder and the mac Debug builds (what master used) don't inline the fragile frames. This PR switched mac from Debug → Release, which is what surfaced it. Only Intel fails; ARM Release generates unwind info the unwinder copes with.
> > This matches a known macOS/clang bug family where exceptions escape try/catch and terminate — see [LLVM #92121](https://github.com/llvm/llvm-project/issues/92121) (custom std::exception subclasses, optimized builds) and [OCaml #9026](https://github.com/ocaml/ocaml/issues/9026)/[#10423](https://github.com/ocaml/ocaml/issues/10423) (compact-unwind vs DWARF; the x86_64 fix was -Wl,-keep_dwarf_unwind).
> > Note: it's not exception slicing — STORM_NEW_EXCEPTION's operator<< correctly returns the derived type.
> 
> > Directions to investigate:
> 
> > Proper fix: stop throwing from SpiritErrorHandler. on_error is meant to return a qi::error_handler_result — record the error message and return qi::fail, then throw the WrongFormatException from normal control flow after phrase_parse returns. This removes the throw-through-Spirit entirely. (Touches the Prism/Formula/Expression/IMCA grammars that share the handler.)
> 
> > Quick validation: try linking the macOS build with -Wl,-keep_dwarf_unwind (and/or -fno-omit-frame-pointer). If CI goes green, it confirms the compact-unwind root cause.
> 
> > Fallback: keep mac on Debug (as on master) if the above aren't pursued now.
> 
> I would suggest for now working around it and possibly making an issue.

This should indeed be a separate issue. Good thing that the CI now catches this issue. Support for Intel will also be dropped with the upcoming release of macOS 27 anyway.

lukovdm:
Not using PCH removed stdint as a header from a lot of files which used its int types. Thus they had to be added manually.

volkm:
Right, thanks for the clarification.

So there are two main parts to this PR:
1. support for compiling without PCH (which needs additional includes)
2. the actual CI changes

I guess it is not easily possible to cherry-pick the commits for 1. and make a separate PR for a nicer division of concerns?
But at least good to know the different aspects for this PR.

lukovdm:
Yes correct, and indeed I think cherry picking out those commits will be quite difficult.

lukovdm:
Also, the latest CI run failed on a bus error in test-parser, I have no clue why this is going wrong. And it seems some caching is not working as the CI runs are taking around 40min again.

volkm:
The PCH changes were part of the separate https://github.com/stormchecker/storm/pull/974 to clean up this PR a bit.

resources/3rdparty/include_spot.cmake
volkm: The issue only happens for clang: `clang++: error: argument unused during compilation: '-pthread' [-Werror,-Wunused-command-line-argument]`
Will silence the warning.

volkm:
The PR should now be ready to be merged.
@lukovdm Can you take a last look at it and see whether you see anything which should be improved? You are the expert :)

volkm:
The one failing tests on macOS Intel will be addressed in a separate PR.

lukovdm:
I will give it a review now.

lukovdm:
LGTM, some of the CIs still take around 45 min, but that is for a later PR I think.

volkm:
Yes, some of the caches could still use some improvements. But at least we have good first version running.

Thanks for all your work in implementing, testing and investigating it @lukovdm!

PR #909, opened 2026-04-15
unif+: correctly handle Markov automata with non-Markovian initial states

### LINKED ISSUE #897: Wrong result with Unif+ and probabilistic initial state
Found by @kvdglind.

We consider the following MA which initially decides between taking either a path with an Erlang distribution or a faster approach but with probability of failure.

```
ma

const int K=2;
const double R=7;

module erlang
	s : [0..6];
	// s=5 sink
	// s=6 done
	chainStage : [0..K-1];

	[goToChain] s=0 -> 1 : (s'=1);
	<> s=1 -> 1 : (s'=2);
	<> s=2 & chainStage<K-1 -> R: (chainStage' = chainStage+1);
	<> s=2 & chainStage=K-1 -> R: (s' = 6);
	[goToProbs] s=0 -> 1 : (s'=3);
	<> s=3 -> 1 : (s'=4);
	<> s=4 -> 0.5 : (s'=6) + 0.5: (s'=5);
	<> s=5 -> true;
	<> s=6 -> true;
endmodule

label "done" = (s=6);
```

We run the following:
```
storm --prism erlang.ma --prop 'Pmin=?[F<=1 "done"]' --mamethod unifplus
```
This yields a wrong upper bound and an assertion error in debug mode:
```
ERROR (SparseMarkovAutomatonCslHelper.cpp:359): Upper bound 0.131893 is smaller than lower bound 0.132121.
Assertion failed: (absDiff > storm::utility::zero<ValueType>()), function checkConvergence, file SparseMarkovAutomatonCslHelper.cpp, line 390.
```

It goes wrong because the MA does not adhere to the *Initial Markov Assumption* (Assumption 8.1 in Timo Gros' Master thesis):
> "all states included in the initial distribution are Markov states."

This assumption is needed to make the transformation to CTMDP work.

I looked deeper into the analysis behavior and we see the following for the computation of the upper bound:
- In the inner iteration 64, the two Markovian states s=1 and s=3 have probability mass 0 and 0.00022797, respectively.
- The initial state minimizes and therefore chooses state s=1 with probability mass 0.
- In the next inner iteration 63 (counting down), the two Markovian states have probability mass 0.00744704 and 0.00167178, respectively.
- Now, the initial state chooses state s=3 with probability mass 0.00167178 and adds it to the probability mass of 0.
- In all subsequent iterations, state s=3 is always the lower probability and therefore chosen.
- In the end however, the probability mass in the initial state is 0.131893 whereas state s=3 has 0.132121. The difference is exactly the probability mass 0.00022797 which was "lost" in the inner iteration 64.

The behavior in the initial state therefore allows to switch the scheduler depending on the time. This is however not correct, because the initial decision immediately takes place at time 0.
I suggest to add a post-processing step here which explicitly accounts for initial states. This is in line with the suggestion of Timo who wrote in his thesis:

> "As we could interpret an initial probabilistic state as an extension of the initial distribution, we can easily transform every MA with initial probabilistic states such that this assumption holds. If an initial probabilistic state would have multiple probabilistic transitions, we would obtain a set of initial distributions. We would then regard a set of MAs with one of the initial distributions each and calculate time-bounded reachability for every MA and take the best result with respect to the subject of interest, i.e. maximal time-bounded reachability. Note that we would keep these probabilistic states whenever they have incoming edges, but still substitute them in the initial distribution."

So after finishing the computation for the upper bound, we should introduce a last model checking call where we compute the optimal probability from the initial state to the first layer of Markovian states. This should fix the issue in my view.

@tquatmann Do you agree or is there a better fix?

tquatmann:
Fixes #897

### DIFF
--- a/src/storm/modelchecker/csl/helper/SparseMarkovAutomatonCslHelper.cpp
+++ b/src/storm/modelchecker/csl/helper/SparseMarkovAutomatonCslHelper.cpp
@@ -98,7 +98,7 @@ class UnifPlusHelper {
             } else {
                 relevantMaybeStates = relevantStates.get() % maybeStates;
             }
-            relevantMarkovianMaybeStates = *relevantMaybeStates & markovianStatesModMaybeStates;
+            relevantMarkovianMaybeStates = relevantMaybeStates.get() & markovianStatesModMaybeStates;
         } else {
             relevantMarkovianMaybeStates = markovianStatesModMaybeStates;
         }
@@ -325,11 +325,11 @@ class UnifPlusHelper {
 
         if (abortedInnerIterations && iteration > 1 && relevantMaybeStates && relevantStates) {
             // We should take the stored solution instead of the current (probably more incorrect) lower/upper values
-            storm::utility::vector::setVectorValues(result, *relevantMaybeStates, bestKnownSolution);
+            storm::utility::vector::setVectorValues(result, relevantMaybeStates.get(), bestKnownSolution);
         } else {
             // Set the values for Markovian "maybe" states
             STORM_LOG_ASSERT(markovianMaybeStates.getNumberOfSetBits() == markovianStatesModMaybeStates.getNumberOfSetBits(),
-                             "unexpected number of Markovian maybe states");
+                             "Unexpected number of Markovian maybe states");
             auto subStateIt = markovianStatesModMaybeStates.begin();
             for (auto const markovianState : markovianMaybeStates) {
                 result[markovianState] = (maybeStatesValuesLower[*subStateIt] + maybeStatesValuesUpper[*subStateIt]) / two;
--- a/src/storm/modelchecker/csl/helper/SparseMarkovAutomatonCslHelper.cpp
+++ b/src/storm/modelchecker/csl/helper/SparseMarkovAutomatonCslHelper.cpp
@@ -88,13 +88,24 @@ class UnifPlusHelper {
         }
 
         boost::optional<storm::storage::BitVector> relevantMaybeStates;
+        storm::storage::BitVector relevantMarkovianMaybeStates;
         if (relevantStates) {
-            relevantMaybeStates = relevantStates.get() % maybeStates;
+            if (!relevantStates->isSubsetOf(markovianStates)) {
+                // Enhance the relevant states to also include Markovian states that are reachable in zero time from relevant probabilistic states
+                auto const enhancedRelevantStates =
+                    storm::utility::graph::getReachableStates(transitionMatrix, *relevantStates, ~markovianStates, markovianStates);
+                relevantMaybeStates = enhancedRelevantStates % maybeStates;
+            } else {
+                relevantMaybeStates = relevantStates.get() % maybeStates;
+            }
+            relevantMarkovianMaybeStates = *relevantMaybeStates & markovianStatesModMaybeStates;
+        } else {
+            relevantMarkovianMaybeStates = markovianStatesModMaybeStates;
         }
         // Store the best solution known so far (useful in cases where the computation gets aborted)
         std::vector<ValueType> bestKnownSolution;
         if (relevantMaybeStates) {
-            bestKnownSolution.resize(relevantStates->size());
+            bestKnownSolution.resize(relevantMaybeStates->getNumberOfSetBits());
         }
 
         // Get the exit rates restricted to only markovian maybe states.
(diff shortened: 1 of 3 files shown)


src/storm/modelchecker/csl/helper/SparseMarkovAutomatonCslHelper.cpp
```diff
@@ -88,13 +88,24 @@ class UnifPlusHelper {
...
-            relevantMaybeStates = relevantStates.get() % maybeStates;
+            if (!relevantStates->isSubsetOf(markovianStates)) {
+                // Enhance the relevant states to also include Markovian states that are reachable in zero time from relevant probabilistic states
+                auto const enhancedRelevantStates =
+                    storm::utility::graph::getReachableStates(transitionMatrix, *relevantStates, ~markovianStates, markovianStates);
+                relevantMaybeStates = enhancedRelevantStates % maybeStates;
+            } else {
+                relevantMaybeStates = relevantStates.get() % maybeStates;
+            }
+            relevantMarkovianMaybeStates = *relevantMaybeStates & markovianStatesModMaybeStates;
```
volkm: Just for consistency:
```suggestion
            relevantMarkovianMaybeStates = relevantMaybeStates.get() & markovianStatesModMaybeStates;
```

src/storm/modelchecker/csl/helper/SparseMarkovAutomatonCslHelper.cpp
```diff
@@ -309,42 +320,50 @@ class UnifPlusHelper {
 
         // Prepare the result vector
         std::vector<ValueType> result(transitionMatrix.getRowGroupCount(), storm::utility::zero<ValueType>());
+        // Set values of target states to 1
         storm::utility::vector::setVectorValues(result, psiStates, storm::utility::one<ValueType>());
 
         if (abortedInnerIterations && iteration > 1 && relevantMaybeStates && relevantStates) {
             // We should take the stored solution instead of the current (probably more incorrect) lower/upper values
-            storm::utility::vector::setVectorValues(result, maybeStates & relevantStates.get(), bestKnownSolution);
+            storm::utility::vector::setVectorValues(result, *relevantMaybeStates, bestKnownSolution);
```
volkm:
```suggestion
            storm::utility::vector::setVectorValues(result, relevantMaybeStates.get(), bestKnownSolution);
```

src/storm/modelchecker/csl/helper/SparseMarkovAutomatonCslHelper.cpp
```diff
@@ -309,42 +320,50 @@ class UnifPlusHelper {
...
         } else {
-            // We take the average of the lower and upper bounds
-            storm::utility::vector::applyPointwise<ValueType, ValueType, ValueType>(
-                maybeStatesValuesLower, maybeStatesValuesUpper, maybeStatesValuesLower,
-                [&two](ValueType const& a, ValueType const& b) -> ValueType { return (a + b) / two; });
-
-            storm::utility::vector::setVectorValues(result, maybeStates, maybeStatesValuesLower);
+            // Set the values for Markovian "maybe" states
+            STORM_LOG_ASSERT(markovianMaybeStates.getNumberOfSetBits() == markovianStatesModMaybeStates.getNumberOfSetBits(),
+                             "unexpected number of Markovian maybe states");
```
volkm:
```suggestion
                             "Unexpected number of Markovian maybe states");
```

src/storm/modelchecker/csl/helper/SparseMarkovAutomatonCslHelper.cpp
```diff
@@ -309,42 +320,50 @@ class UnifPlusHelper {
...
+            auto subStateIt = markovianStatesModMaybeStates.begin();
+            for (auto const markovianState : markovianMaybeStates) {
+                result[markovianState] = (maybeStatesValuesLower[*subStateIt] + maybeStatesValuesUpper[*subStateIt]) / two;
+                ++subStateIt;
+            }
+            // At this point, we only need to set the values for probabilistic "maybe" states.
+            // We derive the values from the values of the other states, which are reached in zero time.
+            uint64_t probSubState = 0;
+            for (auto const probabilisticState : probabilisticMaybeStates) {
```
volkm, line 341: Should we not do the solving for the probabilistic states as before? In lines 227-231 we use
```
// Compute the values at probabilistic states.
probabilisticToMarkovianMultiplier->multiply(env, nextMarkovianStateValues, nullptr, eqSysRhs);
for (auto const& oneStepProb : probabilisticToPsiProbabilities) {
    eqSysRhs[oneStepProb.first] += oneStepProb.second * targetValue;
}
```
I think the psi states should be taken into account?

volkm (approved):
Looks good overall.
Just one main comment regarding psi states, the other comments are very minor.

src/storm/modelchecker/csl/helper/SparseMarkovAutomatonCslHelper.cpp
tquatmann: The psi-states are handled by setting their values in the `result` vector to 1 above. The suggested code would also require to set `nextMarkovianStateValues` first, right?

I'm not sure if there would be a noticeable performance difference.

volkm: You are right. We can leave it as is.

volkm:
Thanks Tim!

PR #913, opened 2026-04-17
Fixed warning on dangling reference to tmp object.

### LINKED ISSUE #912: Building Storm in DEBUG fails: SparseModelFromUmb.cpp
When building the current main branch of Storm locally, I get the following error:
```
.../src/storm/storage/umb/import/SparseModelFromUmb.cpp: In function storm::models::sparse::StateLabeling storm::umb::detail::constructStateLabeling(const storm::umb::UmbModel&):
.../src/storm/storage/umb/import/SparseModelFromUmb.cpp:96:73: error: possibly dangling reference to a temporary [-Werror=dangling-reference]
   96 |         for (auto const& [apName, apIndex] : umbModel.index.aps().value()) {
      |                                                                         ^
```

I am not sure if this is a global issue or due to my local environment, but replacing:
```
for (auto const& [apName, apIndex] : umbModel.index.aps().value()) {
```
with
```
auto apsOpt = umbModel.index.aps();
auto const& aps = apsOpt.value();
for (auto const& [apName, apIndex] : aps) {
```
and applying the same pattern for `constructRewardModels` solves the issue for me.

It would be nice if someone could double-check this behavior, thanks!

volkm:
Fixes #912 with the suggestion of @plindnercs

### DIFF
--- a/src/storm/storage/umb/import/SparseModelFromUmb.cpp
+++ b/src/storm/storage/umb/import/SparseModelFromUmb.cpp
@@ -93,7 +93,8 @@ storm::models::sparse::StateLabeling constructStateLabeling(storm::umb::UmbModel
         stateLabelling.addLabel("init", storm::storage::BitVector(numStates, false));  // default to all states not being initial
     }
     if (umbModel.index.aps().has_value()) {
-        for (auto const& [apName, apIndex] : umbModel.index.aps().value()) {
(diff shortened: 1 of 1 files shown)


tquatmann (approved):
Thanks!

Honestly, I don't see any issue with the old code, but it's possible that I missed something.

PR #916, opened 2026-04-22
Bounded MC support for IDTMCs + Bugfixes + Unification of bounded MC logic

### LINKED ISSUE #914: Interval Models: Off by one error during bounded MC
I just noticed that the results for step-bounded reachability probabilities of interval models seem to suffer from an off by one error with respect to the step-bound.
Consider the following illustration of the IMDP test file `tiny-03.drn`:
Image

The test case `BoundedTiny03maxmin` in `RobustMdpPrctlModelCheckerTest.cpp` checks the properties: `Pmax=? [ F<=3 "target"]` and `Pmin=? [ F<=3 "target"]`, with both robust and cooperative resolution of the uncertainty.
For this setting, the test works, but when changing the properties to: `Pmax=? [ F<=1 "target"]` and `Pmin=? [ F<=1 "target"]`, Storm returns a probability of `0` in every combination of optimization direction and uncertainty resolution, which is not correct to my understanding and also fails the test. Allowing two steps instead of one then leads to the expected result.

I am currently working on adapting the bounded MC implementation for IMDPs to IDTMCs and would also address this issue in my changes. Nonetheless, I wanted to report this error here in case anyone encounters it in the meantime.

plindnercs:
This pull request includes:
- bounded MC support for IDTMCs,
- unification of the bounded MC logic for deterministic and non-deterministic models and
- a fix for the off-by-one error during bounded MC of interval models,

and thus resolves #914.


resources/examples/testfiles/dtmc/tiny-01.drn
```diff
@@ -0,0 +1,28 @@
+@type: DTMC
+@parameters
```
tquatmann, line 3:
```suggestion
@value_type: double
@parameters
```

src/storm/modelchecker/helper/finitehorizon/SparseStepBoundedHorizonHelper.h
tquatmann:
```suggestion
```

tquatmann:
```suggestion
```

src/storm/modelchecker/helper/finitehorizon/SparseStepBoundedHorizonHelper.cpp
```diff
@@ -0,0 +1,198 @@
+#include "SparseStepBoundedHorizonHelper.h"
```
tquatmann:
```suggestion
#include "storm/modelchecker/helper/finitehorizon/SparseStepBoundedHorizonHelper.h"
```

src/storm/modelchecker/helper/finitehorizon/SparseStepBoundedHorizonHelper.cpp
```diff
@@ -0,0 +1,198 @@
...
+#include "storm/adapters/IntervalAdapter.h"
+#include "storm/adapters/IntervalForward.h"
```
tquatmann:
```suggestion
```

The `*Forward` headers don't need to be included when the actual header is included, too.

Same for RationalNumberForward below.

src/storm/modelchecker/helper/finitehorizon/SparseStepBoundedHorizonHelper.cpp
```diff
@@ -0,0 +1,198 @@
...
+#include "storm/adapters/RationalFunctionAdapter.h"
+#include "storm/adapters/RationalNumberAdapter.h"
+#include "storm/adapters/RationalNumberForward.h"
+#include "storm/modelchecker/hints/ExplicitModelCheckerHint.h"
+#include "storm/modelchecker/prctl/helper/SparseMdpEndComponentInformation.h"
```
tquatmann:
```suggestion
```

src/storm/modelchecker/helper/finitehorizon/SparseStepBoundedHorizonHelper.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+#include "storm/models/sparse/StandardRewardModel.h"
+
+#include "storm/storage/BitVector.h"
+#include "storm/utility/graph.h"
+#include "storm/utility/macros.h"
+#include "storm/utility/vector.h"
+
+#include "storm/environment/solver/MinMaxSolverEnvironment.h"
```
tquatmann:
```suggestion
```

src/storm/modelchecker/helper/finitehorizon/SparseStepBoundedHorizonHelper.cpp
```diff
@@ -0,0 +1,198 @@
...
+            }
+        }
+
+        if (lowerBound == 0) {
+            maybeStates &= ~psiStates;
+        }
+
+        if constexpr (storm::IsIntervalType<ValueType>) {
+            maybeStates &= ~psiStates;
+        }
```
tquatmann: Why are we always excluding psiStates for interval models?

Also, this avoids potentially excluding them twice :)
```suggestion
        if (storm::IsIntervalType<ValueType> || lowerBound == 0) {
            maybeStates &= ~psiStates;
        }
```

src/storm/modelchecker/helper/finitehorizon/SparseStepBoundedHorizonHelper.cpp
```diff
@@ -0,0 +1,198 @@
...
+        optimizationDirection = OptimizationDirection::Maximize;
+    } else {
+        optimizationDirection = goal.direction();
+    }
+
+    // If we identify the states that have probability 0 of reaching the target states, we might be able to exclude them in the further analysis.
+    // For the 'maybeStates' we definitely have to compute the values.
+    storm::storage::BitVector maybeStates = computeMaybeStates(goal, transitionMatrix, backwardTransitions, phiStates, psiStates, lowerBound, upperBound, hint);
+    storm::storage::BitVector makeZeroColumns;
+
```
tquatmann: Add a comment here that explains the procedure for lowerbound!=0 and/or interval types

plindnercs:
Thanks for your feedback @tquatmann! I was able to simplify the logic of computing the 'maybeStates' a bit and added some comments for further clarification.

tquatmann:
I had some trouble following the code in the SparseStepBoundedHorizonHelper (both, interval and non-interval code). Of course, this has been an issue already before this PR.

Anyway, I have spent some effort to (hopefully) simplify the computations and streamline the different cases. Would be good if someone can have a short look at `SparseStepBoundedHorizon.cpp` before merging.

src/storm/modelchecker/helper/finitehorizon/SparseStepBoundedHorizonHelper.cpp
```diff
@@ -0,0 +1,179 @@
...
+    if (lowerBound > upperBound) {
+        return std::vector<SolutionType>(transitionMatrix.getRowGroupCount(), storm::utility::zero<SolutionType>());
+    }
+
+    storm::solver::OptimizationDirection const optimizationDirection =
+        transitionMatrix.hasTrivialRowGrouping() ? OptimizationDirection::Maximize : goal.direction();
+
+    // We identify states that must have probability 0 of reaching the target states to exclude them in the further analysis.
+    storm::storage::BitVector const probGreater0States =
+        detail::computeProbGreater0States<ValueType, SolutionType>(goal, transitionMatrix, backwardTransitions, phiStates, psiStates, upperBound, hint);
```
sjunges, line 76: Should one do this analysis also in two phases, as in, first look for upperBound-lowerBound, as  that is a smaller number it may be quicker and more states will be zero, which is good for the first phase, and for the second phase, we can then use the actual zeros from the first phase with again a reduced number of iterations. 

I do think the current implementation is also correct, just potentially a bit slower (caching effects may also invalidate anything i said)...

sjunges:
LGTM! Many thanks both.

src/storm/modelchecker/helper/finitehorizon/SparseStepBoundedHorizonHelper.cpp
tquatmann: Indeed, this could work, and would probably be a bit faster in cases where the bounds are small compared to the number of states. Not sure how common this case is.

However, I think I will merge this PR for now, so that we have a correct implementation in `master`.

PR #917, opened 2026-04-28
ConditionalHelper: fix printing warning in harmless cases

tquatmann:
The warning
```cpp
                STORM_LOG_WARN("Precision of non-exact type reached during bisection method. Result might be inaccurate.");
```

was printed in all cases where lower and upper bounds became close.
However, this only becomes problematic, if we don't terminate in that iteration (because then the subtile differences between lower- and upper bound values tend to become indistinguishable in floating-point arithmetic).


volkm:
LGTM.
The clang CI failure is unrelated.

PR #918, opened 2026-04-28
Fix compilation on recent clang

tquatmann:
Should fix the failing build

https://github.com/stormchecker/storm/actions/runs/24984410065/job/73154004843

Also avoids creating an unnecessary copy of the properties vector.


volkm:
Great, thanks! LGTM.

PR #925, opened 2026-05-06
Fix incorrect assert in observation trace unfolder

lukovdm:
An assert in src/storm-pomdp/transformer/ObservationTraceUnfolder.cpp was incorrect and should have been inverted.


sjunges:
Oops. LGTM!

PR #926, opened 2026-05-06
 CI: revised workflow for creating deb package

volkm:
- rename workflow to `create_deb`
- made workflow reusable: it now takes the Storm version for the Debian package as input. Note that this version is only for the Debian package and not the CMake version, because CMake does not like suffixes such as `nightly`.
- separated the test stage from the build stage. Variable `artifact` is used to map tests to a specific package version. We currently need to build for Debian 12 and Debian 13 because Ginac changes from `..so.11` to `so.13` and libraries cannot be found otherwise.


.github/workflows/create_deb.yml
```diff
@@ -0,0 +1,137 @@
...
+        distro:
+          - {name: "debian:12",
+             tag: "debian-12",
+             artifact: "debian12",
+            }
+          - {name: "debian:13",
+             tag: "debian-13",
+             artifact: "debian13",
+            }
+          - {name: "ubuntu:24.04",
```
sjunges, line 96: you can consider adding ubuntu 26.04 these days... :-)

.github/workflows/create_deb.yml
```diff
@@ -0,0 +1,137 @@
...
+      - name: Install and test package
+        run: |
+          docker run -d -it --name ci-test ${{ matrix.distro.name }}
+          docker cp package/*.deb ci-test:/opt/
+          docker exec ci-test bash -c "
+            set -e
+            apt-get update
+            apt-get install -y /opt/*.deb
+            storm --version
+          "
```
sjunges, line 123: I would consider splitting the commands so that it is clearer where the error occurs (in installing or in running storm. 

Same below for starter project, I think splitting into being able to build the starter project from running it (and potentially just the download step) could help?

sjunges:
LGTM, really cool!

.github/workflows/create_deb.yml
volkm: `ubuntu:rolling` currently also points to 26.04. But it is still good to add it here as an LTS.

.github/workflows/create_deb.yml
volkm: Good point. Done.

PR #927, opened 2026-05-07
Version 1.13.0

volkm:


volkm:
Can you check the release note?
In particular for https://github.com/stormchecker/storm/pull/886 and https://github.com/stormchecker/storm/pull/887 details on the changes were limited.

lukovdm:
I guess we could also expand upon the interval point by saying we prepared the internals of storm to make future additions in interval model checking easier, e.g., by adding interval support to multipliers, or maybe allowing for having a different valuetype and solutiontype in more places. There was a lot of internal work done that should allow intervals to become a more first party part of storm. I am not sure how internal or just cli/stormpy user facing the release notes should be.

tquatmann:
LGTM!

PR #928, opened 2026-05-08
import/export of umb models with rational intervals

tquatmann:
Also adds CLI support for interval Prism models in --exact mode.

### DIFF
--- a/src/storm-cli-utilities/model-handling.h
+++ b/src/storm-cli-utilities/model-handling.h
@@ -363,15 +363,15 @@ auto castAndApply(std::shared_ptr<storm::models::ModelBase> const& model, auto c
     // branch on type of value representation
     if (model->supportsParameters()) {
         return castAndApplyVT.template operator()<storm::RationalFunction>();
-    } else if (model->isExact()) {
-        if (model->supportsUncertainty()) {
+    } else if (model->supportsUncertainty()) {
+        if (model->isExact()) {
             return castAndApplyVT.template operator()<storm::RationalInterval>();
         } else {
-            return castAndApplyVT.template operator()<storm::RationalNumber>();
+            return castAndApplyVT.template operator()<storm::Interval>();
         }
     } else {
-        if (model->supportsUncertainty()) {
-            return castAndApplyVT.template operator()<storm::Interval>();
+        if (model->isExact()) {
+            return castAndApplyVT.template operator()<storm::RationalNumber>();
         } else {
             return castAndApplyVT.template operator()<double>();
         }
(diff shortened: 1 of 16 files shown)


src/storm-cli-utilities/model-handling.h
```diff
@@ -364,11 +364,17 @@ auto castAndApply(std::shared_ptr<storm::models::ModelBase> const& model, auto c
     if (model->supportsParameters()) {
         return castAndApplyVT.template operator()<storm::RationalFunction>();
     } else if (model->isExact()) {
-        return castAndApplyVT.template operator()<storm::RationalNumber>();
-    } else if (model->supportsUncertainty()) {
-        return castAndApplyVT.template operator()<storm::Interval>();
+        if (model->supportsUncertainty()) {
```
sjunges: Not terribly important, but I would have expected to first ask whether it supports uncertainty (after asking whether we support parameters), and if so, whether we have exact set.

src/storm-cli-utilities/model-handling.h
```diff
@@ -566,9 +572,13 @@ std::shared_ptr<storm::models::ModelBase> buildModelSparse(SymbolicInput const&
     // Adapt the ValueType if it does not support intervals and the input is an interval model
     if (!storm::IsIntervalType<ValueType> && input.model.is_initialized() && input.model->isPrismProgram() &&
         input.model->asPrismProgram().hasIntervalUpdates()) {
-        STORM_LOG_THROW((std::is_same_v<ValueType, storm::IntervalBaseType<storm::Interval>>), storm::exceptions::NotSupportedException,
-                        "Can not build interval model for the provided value type.");
-        return storm::api::buildSparseModel<storm::Interval>(input.model.get(), options);
+        if (std::is_same_v<ValueType, storm::IntervalBaseType<storm::Interval>>) {
```
sjunges: I dont understand what !storm::IsIntervalType<ValueType> guards if we then ask whether valuetype  is the same as ...

src/storm-cli-utilities/model-handling.h
```diff
@@ -566,9 +572,13 @@ std::shared_ptr<storm::models::ModelBase> buildModelSparse(SymbolicInput const&
     // Adapt the ValueType if it does not support intervals and the input is an interval model
     if (!storm::IsIntervalType<ValueType> && input.model.is_initialized() && input.model->isPrismProgram() &&
```
sjunges, line 573: Does Jani not allow interval-based transition probabilities?

sjunges:
LGTM, many thanks. I am only confused by this buildSparseModel implementation.

src/storm-cli-utilities/model-handling.h
tquatmann: No, JANI has no support for that afaik.

src/storm-cli-utilities/model-handling.h
tquatmann: The outer "if" is to check whether we need to change ValueType to some interval type.
The inner "if" then checks whether that interval type should be `storm::Interval` or `storm::RationalInterval`.  That check is based on whether ValueType is double or storm::RationalNumber.

Note that `storm::IntervalBaseType<storm::Interval>>` evaluates to `double`.

I revised the code a little bit, hoping that it is now clearer.

sjunges:
Thanks, this is much easier to digest 👍

PR #933, opened 2026-05-19
Multi-objective: fix issues with negative rewards and transition rewards

tquatmann:
Thanks @carolinager for pointing out these issues.


sjunges:
LGTM!

PR #934, opened 2026-05-19
RationalInterval: Fix DRN parsing and added UMB/DRN test cases.

tquatmann:
Rational-interval DRN files could not be parsed so far because that case was missing.

Thanks @plindnercs for pointing this out!


volkm (approved):
LGTM. Thanks

PR #937, opened 2026-06-11
Revised State  and observation valuations

tquatmann:
Replaces the sparse StateValuations by a new implementation that stores the values as specified by [umb](https://pmc-tools.github.io/umb/spec).

## Example CLI Usage

Note that we still have to add the --buildstateval switch

```terminal
# create umb file with state valuations
 storm --buildstateval --prism crowds5_5.pm --exportbuild crowds-storm.umb 
# import umb file with state valuations
storm --buildstateval -umb crowds-storm.umb
```

## Observation Valuations

For POMDPs, we can also use the --buildobsval switch to add observation valuations.
This no longer treats observations as strings, but rather as a valuation.

## Compatibility

So far, I've manually tested whether we can exchange models with state valuations with prism.

## Implementation Details

The core parts are in `storm/storage/umb/model/Valuations.h`. Reading from and writing to variables works via callbacks. This has the advantage, that we always access the variable value in the most suitable type (Side note: for integer variables, the "right" type (uint64_t vs. int64_t vs. arbitrary precision Integer) is not easy to determine).

`storm/storage/sparse/valuations.h` serves as a higher-level wrapper that doesn't expose callbacks. It mostly restores the old functionalities of `StateValuations.h`. Notable changes:

- the old `blowUp` method is now `selectEntities`
- hash() now does something properly (it just returned 0 before)
- Iterating over all values is now slightly different. We can either invoke `getAllVariables()` and then get the value for each variable or we use a callback function via `getUmbValuations().readCallback...`.


sjunges:
I am somewhat confused that creating state valuations is now done via the umb (sub)library, rather than via what is in sparse. In e.g., the monitor-verifier, this yields very unexpected inclusions.

src/storm-pomdp/generator/GenerateMonitorVerifier.cpp
```diff
@@ -331,82 +334,42 @@ std::shared_ptr<MonitorVerifier<ValueType>> GenerateMonitorVerifier<ValueType>::
...
-                                booleanValues.push_back(val.getBooleanValue());
-                            } else if (val.isInteger()) {
-                                integerValues.push_back(val.getIntegerValue());
-                            } else if (val.isRational()) {
-                                rationalValues.push_back(val.getRationalValue());
-                            }
-                            break;
+                stateValuations.writeCallback(prodToIndexMap[std::make_pair(i, j)], [this, &oldValuations, i, j](auto, auto const& var, auto& value) {
+                    if (var == monvar || var == mcvar) {
+                        if constexpr (std::is_same_v<std::remove_cvref_t<decltype(value)>, int64_t>) {
```
sjunges: ehm; what is this line doing? :-)

src/storm-pomdp/transformer/ObservationTraceUnfolder.cpp
```diff
@@ -42,9 +45,15 @@ std::shared_ptr<storm::models::sparse::Mdp<ValueType>> ObservationTraceUnfolder<
...
-    storm::storage::sparse::StateValuationsBuilder svbuilder;
-    svbuilder.addVariable(svvar);
-    svbuilder.addVariable(tsvar);
+    storm::umb::Valuations stateValuations = [this, &observations]() {
+        storm::umb::ValuationDescriptionBuilder svBuilder(exprManager);
+        svBuilder.addIntegerVariable(svvar, -1, static_cast<int64_t>(model.getNumberOfStates()) - 1);
+        svBuilder.addIntegerVariable(tsvar, -1, observations.size() - 1);
+        return storm::umb::Valuations(svBuilder.buildClassDescription(), exprManager);
+    }();
+    auto addStateValuation = [this, &stateValuations](int64_t s, int64_t t) {
```
sjunges, line 57: Should we adopt a naming convention for lambdas? I missed where the function was defined.

src/storm/generator/VariableInformation.cpp
```diff
@@ -94,13 +94,17 @@ uint64_t getBitWidthLowerUpperBound(bool const& hasLowerBound, int64_t& lowerBou
 
 VariableInformation::VariableInformation(storm::prism::Program const& program, uint64_t reservedBitsForUnboundedVariables, bool outOfBoundsState)
     : totalBitOffset(0) {
+    auto getFreshName = [&program](std::string const& baseName) {
```
sjunges: Do we never rely on these names coinciding etc. I see how this is a local fix, but I wonder whether it hides problems downstream.

src/storm/storage/sparse/Valuations.h
```diff
@@ -0,0 +1,142 @@
...
+     */
+    std::set<storm::expressions::Variable> getAllVariables() const;
+
+    /*!
+     * Returns true iff the variable is relevant for the given entity
+     */
+    bool entityHasVariable(uint64_t entity, storm::expressions::Variable const& variable) const;
+
+    // --- getters for variable values ---
+    // optional varians have no value iff either entityHasVariable(entity, variable) is false or the value is of optional type and not set.
```
sjunges: variants

src/storm/storage/sparse/Valuations.h
```diff
@@ -0,0 +1,142 @@
...
+     * @return the variables that this stores valuations for
+     */
+    std::set<storm::expressions::Variable> getAllVariables() const;
+
+    /*!
+     * Returns true iff the variable is relevant for the given entity
+     */
+    bool entityHasVariable(uint64_t entity, storm::expressions::Variable const& variable) const;
+
+    // --- getters for variable values ---
```
sjunges: not a fan of such comments :-)

src/storm/storage/sparse/Valuations.cpp
```diff
@@ -0,0 +1,278 @@
...
+    : umbValuations(std::make_unique<storm::umb::Valuations>(umbValuationDescription, manager)) {
+    umbValuations->resize(numEntities);
+    // Intentionally empty
+}
+
+Valuations::Valuations(storm::umb::Valuations&& umbValuations) : umbValuations(std::make_unique<storm::umb::Valuations>(std::move(umbValuations))) {
+    // Intentionally empty
+}
+
+// The following need to be defined in the .cpp file since the umb::Valuations type definition must be complete. This allows forward-declaring the
```
sjunges: I dont understand why these definitions cannot be put in the header and the comment does not help me at all.

sjunges:
Overall looks very good. I dont like that the valuations are in the UMBI folder as noted above and the callbacks are partially a bit hard to read, but not sure how to improve that.

tquatmann:
> I am somewhat confused that creating state valuations is now done via the umb (sub)library, rather than via what is in sparse. In e.g., the monitor-verifier, this yields very unexpected inclusions.

The internal storage should be UMB compliant, which is why the file is currently located there.
 I can see that this is currently weird when we just need a data structure for valuations without an immediate intention to export it. My suggestion is to

- `mkdir storm/storage/valuations` and move relevant files there.
- This includes  `storm/storage/umb/model/Valuations.[h|cpp]` (rename to `ValuationsStorage`), `storm/storage/umb/utility/ValuationDescriptionBuilder.[h|cpp]`, and `storm/storage/umb/model/ValuationDescription.[h|cpp]`.
- Add a corresponding comment in those files that need to be UMB compliant
- Rename `storm::storage::sparse::StateValuations::getUmbValuations()` to `getStorage()`

There will still be some transitive(!) umb includes, but I guess that's ok

tquatmann:
> the callbacks are partially a bit hard to read, but not sure how to improve that.

Me neither. They are quite flexible and efficient but that also makes them awkward sometimes. I'm open for feedback

src/storm-pomdp/transformer/ObservationTraceUnfolder.cpp
tquatmann: Any suggestions? I don't think such a convention is standard. I added a comment here to make it more visible.

src/storm-pomdp/generator/GenerateMonitorVerifier.cpp
tquatmann: I added a comment that hopefully explains things.

src/storm/generator/VariableInformation.cpp
tquatmann: For example [Prism allows](https://www.prismmodelchecker.org/manual/ThePRISMLanguage/PartiallyObservableModels) this

```
observable "s" = s;
```

Now I wonder if even this would be allowed:
```
observable "s" = s+x;
```

I can see that this is problematic. We could also just fail with an error message in the second case, i.e., whenever there is a `observable "x" = expr;` statement where `x` is also a used variable identifier and `expr` is not just the variable expression `x`.

tquatmann: Anything else would require a second expression manager for the "observation valuations" which I'd prefer not to have

src/storm/storage/sparse/Valuations.cpp
tquatmann: Short answer: It doesn't compile otherwise.

Medium answer: In those function, we potentially need to delete a `storm::umb::Valuations*` due to the `umbValuations` member. This requires `storm::umb::Valuations` to be a complete type, which is not true in the `.h` but true in the `.cpp`.

Longish answer:

- `#include "storm/storage/sparse/Valuations.h"` should be somewhat lightweight imo, so we should not have `#include "storm/storage/umb/model/Valuations.h"` in  `storm/storage/sparse/Valuations.h`.
- the member `std::unique_ptr<storm::umb::Valuations> umbValuations` will thus be a unique ptr to an incomplete type, which is mostly fine *except* for its deleter (search for "incomplete type" [here](https://en.cppreference.com/cpp/memory/unique_ptr))
- This intuitively makes sense: When we want to delete a `T*` pointer, we must know what to do, so we must know the "insights" of the type `T`. Deleting a `std::unique_ptr<T>` has the same problem.
- Consequently, every function that might need to call the deleter of `umbValuations` needs a complete definition of `storm::umb::Valuations` and therefore can not be implemented in the `.h`.
- All these might need to delete `umbValuations`:
```
Valuations::~Valuations() = default;
Valuations::Valuations(Valuations&& other) = default;
Valuations& Valuations::operator=(Valuations&& other) = default;
```

Some further information can be found [here](https://howardhinnant.github.io/incomplete.html)

sjunges:
Thanks for the fixes and clarifications. LGTM. Shall we merge or should I first create updated bindings?

tquatmann:
I made two more commits following the discussions above:
- one commit creates `src/storm/storage/valuations/`, moves things there (in particular of the umb scope), and adapted namespaces and function names accordingly 
- one that rejects Prism Programs in one of the two cases:
  - There is a variable or constant named `_OutOfBoundsBit` and `--build-out-of-bounds-state --buildstateval` is set
  - There is a label `observable "s" = <expr>;` and `s` is already a variable and the expression `<expr>` is not equal to `s`.

tquatmann:
> Shall we merge or should I first create updated bindings?

imo, we can delay merging this a bit to reduce the time in which stormpy is broken with storm's main.

Feel free to merge this at your convenience :)

And thanks for the great feedback!

src/storm/generator/VariableInformation.cpp
sjunges: i think this observable "s" = s; is an antipattern... i would not allow it? :-)

tquatmann: `observable "s" = s;` is explicitly mentioned on the PRISM webpage.
I have seen a number of PRISM files with e.g. `label "done" = done;`, so I get where also allowing this for `observable` is coming from.
IMO `observable "s" = s;` is not bad enough to justify breaking compatibility with PRISM.

It also avoids having to mix both syntaxes, which is arguably quite clean

```
observables s endobservables
observable "parity" = x % 2;
```

vs.

```
observable "s" = s;
observable "parity" = x % 2;
```

tquatmann: After internal discussions, we decided to allow the pattern `observable "s" = s` but not `observable "s" = expr` for variable `s` and `expr` ≠ `s` (as it's implemented right now)

src/storm/storage/valuations/ValuationsStorage.h
```diff
@@ -0,0 +1,770 @@
...
+                Integer value = varInfo.description.type.type == Int ? readInteger<true>(getRawBytes(entity), varInfo.bitOffset, bitSize)
+                                                                     : readInteger<false>(getRawBytes(entity), varInfo.bitOffset, bitSize);
+                value += storm::utility::convertNumber<Integer>(varInfo.description.offset.value_or(0));
+                if (invokeCallback(value)) {
+                    return;
+                }
+                break;
+            }
+            case Double:
+                // Reaching this part should not be possible as varInfo.fits64Bit would be true
```
sjunges: I dont understand this. This seems similar to the bool case; it is not necessarily reasonable, but you can fit a double in more than 64 bits?

src/storm/storage/valuations/ValuationsStorage.h
```diff
@@ -0,0 +1,770 @@
...
+                uint64_t const b = bitSize / 2;
+                storm::RationalNumber const numerator = readInteger<true>(getRawBytes(entity), varInfo.bitOffset, b);
+                storm::RationalNumber const denominator = readInteger<false>(getRawBytes(entity), varInfo.bitOffset + b, b);
+                if (invokeCallback(storm::RationalNumber(numerator / denominator))) {
+                    return;
+                }
+                break;
+            }
+            case String: {
+                // Reaching this part should not be possible as varInfo.fits64Bit would be true
```
sjunges: Same thing as above

src/storm/storage/valuations/ValuationsStorage.h
tquatmann: A bool can have an arbitrary size (≥1 bit) whereas for double and string the size must always be 64 bits.

From the [spec](https://pmc-tools.github.io/umb/spec):

```
var Type = schema({  
  "type": [  
    "bool", // Boolean: 0 is false, anything else is true; default size is 1  
    ...
    "double", // continuous numeric; 64-bit IEEE 754 floating-point number; size is/must always be 64  
    ...
   "string", // size is/must always be 64 (for a uint64 index into a string mapping)  
  ...
```

sjunges: Ah, yes, I think I violated that spec somehwere in my test. Let me try to figure that out.

sjunges: And what happens here if you call getRationalValue if the variable has type double?

tquatmann: Ahh yes, this is likely the problem. Thanks for catching!
Would we want an implicit conversion or an explicit error in that case?

randriu:
Will this support multiple valuation classes? As per UMB spec, an ATS can have multiple `ValuationClassDescription` and each entity can be associated with a different valuation class.

tquatmann:
> Will this support multiple valuation classes? As per UMB spec, an ATS can have multiple `ValuationClassDescription` and each entity can be associated with a different valuation class.

Yes, this is supported, although I have to admit that it is not well tested. I could not get my hands on any "real" UMB model that uses this feature.

randriu:
Great. I'm not sure I'm umbi supports this, but if it does or once it does, I'll test it.

sjunges:
Lets merge!

PR #938, opened 2026-06-16
Bump gmm to 5.5

volkm:
- the patch from https://github.com/stormchecker/storm/pull/844 is still needed
- disable support for BLAS, LAPACK, etc via patch


volkm:
The linker error seems to stem from a missing BLAS or LAPACK library (according to Copilot).
Maybe we should properly build gmm as a dependency and not just include the sources?

sjunges:
Probably better. There seems to be initial cmake support for getfem/gmm++ these days, which would probably be good to use then?

volkm:
The CMake support is not part of the official releases. In this case, we need to use the git it seems.

I looked a bit into it and "building" gmm does not produce a binary, the installation step just copies the source files. So why does it not compile anymore? The new release ships a config with enabled support for BLAS, LAPACK, etc. whereas previously this was not the case.

Possible approaches:
1. use `autoreconf`, `./configure`, `make` to properly configure and build gmm.
   Advantage: more control over configuration.
   Disadvantage: needs reconfiguration, also needs autoconf (but is needed anyway for CUDD)
2. just patch the config file to disable support for BLAS, LPACK
   Advantage: only minor change in one file, no need to run `./configure`
   Disadvantage: more brittle long-term
3. Try to use CMake support
   Advantage: better integration, future-proof
   Disadvantage: support is currently WIP, requires download of git instead of release

sjunges:
I prefer 2 or 3.

volkm:
I will try approach 2 for now.
The CMake build [hardcodes](https://github.com/getfem/getfem/blob/master/CMakeLists.txt#L637) dependencies to BLAS and LAPACK. It was also not clear to me whether just a build of gmm (without GetFEM) is possible.

sjunges:
LGTM, many thanks!

PR #941, opened 2026-06-28
Fix scheduler extraction for topological solver

### LINKED ISSUE #940: MDP scheduler chooses incorrect action
When exporting a scheduler for the MDP model ``mdp_sched_error.drn`` below using command
```
--exact --io:explicit-drn "mdp_sched_error.drn" --build:buildchoicelab--io:prop 'Pmax=? [F "target"]' --io:exportscheduler "mdp_sched.json"
```
the scheduler extraction provides an incorrect action for state 1.
Instead of action``go``, which ensures progress, the extraction yields ``cycle`` as the chosen action, which corresponds to a scheduler with value 0.
The model checking itself yields the correct result ``1/2``.

---
```
@type: MDP
@value_type: rational
@parameters

@reward_models

@nr_states
5
@nr_choices
7
@model
state 0 init
	action go
		1 : 1
state 1 
	action cycle
		1 : 1
	action go
		2 : 1
state 2
	action win
		3 : 1/2
		4 : 1/2
	action go
		4 : 1
state 3 target
	action __loop__
		3 : 1
state 4 bottom
	action __loop__
		4 : 1
```

tquatmann:
The scheduler is already computed correctly for each SCC. Then, for whatever reason, it is currently overwritten afterwards by a potentially incorrect scheduler.

Fixes #940

### DIFF
--- a/src/storm/solver/TopologicalMinMaxLinearEquationSolver.cpp
+++ b/src/storm/solver/TopologicalMinMaxLinearEquationSolver.cpp
@@ -149,15 +149,6 @@ bool TopologicalMinMaxLinearEquationSolver<ValueType, SolutionType>::internalSol
                 break;
             }
         }
-
-        // If requested, we store the scheduler for retrieval.
-        if (this->isTrackSchedulerSet()) {
-            if (!auxiliaryRowGroupVector) {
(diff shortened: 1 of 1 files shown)


sjunges:
LGTM

PR #942, opened 2026-06-29
Require CMake >=3.25

volkm:
Same as https://github.com/stormchecker/carl-storm/pull/142
The oldest supported OS is Debian 12 which ships with CMake 3.25.


sjunges:
LGTM.

PR #944, opened 2026-06-29
Lexicographic: Fail for unsupported properties, remove deprecated --lex switch.

tquatmann:
- Lexicographic model checking now fails for unsupported inputs (instead of giving an unexpected result)
- The deprecated --lex switch has been removed. Lexicographic model checking can be invoked using the `multilex(..)` syntax for PRISM-style properties or via JANI


sjunges:
LGTM

PR #946, opened 2026-07-01
Make dtmc model checker throw on unreachable condition when checking conditional probabilities.

lukovdm:
Includes fix and added tests for this case.


volkm (approved):
LGTM.
Maybe use a different type of exception such as `InvalidPropertyException`?

lukovdm:
I was being consistent the the conditional MDP case, but I agree it is a bit weird. I will change both the DTMC and the MDP one to `InvalidPropertyException`.

src/storm/modelchecker/prctl/helper/SparseDtmcPrctlHelper.cpp
```diff
@@ -860,33 +860,33 @@ std::vector<SolutionType> SparseDtmcPrctlHelper<ValueType, RewardModelType, Solu
...
         // Prepare result vector.
         std::vector<ValueType> result(transitionMatrix.getRowCount(), storm::utility::infinity<ValueType>());
-
-        if (!conditionStates.empty()) {
-            BaierTransformedModel transformedModel =
-                computeBaierTransformation(env, transitionMatrix, backwardTransitions, targetStates, conditionStates, boost::none);
-
-            if (transformedModel.noTargetStates) {
-                storm::utility::vector::setVectorValues(result, transformedModel.beforeStates, storm::utility::zero<ValueType>());
+        STORM_LOG_THROW(!conditionStates.empty(), storm::exceptions::InvalidPropertyException,
```
tquatmann: Do we have to check that  conditionStates are reachable from all relevant (i.e. initial) states?

Something like

```cpp
auto const conditionGreaterZeroStates = storm::utility::graph::performProbGreater0(backwardTransitions, allStates, conditionStates);
STORM_LOG_THROW(goal.hasRelevantStates && goal.getRelevantStates().isSubsetOf(conditionGreaterZeroStates), ...);
```

PR #953, opened 2026-07-24
Remove unused exceptions

volkm:
Some of the exceptions sound like they should be useful but apparently we never use them.
I am also happy to keep them and actually start using them.

Related to https://github.com/stormchecker/storm/issues/952


tquatmann:
LGTM!

PR #955, opened 2026-07-25
let an agent fix #313, #319, #335.

### LINKED ISSUE #313: Weird behaviour which causes ERROR (Type.cpp:235): Operator requires boolean operands.
Hello,

while working with a PRISM model using Storm (1.7.0 built from source on Apple M1 chip) for my thesis I discovered a peculiar behaviour with Storm.

When a formula, which has int return type, is used in different formula which has a return type of boolean, Storm gives out the error "_ERROR (Type.cpp:235): Operator requires boolean operands._".

The actual PRISM (4.7) model checker doesn't have a problem with this. Through more experimenting I discovered that Storm still managed to **correctly evaluate the formula**.

I'm attaching a simple PRISM model which can replicate the issue and contains the specific formulas I was using. I have also discovered, that moving the int formula above the boolean formula manages to avoid the error. Unfortunately Github doesn't support the .pm file format so here is the raw text:

```
mdp

//formula current_min_leader_calc = min((A_mark?A_leader:8), (B_mark?B_leader:8));

formula leader_set = ((A_mark & A_leader = current_min_leader_calc | A_mark = false)?true:false) & ((B_mark & B_leader = current_min_leader_calc | B_mark = false)?true:false);

formula current_min_leader_calc = min((A_mark?A_leader:8), (B_mark?B_leader:8));

//It seems that using a formula, which has a int return type, inside a formula which has a boolean return type causes the error ERROR (Type.cpp:235): Operator requires boolean operands.
//Although it seems that the formula does work.
//BUT if the current_min_leader_calc is positioned above the leader_set formula (the commented formula), then there are no errors.

module whatever

	A_mark : bool init true;
	B_mark : bool init true;

	test1 : bool init false;

	A_leader : [0..10] init 1;
	B_leader : [0..10] init 1;

	[] A_mark = true & leader_set & test1 = false -> (test1'=true);
	//If A_mark is set to false, the model has only 1 state, which means the formula was evaluated correctly to false.
	//If A_mark is set to true, the model has 2 states, so test1 was updated to true, so the formula was evaluated correctly to true.


endmodule
```

### LINKED ISSUE #319: PRISM file parser error
Hi,

i was trying to use [this](https://www.prismmodelchecker.org/tutorial/examples/circadian/circadian.sm) prism model from the prism tutorials with storm.  I get the following error message:
```
Storm 1.7.1 (dev)

Date: Wed Dec  7 15:52:08 2022
Command line arguments: --prism circadian.prism
Current working directory: /home/mfg/

ERROR (SpiritErrorHandler.h:27): Parsing error at 28:3:  expecting "endmodule", here:
          ma : [0..N]; // amount of activator mRNA
          ^

ERROR (storm.cpp:21): An exception caused Storm to terminate. The message of the exception is: WrongFormatException: Parsing error at 28:3:  expecting "endmodule", here:
          ma : [0..N]; // amount of activator mRNA
          ^

```

Is there a workaround (or fix) available?

### LINKED ISSUE #335: Prism parser: Player is a keyword for MDPs
Imho, player should only be a reserved keyword when parsing games. In particular, player is not a keyword for regular prism and thus, marking player as a keywords yields incompatibility with prism.

sjunges:
Fixes #313, fixes #319, fixes #335 with minor bugs around building prism files.

### DIFF
--- a/src/storm-parsers/parser/PrismParserGrammar.cpp
+++ b/src/storm-parsers/parser/PrismParserGrammar.cpp
@@ -90,7 +90,11 @@ PrismParserGrammar::PrismParserGrammar(std::string const& filename, Iterator fir
       filename(filename),
       annotate(first),
       manager(new storm::expressions::ExpressionManager()),
-      expressionParser(new ExpressionParser(*manager, keywords_, false, false)) {
+      // "player"/"endplayer" are only reserved for SMGs (see isValidIdentifier), so they must not be
+      // blanket-rejected as identifiers inside expressions either. Note: qi::symbols' copy constructor
+      // aliases the underlying trie (shared_ptr), so expressionKeywords_ is built independently, not by
+      // copying keywords_ and removing entries (that would mutate keywords_ too).
+      expressionParser(new ExpressionParser(*manager, expressionKeywords_, false, false)) {
     ExpressionParser& expression_ = *expressionParser;
     boolExpression = (expression_[qi::_val = qi::_1])[qi::_pass = phoenix::bind(&PrismParserGrammar::isOfBoolType, phoenix::ref(*this), qi::_val)];
     boolExpression.name("boolean expression");
@@ -425,7 +429,8 @@ PrismParserGrammar::PrismParserGrammar(std::string const& filename, Iterator fir
     qi::on_success(assignmentDefinition, setLocationInfoFunction);
 
     // Enable error reporting.
-    qi::on_error<qi::fail>(start, handler(qi::_1, qi::_2, qi::_3, qi::_4));
+    qi::on_error<qi::fail>(
+        start, (phoenix::bind(&PrismParserGrammar::reportRejectedKeywordIdentifier, phoenix::ref(*this)), handler(qi::_1, qi::_2, qi::_3, qi::_4)));
 }
 
 void PrismParserGrammar::moveToSecondRun() {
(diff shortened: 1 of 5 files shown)


src/storm-parsers/parser/PrismParserGrammar.cpp
```diff
@@ -90,7 +90,11 @@ PrismParserGrammar::PrismParserGrammar(std::string const& filename, Iterator fir
       filename(filename),
       annotate(first),
       manager(new storm::expressions::ExpressionManager()),
-      expressionParser(new ExpressionParser(*manager, keywords_, false, false)) {
+      // "player"/"endplayer" are only reserved for SMGs (see isValidIdentifier), so they must not be
+      // blanket-rejected as identifiers inside expressions either. Note: qi::symbols' copy constructor
+      // aliases the underlying trie (shared_ptr), so expressionKeywords_ is built independently, not by
+      // copying keywords_ and removing entries (that would mutate keywords_ too).
```
tquatmann: I find this comment distracting at this point in the code. It's good to mention the issue somewhere, but imo not here.

src/storm-parsers/parser/PrismParserGrammar.h
```diff
@@ -127,6 +127,15 @@ class PrismParserGrammar : public qi::grammar<Iterator, storm::prism::Program(),
         }
     };
 
+    // Same as keywordsStruct, without "player"/"endplayer" (see expressionKeywords_).
+    struct expressionKeywordsStruct : qi::symbols<char, uint_fast64_t> {
+        expressionKeywordsStruct() {
+            add("dtmc", 1)("ctmc", 2)("mdp", 3)("ctmdp", 4)("ma", 5)("pomdp", 6)("pta", 7)("smg", 8)("const", 9)("int", 10)("bool", 11)("module", 12)(
+                "endmodule", 13)("rewards", 14)("endrewards", 15)("true", 16)("false", 17)("min", 18)("max", 19)("floor", 20)("ceil", 21)("init", 22)(
+                "atLeastOneOf", 23)("atMostOneOf", 24)("exactlyOneOf", 25)("endinit", 26)("invariant", 27)("endinvariant", 28);
```
tquatmann: we could argue that the player/endplayer issue for non-games also applies to invariant/endinvariant for non-pta models

tquatmann:
good agent. 🤖

src/storm-parsers/parser/PrismParserGrammar.cpp
sjunges: Agree, such comments should not be.

src/storm-parsers/parser/PrismParserGrammar.h
sjunges: Yes, we can do that. Invariant is less of a standard variable name though :-)

PR #957, opened 2026-07-27
Add missing <map> include in Formula.h

tquatmann:
Formula.h uses `std::map` but did not include `<map>` directly, relying on transitive includes.


volkm (approved):
LGTM.

PR #958, opened 2026-07-29
Introduce STORM_LOG_STATS

sjunges:
as discussed in https://github.com/stormchecker/storm/discussions/829


src/storm-cli-utilities/cli.cpp
```diff
@@ -78,6 +80,16 @@ void setLogLevel() {
...
     }
+
+    // Statistics and progress messages live on their own log channels so they can be enabled
+    // independently of general verbosity.
+    if (core.isShowStatisticsSet()) {
+        storm::utility::setStatisticsLogLevel(l3pp::LogLevel::INFO);
+    }
+    if (general.isShowProgressSet()) {
+        storm::utility::setProgressLogLevel(l3pp::LogLevel::INFO);
+    }
```
tquatmann, line 91: Why is statistics in CoreSettings and progress in GeneralSettings?

src/storm-counterexamples/counterexamples/SMTMinimalLabelSetGenerator.h
```diff
@@ -1976,27 +1975,28 @@ class SMTMinimalLabelSetGenerator {
         stats.solverTime = std::chrono::duration_cast<std::chrono::milliseconds>(totalSolverTime);
         stats.iterations = iterations;
 
-        if (storm::settings::getModule<storm::settings::modules::CoreSettings>().isShowStatisticsSet()) {
+        {
             storm::storage::FlatSet<uint64_t> allLabels;
             for (auto const& e : labelSets) {
                 allLabels.insert(e.begin(), e.end());
             }
```
tquatmann, line 1981: This code is now executed regardless of whether statistics are enabled. This particular case seems *okay*, but generally speaking: is there any way to execute code only if statistics are enabled?

src/storm-cli-utilities/cli.cpp
sjunges: Good question, but that was not part of this PR and restructuring the settings would be I would do in a different PR.

src/storm-counterexamples/counterexamples/SMTMinimalLabelSetGenerator.h
sjunges: As with logging, code is not executed if the logging is disabled. The trick is that you then have to basically put this into a lambda/a function, similar to regular logging. 

You can of course currently also access the settings if you want to.

volkm: I agree with @tquatmann  here that there are now some code parts where preprocessing is done before the logging. This could potentially still be executed even even if nothing is logged.
I would prefer to either keep the statistics check here or use a lambda function.

volkm:
Looks overall good to me.

Some functions now directly write to cout. Would it make sense (in a separate PR) to output to a general ostream instead of cout?

src/storm-counterexamples/counterexamples/SMTMinimalLabelSetGenerator.h
sjunges: This code fragment is incredibly cheap and I really think that that lambda will just make the code harder to read. I understood @tquatmann as a general question, not as a concern regarding this code.

volkm: I also meant it as a general comment. There are other places such as GameBasedMdpModelChecker.cpp or SparseExplorationModelChecker.cpp where some preprocessing takes place. It might also be cheap, but just wanted to be sure.

sjunges: Ah, oka.. These places should be cheap based on my understanding of the code. I have not benchmarked those places; they are all a bit out of the main code paths in Storm these days. I could lambda them to have code parts where this happens, if you prefer this.

src/storm/modelchecker/exploration/SparseExplorationModelChecker.cpp
```diff
@@ -131,10 +132,9 @@ std::tuple<StateType, typename ModelType::ValueType, typename ModelType::ValueTy
         }
     }
 
-    // Show statistics if required.
-    if (storm::settings::getModule<storm::settings::modules::CoreSettings>().isShowStatisticsSet()) {
-        stats.printToStream(std::cout, explorationInformation);
-    }
+    std::stringstream statsStream;
+    stats.printToStream(statsStream, explorationInformation);
```
volkm, line 136: Processing is only gathering some numbers, should be fast.

src/storm/modelchecker/multiobjective/pcaa/StandardPcaaWeightVectorChecker.cpp
```diff
@@ -124,21 +122,18 @@ void StandardPcaaWeightVectorChecker<SparseModelType>::initialize(
...
-            }
-            STORM_PRINT_AND_LOG(numLraMecStates << " states lie on such an end component.\n");
+    STORM_LOG_STATISTICS("Weight Vector Checker Statistics:\n");
+    STORM_LOG_STATISTICS("Final preprocessed model has " << transitionMatrix.getRowGroupCount() << " states.\n");
+    STORM_LOG_STATISTICS("Final preprocessed model has " << transitionMatrix.getRowCount() << " actions.\n");
+    if (lraMecDecomposition) {
+        STORM_LOG_STATISTICS("Found " << lraMecDecomposition->mecs.size() << " end components that are relevant for LRA-analysis.\n");
+        uint64_t numLraMecStates = 0;
+        for (auto const& mec : this->lraMecDecomposition->mecs) {
+            numLraMecStates += mec.size();
```
volkm, line 132: Should be okay as processing time.

src/storm/modelchecker/reachability/SparseDtmcEliminationModelChecker.cpp
```diff
@@ -318,13 +317,13 @@ SparseDtmcEliminationModelChecker<SparseDtmcModelType>::computeLongRunValues(sto
         std::chrono::high_resolution_clock::duration totalTime = totalTimeEnd - totalTimeStart;
         std::chrono::milliseconds totalTimeInMilliseconds = std::chrono::duration_cast<std::chrono::milliseconds>(totalTime);
```
volkm, line 316: No overhead

src/storm/storage/bisimulation/BisimulationDecomposition.cpp
```diff
@@ -242,19 +240,19 @@ void BisimulationDecomposition<ModelType, BlockDataType>::computeBisimulationDec
 
     std::chrono::high_resolution_clock::duration totalTime = std::chrono::high_resolution_clock::now() - totalStart;
 
-    if (storm::settings::getModule<storm::settings::modules::CoreSettings>().isShowStatisticsSet()) {
+    {
         std::chrono::milliseconds initialPartitionTimeInMilliseconds = std::chrono::duration_cast<std::chrono::milliseconds>(initialPartitionTime);
         std::chrono::milliseconds refinementTimeInMilliseconds = std::chrono::duration_cast<std::chrono::milliseconds>(refinementTime);
         std::chrono::milliseconds extractionTimeInMilliseconds = std::chrono::duration_cast<std::chrono::milliseconds>(extractionTime);
         std::chrono::milliseconds quotientBuildTimeInMilliseconds = std::chrono::duration_cast<std::chrono::milliseconds>(quotientBuildTime);
         std::chrono::milliseconds totalTimeInMilliseconds = std::chrono::duration_cast<std::chrono::milliseconds>(totalTime);
```
volkm, line 248: No overhead

src/storm-gamebased-ar/modelchecker/abstraction/GameBasedMdpModelChecker.cpp
```diff
@@ -1744,48 +1745,44 @@ template<storm::dd::DdType Type, typename ModelType>
...
-        std::cout << "    * abstraction: " << totalAbstractionTimeMillis << "ms (" << 100 * static_cast<double>(totalAbstractionTimeMillis) / totalTimeMillis
-                  << "%)\n";
-        if (this->solveMode == storm::settings::modules::AbstractionSettings::SolveMode::Sparse) {
-            std::cout << "    * translation: " << totalTranslationTimeMillis << "ms ("
-                      << 100 * static_cast<double>(totalTranslationTimeMillis) / totalTimeMillis << "%)\n";
-            if (fixPlayer1Strategy || fixPlayer2Strategy) {
-                std::cout << "    * strategy processing: " << totalStrategyProcessingTimeMillis << "ms ("
-                          << 100 * static_cast<double>(totalStrategyProcessingTimeMillis) / totalTimeMillis << "%)\n";
-            }
+    storm::gbar::abstraction::AbstractionInformation<Type> const& abstractionInformation = abstractor.getAbstractionInformation();
```
volkm, line 1748: Only returns the object and the rest should also not take much time.

src/storm-counterexamples/counterexamples/SMTMinimalLabelSetGenerator.h
volkm: I went through the code paths and it seems that they do not really add much overhead (see comments). For clarity sake, we should maybe leave it as is.

src/storm/utility/ProgressMeasurement.cpp
```diff
@@ -23,14 +22,12 @@ void ProgressMeasurement::startNewMeasurement(uint64_t startCount) {
...
-            // Message already contains line break at the end.
-            STORM_PRINT_AND_LOG(message);
-            return true;
-        }
+    std::stringstream stream;
+    if (updateProgress(count, stream)) {
+        std::string message = stream.str();
+        // Message already contains line break at the end.
+        STORM_LOG_PROGRESS(message);
+        return true;
```
volkm: This part could be expensive if we call it thousands of times. I think we should only execute it if we need to display it.

volkm:
I think we can keep the current state. I only suggest to revise updateProgress because this can be called thousands of times.

sjunges:
I have updated this, using new macros that make the code rather reasonable to write, I think.

volkm:
Yes, the lazy channels is a nice idea.
I am not the biggest expert on lambdas but it looks good to me.

sjunges:
@tquatmann @volkm Ready to be merged?

src/storm/utility/logging.h
```diff
@@ -24,3 +26,37 @@ using carl::operator<<;
...
+#define STORM_LOG_LAZY_ON_CHANNEL(channel, level, ...)                                                \
+    do {                                                                                              \
+        auto storm_log_lazy_channel = l3pp::Logger::getLogger(channel);                               \
+        if (storm_log_lazy_channel->getLevel() <= (level)) {                                          \
+            std::ostringstream storm_log_lazy_stream;                                                 \
+            if ((__VA_ARGS__)(storm_log_lazy_stream)) {                                               \
+                storm_log_lazy_channel->log(level, __L3PP_LOG_RECORD) << storm_log_lazy_stream.str(); \
+            }                                                                                         \
+        }                                                                                             \
+    } while (false)
```
tquatmann: Why do we need a variadic macro and not just `callback` as an argument?

Could then add a  `static_assert(std::is_invocable_r_v<bool, decltype(callback), std::ostream&>,
                  "callback must have signature bool(std::ostream&)");` (needs  `#include <type_traits>`) for better compiler error message.

tquatmann (approved):
LGTM! Just a minor comment.

sjunges:
@volkm Do you know why this is not scheduled for CI actions anymore?

volkm:
https://www.githubstatus.com/
Problems with Github actions in the last few hours

PR #959, opened 2026-07-29
rename --additional-checks to --check-zeno,

### LINKED ISSUE #803: Storm --additional-checks
Following up on a discussion #802, the debug settings `additional-checks` are used only once in the current code base, to check for the presence of zeno cycles. 

@tquatmann @volkm I would suggest to do the following. Move the flag to the build-settings, call it --check-zeno. 
We can then discuss whether we want to assume-zeno and invert the logic to say that we always check for zeno cycles unless we skip them.

(There is of course a question what to do with MAs that are built differently...)

sjunges:
fixes #803


tquatmann (approved):
LGTM!

PR #960, opened 2026-07-29
more robust auto format

sjunges:
The auto-format / ninja format / make format often used a formatter that was a slightly different version. This updated autoformat script tries to avoid that.


volkm:
Good idea. LGTM.

It would be even more robust, if we could define the clang-format version in one central location. Currently `formatapply` and `formatcheck` also need to be consistent. But I did not see an easy solution.

PR #961, opened 2026-07-29
fix #701: allow observables throughout

### LINKED ISSUE #701: PRISM parser for POMDPs fails when observables is not the first definition after model header
If I try to parse a POMDP in PRISM format using the "observables ... endobservables" syntax [(PRISM documentation)](https://www.prismmodelchecker.org/manual/ThePRISMLanguage/PartiallyObservableModels), it seems to only work if this definition is the first thing that's defined after the `pomdp` header.

In the attachment (see below), there are 2 files with the same POMDP defined, the only difference is that the `test-fail.prism` contains a definition of some formula before the definition of observables. The `test-ok.prism` is parsed fine and `test-fail.prism` fails with:

```
ERROR (SpiritErrorHandler.h:26): Parsing error at 6:5:  expecting "=", here:
                x
            ^

ERROR (storm-cli.cpp:63): An exception caused Storm to terminate. The message of the exception is: WrongFormatException: Parsing error at 6:5:  expecting "=", here:
                x
            ^
```

PRISM parses both of these without any issues.

[pomdps-test.zip](https://github.com/user-attachments/files/19957160/pomdps-test.zip)

sjunges:
fixes #701: allow observables throughout.

Turned out to be a simple fix in the grammar

### DIFF
--- a/src/storm-parsers/parser/PrismParserGrammar.cpp
+++ b/src/storm-parsers/parser/PrismParserGrammar.cpp
@@ -229,8 +229,9 @@ PrismParserGrammar::PrismParserGrammar(std::string const& filename, Iterator fir
          qi::lit("endinit"))[qi::_pass = phoenix::bind(&PrismParserGrammar::addInitialStatesConstruct, phoenix::ref(*this), qi::_1, qi::_r1)];
     initialStatesConstruct.name("initial construct");
 
-    observablesConstruct = (qi::lit("observables") > (identifier % qi::lit(",")) >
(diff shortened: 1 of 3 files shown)


tquatmann:
If I see this correctly (boost::spirit and I have a special relationship...), then this new grammar allows multiple `observables .. endobservables` constructs. Is that intended?

sjunges:
Well-spotted. The prism manual is not so clear on this, but it yields strange situatiosn if we allow it. Now handled as init blocks.

tquatmann (approved):
LGTM!

PR #964, opened 2026-07-30
Add [[maybe_unused]] to variables only used in assert

volkm:
Were triggered for `CMAKE_BUILD_TYPE=Release STORM_DEVELOPER=ON`.
(Need to add as config to CI)

Also similar changes in https://github.com/stormchecker/carl-storm/pull/144


sjunges:
LGTM

PR #966, opened 2026-07-31
Cleanup/bisim and building options

sjunges:
Builds upon #958.

Towards the elimination of an implicit global state via the settings. work on bisimulation and model building. 

Note that this also enforces building and bisimulation with zero tolerance on exact models as a design decision."
[cleanup/bisim-and-building-options 2837c28ed] towards the elimination of an implicit global state via the settings: work on bisimulation and model building. Note that this also enforces building and bisimulation with zero tolerance on exact models as a design decision


src/storm/builder/BuilderOptions.h
```diff
@@ -214,6 +214,27 @@ class BuilderOptions {
...
 
+    /**
+     * Sets the tolerance used for checking whether a distribution sums to one.
+     * @param newValue The new tolerance
+     * @return this
+     */
+    BuilderOptions& setStochasticTolerance(double newValue);
+
+    /**
+     * Should the progress of state space exploration be printed?
```
volkm:
```suggestion
     * Sets whether the progress of state space exploration should be printed.
```

src/storm/storage/bisimulation/BisimulationDecomposition.cpp
```diff
@@ -54,7 +54,8 @@ BisimulationDecomposition<ModelType, BlockDataType>::Options::Options()
       keepRewards(false),
       type(BisimulationType::Strong),
       bounded(false),
-      discounted(false) {
+      discounted(false),
+      tolerance(storm::NumberTraits<ValueType>::IsExact ? storm::utility::zero<ValueType>() : storm::utility::convertNumber<ValueType>(1e-6)) {
```
volkm: Magic number, but we already use it in many places anyway.

volkm (approved):
LGTM. Thanks for tackling it.

src/storm/storage/bisimulation/BisimulationDecomposition.cpp
sjunges: Still, should not occur here. Let me try to eliminate that appropriately.

sjunges: This was a bit more work than I thought. I am now moving some default readings to the api/ Getting rid of them there may be something for another PR (as we may want to discuss this).

sjunges:
Along with magic constants, I also fixed two remaining fix-deadlock occurrences.

volkm:
The update looks good to me.

I agree that we should later try to get rid of the settings in `bisimulation.h` by explicitly requiring a tolerance.

sjunges:
@copilot resolve the merge conflicts in this pull request

PR #967, opened 2026-07-31
Consistent throwing of exceptions

### LINKED ISSUE #817: Consistently use MissingLibraryException
The exception should be used in all cases where a 3rdparty library is not available.

volkm:
- Use `STORM_LOG_THROW` everywhere
- Simplified some parts by moving the condition inside `STORM_LOG_THROW`
- Use `MissingLibraryException` for 3rdparty dependencies. Fixes #817 
- Consistent exception message: starting with upper case letter, ending with punctuation mark

Contributions from opencode [bot@opencode.ai](mailto:bot@opencode.ai) (model: opencode/big-pickle).


sjunges:
Does someone have the github copilot for students/educators here? This looks like a very mechanical job and it would be great if they could point their agent at reviewing this :-)

sjunges:
Note that this fixes #817 to the best of my understanding (good to mention in opening post).

volkm:
It is indeed very mechanical. I manually reviewed the code changes for the conditionals and everything looks good to me (so the first six commits). I did not go through all message changes in the last commit in detail though.

volkm:
Good catches from the review. Will merge it if the CI runs through.

PR #970, opened 2026-08-01
Cleanup/lp solver settings

sjunges:
Reduce the dependency on a global state. 

- Needed to push the environment through some POMDP code, even when that is outdated code.
- Dropped the smtlib solver and some unused includes as a side effect.


src/storm-pomdp/modelchecker/BeliefExplorationPomdpModelChecker.h
volkm, line 301:
```suggestion
     * @param env Environment
     * @param clippingStateId the state ID of the clipping belief
```

volkm, line 315:
```suggestion
     * @param env Environment
     * @param clippingStateId the state ID of the clipping belief
```

src/storm/solver/GurobiLpSolver.cpp
```diff
@@ -35,7 +33,8 @@ GRBenv* GurobiEnvironment::operator*() {
 }
 #endif
 
-void GurobiEnvironment::initialize() {
+void GurobiEnvironment::initialize(storm::GurobiSolverEnvironment const& gurobiSettings, bool debug) {
```
volkm, line 36: I found it a bit confusing that the environment is called `Settings`. But given that there are also `GurobiSolverEnvironment` and `GurobiEnvironment`, I also have no better name.

volkm (approved):
LGTM.

The doc comments are not important.

volkm:
A more general questions (maybe for another discussion): are there any sanity checks on the values for environments? Previously, the settings took care of it, but if we now allow changes via the environment we might want to have some checks somewhere. But maybe we only need it in stormpy.

sjunges:
> A more general questions (maybe for another discussion): are there any sanity checks on the values for environments? Previously, the settings took care of it, but if we now allow changes via the environment we might want to have some checks somewhere. But maybe we only need it in stormpy.

I raised this point in https://github.com/stormchecker/storm/issues/977. I would like to do this in a separate step and we should potentially discuss how to handle this.

PR #974, opened 2026-08-04
Add option to enable/disable precompiled headers

volkm:
CMake option `STORM_COMPILE_WITH_PCH` (default `ON`) allows to configure whether to compile with/without precompiled headers.
Includes various system includes which are needed without PCH.

Extracted the relevant parts from @lukovdm PR https://github.com/stormchecker/storm/pull/892 to have a better separation.


sjunges:
LGTM

PR #979, opened 2026-08-05
separate model-handling, cli-utils no longer depend on counterexamples

### LINKED ISSUE #209: Storm-cli-utilities requires storm-counterexamples
The file `model-handling.h` requires the library `storm-counterexamples` for e.g. printing counterexamples.
I am wondering if we could remove this dependency and move the support for counterexamples into the dedicated library.

sjunges:
Fixes #209.

Tackling the settings is a separate concern.


src/storm-cli/CMakeLists.txt
```diff
@@ -1,9 +1,10 @@
 # Create main binary storm.
 add_executable(storm-cli ${PROJECT_SOURCE_DIR}/src/storm-cli/storm-cli.cpp)
-target_link_libraries(storm-cli storm storm-cli-utilities)
+target_link_libraries(storm-cli storm storm-cli-utilities storm-counterexamples storm-gamebased-ar)
+target_include_directories(storm-cli PRIVATE "${PROJECT_SOURCE_DIR}/src")
```
volkm, line 4: Is it necessary to include the whole src directory?
But I see that we also do the same for storm-pars-cli and storm-pomdp-cli, so I am fine with it.

src/storm/settings/modules/CounterexampleGeneratorSettings.h
volkm, line 1: Why is the move to storm necessary? Can it not stay in storm-counterexamples?

sjunges: No, there is still a dependency on the counterexample settings somehwere in the build (for better or worse). As we work towards moving the settings into the cli-utilities, this problem will more or less solve itself.

src/storm-cli/CMakeLists.txt
sjunges: Yep, this is what we also did for the other CLIs and i think we do this for consistency. Note that the includes are private. I didnt want to touch this point in this PR.

sjunges:
Woops, forgot to commit. Will do this once I am back on my machine.

src/storm-cli-utilities/model-handling.h
```diff
@@ -1,30 +1,24 @@
 #pragma once
 
-#include <filesystem>
+#include <boost/algorithm/string/join.hpp>
 #include <type_traits>
 
 #include "storm/api/storm.h"
 
 #include "AutomaticSettings.h"
```
volkm:
```suggestion
#include "storm-cli-utilities/AutomaticSettings.h"
```

volkm: Not part of the original change but still good to have consistent

volkm (approved):
LGTM.
Mostly moving complete functions from storm-cli-utilities/model-handling.h to storm-cli/model-handling-main-cli.h. I locally checked the diff that these functions do not change.

PR #980, opened 2026-08-05
Support for linter via clang-tidy

volkm:
Linter support via clang-tidy.

Enable with `STORM_ENABLE_CLANG_TIDY` in CMake. Provides the following targets:
- `clang-tidy`: Runs clang-tidy (as configured in .clang-tidy) over the Storm sources.
- `clang-tidy-fix`: Like clang-tidy, but automatically applies the suggested fixes.
- `clang-tidy-changed`: Like clang-tidy, but only checks the files changed in git (staged, unstaged and untracked).
- `clang-tidy-changed-fix`: Like clang-tidy-changed, but automatically applies the suggested fixes.

As first step, I only activated one rule in `.clang-tidy`. Will add separate PRs enabling and fixing more checks.

CI support via [cpp-linter-action](https://cpp-linter.github.io/cpp-linter-action/). This does both linting and format check. The CI only checks the changed code and then uses review comments to indicate changes. An example can be seen [in my fork](https://github.com/volkm/storm/pull/8).
Note that these review comments are currently experimental and might be a bit noisy if too many things need to be changed. There is also the alternative of just a single comment, see [this PR example](https://github.com/volkm/storm/pull/6).

I personally liked the review comments, because they allow to directly apply the suggested changes. I recommend to apply changes in a batch though (tab `Files changed` and then select `Add suggestion to batch`).

The CI provides also a `lint-full.yml` workflow which can be manually triggered. It runs over the whole code base.

Clang-tidy can also be run locally with the script `resources/cmake/clang-tidy.sh`. This is similar to the auto-format scripts and provides several additional options (see help command).

I also slightly revised the auto-format scripts which now provides a check option (Cmake target `format-check`) to only check the formatting.
Both scripts take the clang-format and clang-tidy version from `resources/cmake/clang-versions`. This should ensure consistency between the local versions and the CI.

Note that the CI action for formatting only supports clang-format<=20 whereas the clang-tidy CI supports <=22. In particular, the clang-tidy CI also runs clang-formatting as an integrated check.
If things go smoothly we could remove the formatcheck we currently use and only use lint-check. Or at least we should use consistent versions.

Lastly, I added CMake options `STORM_ENABLE_CLANG_TIDY` and `STORM_ENABLE_CLANG_FORMAT`. The FORMAT option is not really necessary, I mainly added it for symmetry reasons. The TIDY option mainly guards `set(CMAKE_EXPORT_COMPILE_COMMANDS ON)` but this should not have a big impact in my view. So we could think about removing both options and simply always enabling the targets.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,56 @@
+# See https://clang.llvm.org/extra/clang-tidy/ for a list of available checks.
+# Disable everything first, then enable individual checks.
+Checks: >-
+  -*,
+  readability-braces-around-statements
+# Candidate checks that are not (yet) enabled:
+# bugprone-branch-clone,
+# bugprone-copy-constructor-init,
+# bugprone-dangling-handle,
+# bugprone-infinite-loop,
+# bugprone-misplaced-widening-cast,
+# bugprone-redundant-expression,
+# bugprone-signed-char-misuse,
+# bugprone-unused-return-value,
+# bugprone-use-after-move,
+# bugprone-virtual-near-miss,
+# clang-analyzer-core.*,
+# clang-analyzer-cplusplus.*,
+# clang-analyzer-deadcode.*,
+# clang-analyzer-nullability.*,
+# clang-analyzer-security.*,
+# clang-analyzer-unix.*,
+# misc-misplaced-const,
+# misc-static-assert,
+# misc-throw-by-value-catch-by-reference,
+# misc-unused-using-decls,
+# modernize-deprecated-headers,
+# modernize-make-shared,
+# modernize-make-unique,
+# modernize-return-braced-init-list,
+# modernize-use-default-member-init,
+# modernize-use-nullptr,
+# modernize-use-using,
+# performance-for-range-copy,
+# performance-implicit-conversion-in-loop,
+# performance-move-constructor-init,
+# performance-unnecessary-copy-initialization,
+# performance-unnecessary-value-param,
+# readability-container-size-empty,
+# readability-else-after-return,
+# readability-non-const-parameter,
+# readability-redundant-member-init,
+# readability-redundant-smartptr-get,
+# readability-redundant-string-cstr,
+# readability-simplify-boolean-expr,
+# readability-static-accessed-through-instance
+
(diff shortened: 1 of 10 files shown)


sjunges:
Nice. LGTM! I also think the inline comments are fine. 

Would be good if we have an auto-format- like script to also run this locally?

volkm:
You can run `make clang-tidy` to run this locally. But it runs then on the whole code base.
I can see whether it is possible to locally only run on specific or changed files.

sjunges:
Make clang-tidy is good for me. We have then make format and make clang-tidy. Maybe we should have a make style or so?

volkm:
Updated the main comment after the recent changes.

Two decisions:
- should we keep or remove the CMake options
- how should we handle the formatting (especially the versions) in the CI

volkm:
The clang-tidy script was generated by opencode (model: Big Pickle). I did not check every line in great detail.

sjunges:
I think the cmake commands are good as they reduce the number of tool dependencies we have on a minimal system?

sjunges:
LGTM.

volkm:
> I think the cmake commands are good as they reduce the number of tool dependencies we have on a minimal system?

The CMake part does not look for the clang-tidy/clang-format binaries. It just adds the targets and some additional CMake settings (for clang-tidy). The search for the binaries is done in the scripts. So even if there are no binaries, the CMake configuration will be successful but it will fail when running the actual script. I deliberately left all logic in the scripts such that they can also be used standalone.

I am open to discussing this design decision if we want to do it differently.

sjunges:
Ah, I see. Well, we could ask CMAKE to check whether these tools exist so that one gets cleaner error messages? Either way, I think having these options to disable it is not bad, even if we dont use that right now.

volkm:
Would we then duplicate the logic in CMake and the scripts? Because that is not so nice to maintain.
Alternatively, we could move the logic to CMake and only have an `--binary` option in the scripts which expects the right binary, but will not check again whether it is valid.

We could also merge it for now and open a separate discussion for this.

sjunges:
Hmm, lets just keep it as it is then :-)

PR #981, opened 2026-08-05
Replaced auto const with uint64_t in BitVector iterator

volkm:
Addresses a coding style raised in https://github.com/stormchecker/storm/issues/978

Having an automatic check for this (to prevent future issues) is not really possible with clang-tidy out-of-the-box. LLMs suggested to write a clang-tidy plugin to check this, but this might be overkill.


sjunges:
This is nice, but I would like to hold off this PR for a bit? 

I also dont think we need a plugin for this :-)

volkm:
@sjunges Can I merge it?

sjunges:
yes

PR #982, opened 2026-08-06
Removed STORM_DEV definition

volkm:


src/storm/modelchecker/reachability/SparseDtmcEliminationModelChecker.cpp
```diff
@@ -245,9 +245,7 @@ SparseDtmcEliminationModelChecker<SparseDtmcModelType>::computeLongRunValues(Env
     while (priorityQueue->hasNext()) {
         storm::storage::sparse::state_type state = priorityQueue->pop();
         stateEliminator.eliminateState(state, true);
-#ifdef STORM_DEV
         STORM_LOG_ASSERT(checkConsistent(flexibleMatrix, flexibleBackwardTransitions), "The forward and backward transition matrices became inconsistent.");
```
sjunges: Should we then also remove these expensive checks? They seem quite conservative?

volkm: Might indeed be the best idea.

sjunges:
LGTM.

PR #983, opened 2026-08-06
Add CVaR model checking for sparse MDP rewards

pjtimm:
## Summary

This PR adds scalar CVaR model checking for reward-reachability queries on sparse MDPs. It allows users to analyze the tail of a reward or cost distribution.

The weighted-reachability backend uses an LP for terminal-reward models. SSP minimum-cost and maximum-reward queries use Pareto front value iteration.

## Scope

Queries have the form

`R{"reward"}min/max=? [ F "target" ]`

together with `--cvar <alpha>`, where `0 < alpha < 1`.

The current implementation supports sparse MDPs with one initial state. Each backend checks its own assumptions on rewards, target states, and SSP structure and rejects unsupported models.

## Testing

Created tests in `CvarQueryTest` pass. They cover query parsing and validation, the weighted-reachability and SSP backends, cost and reward interpretations, preprocessing, exact rational inputs, and rejection of unsupported models.


tquatmann:
Thank you for the awesome contribution! As discussed offline, we will put this PR on hold while we continue working on a more complete/polished implementation.

PR #986, opened 2026-08-06
speed up compilation of parser, disable compile-time-profiling on parsers

sjunges:
Speed up the compilation time for the parsers by disabling debug info for them (looks horrible anyway).
Slightly hacky way to disable compile-time-profiling if it is switched on  on these compiliation units as they really cost a lot of overhead which is annoying when working on the compile times.


volkm (approved):
LGTM.

PR #987, opened 2026-08-06
make storm-pars compile much faster

sjunges:
Storm-pars was one of the remaining big compile-time-users and this PR improves significantly upon that situation, without changing the behavior of storm. 

Relates to https://github.com/stormchecker/storm/discussions/763.


src/storm-pars-cli/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ add_executable(storm-pars-cli ${PROJECT_SOURCE_DIR}/src/storm-pars-cli/storm-par
                               ${PROJECT_SOURCE_DIR}/src/storm-pars-cli/print.cpp
                               ${PROJECT_SOURCE_DIR}/src/storm-pars-cli/monotonicity.cpp
                               ${PROJECT_SOURCE_DIR}/src/storm-pars-cli/feasibility.cpp
```
volkm:
```suggestion
                              ${PROJECT_SOURCE_DIR}/src/storm-pars-cli/feasibility.cpp
                              ${PROJECT_SOURCE_DIR}/src/storm-pars-cli/monotonicity.cpp
                              ${PROJECT_SOURCE_DIR}/src/storm-pars-cli/print.cpp
```

volkm: Easier to maintain in alphabetical order.

src/storm-pars-cli/solutionFunctions.h
```diff
@@ -1,76 +1,54 @@
...
+extern template void verifyProperties<storm::RationalFunction>(
+    std::vector<storm::jani::Property> const&,
+    std::function<std::unique_ptr<storm::modelchecker::CheckResult>(std::shared_ptr<storm::logic::Formula const> const&)> const&,
+    std::function<void(std::unique_ptr<storm::modelchecker::CheckResult> const&)> const&);
+
+extern template void computeSolutionFunctionsWithSparseEngine<storm::RationalFunction>(
+    std::shared_ptr<storm::models::sparse::Model<storm::RationalFunction>> const&, storm::cli::SymbolicInput const&);
+
+extern template void computeSolutionFunctionsWithSymbolicEngine<storm::dd::DdType::Sylvan, storm::RationalFunction>(
+    std::shared_ptr<storm::models::symbolic::Model<storm::dd::DdType::Sylvan, storm::RationalFunction>> const&, storm::cli::SymbolicInput const&);
```
volkm, line 52: Is this needed if the templates are explicitly instantiated in the cpp file?

src/storm-pars/api/export.cpp
```diff
@@ -0,0 +1,53 @@
+#include <algorithm>
+#include <fstream>
+#include <iterator>
+#include <vector>
+
+#include "storm-pars/api/export.h"
+#include "storm/analysis/GraphConditions.h"
+#include "storm/io/file.h"
```
volkm, line 9:
```suggestion
#include "storm-pars/api/export.h"

#include <algorithm>
#include <fstream>
#include <iterator>
#include <vector>

#include "storm/analysis/GraphConditions.h"
#include "storm/io/file.h"
```

src/storm-pars/api/export.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+namespace storm {
```
volkm:
```suggestion
namespace storm::pars {
```

src/storm-pars/api/export.cpp
```diff
@@ -0,0 +1,53 @@
...
+        std::transform(constraintCollector->getGraphPreservingConstraints().begin(), constraintCollector->getGraphPreservingConstraints().end(),
+                       std::back_inserter(stringConstraints),
+                       [](carl::Formula<typename storm::Polynomial::PolyType> const& c) -> std::string { return c.toString(); });
+        std::copy(stringConstraints.begin(), stringConstraints.end(), std::ostream_iterator<std::string>(filestream, "\n"));
+    }
+    storm::io::closeFile(filestream);
+}
+
+}  // namespace api
+}  // namespace storm
```
volkm:
```suggestion
}  // namespace storm::pars
```

src/storm-pars/api/export.h
```diff
@@ -1,12 +1,18 @@
...
-#include "storm/analysis/GraphConditions.h"
+#include <optional>
+
+#include "storm/adapters/RationalFunctionAdapter.h"
 #include "storm/exceptions/NotSupportedException.h"
-#include "storm/io/file.h"
 #include "storm/utility/OptionalRef.h"
 #include "storm/utility/macros.h"
 
 namespace storm {
```
volkm:
```suggestion
namespace storm::pars {
```

src/storm-pars/api/export.h
```diff
@@ -15,42 +21,9 @@ void exportParametricResultToFile(std::optional<ValueType>, storm::OptionalRef<s
...
-        std::copy(stringConstraints.begin(), stringConstraints.end(), std::ostream_iterator<std::string>(filestream, "\n"));
-    }
-    storm::io::closeFile(filestream);
-}
+void exportParametricResultToFile(std::optional<storm::RationalFunction> result,
+                                  storm::OptionalRef<storm::analysis::ConstraintCollector<storm::RationalFunction> const> const& constraintCollector,
+                                  std::string const& path);
+
 }  // namespace api
 }  // namespace storm
```
volkm:
```suggestion
}  // namespace storm::pars
```

src/storm-pars/api/region.h
```diff
@@ -191,42 +188,11 @@ std::shared_ptr<storm::models::sparse::Model<ParametricType>> preprocessSparseMo
...
-                                                                                                          ImpreciseType, PreciseType>>();
-            }
-        default:
-            STORM_LOG_THROW(false, storm::exceptions::UnexpectedException, "Unexpected region model checker type.");
-    }
-    return nullptr;
-}
+                                                                                                  storm::models::ModelType modelType);
+
+extern template std::unique_ptr<storm::modelchecker::RegionModelChecker<storm::RationalFunction>>
```
volkm, line 193: Again, is this needed?

src/storm-pars/api/export.cpp
volkm: I see that this is currently not consistent in storm-pars anyway. So maybe it should be a separate PR.

volkm:
The main part looks good to me. I did not review the changes in the tests yet.

src/storm-pars/api/region.h
```diff
@@ -191,42 +188,11 @@ std::shared_ptr<storm::models::sparse::Model<ParametricType>> preprocessSparseMo
...
+createRegionModelChecker<storm::RationalFunction, double, storm::RationalNumber>(storm::modelchecker::RegionCheckEngine engine,
+                                                                                 storm::models::ModelType modelType);
```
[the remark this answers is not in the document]
sjunges: This is the same pattern all over the C++ API in storm. We will not resolve this with this PR.

src/storm-pars-cli/solutionFunctions.h
sjunges: Yes, because they are (as of now) superexpensive to compile. Let me add a comment about that.

src/storm-pars/api/region.h
sjunges: Yes, trying to reduce the compilation times here was mostly about compiling this once, not multiple times.

src/storm-pars-cli/solutionFunctions.h
volkm: Ok, I should read more about extern then. I though that by moving it to the cpp it would only be compiled once (when compiling the cpp).

sjunges: I think the cpp makes sure it compiles at least once, the extern tells anyone including the header that this is already compiled elsewhere in the lib?

volkm: That is also my understanding. But because the body is only in the cpp, other translation units cannot instantiate the header anyway. So in my understanding it should be possible to remove the external without affecting the compilation performance. But for documentation purposes I think it is good to keep it and indicate that we actually want it to be extern.

volkm: Maybe it also allows to change RationalFunctionAdapter to RationalFunctionForward in solutionFunctions.h and export.h?

sjunges: Yeah, I think you are also right about that it compiles without, but I like it this way for now.

Regarding the forward vs the adapter. Maybe, I can try, but it wont change ocmpile times, as anyting in parametric will need to know what rational functions are anyway.

volkm: I am fine with leaving things as is.

PR #990, opened 2026-08-07
one approach to skipping a few expensive tests

sjunges:
Very few tests are causing >80% of the test run time. If we disable them like this, our CI runs much faster.


volkm (approved):
I am fine with this for the moment.
We can improve later on if needed.

Note that there is already a [disabled test](https://github.com/stormchecker/storm/blob/master/src/test/storm-pars/transformer/BinaryDtmcTransformerTest.cpp#L136). This might still fail.

Issue #91, opened 2020-10-08
Difference results in bisimulation by using dd engine and sparse angine


sadeghrk2:
Hi,

When I use bisimulation for MDPs with dd and sparse engines the number of states of the reduced models are different for som case studies.

Several sample files are attached as follows:

sadeghrk2:
[coin_4_4_dd.log](https://github.com/moves-rwth/storm/files/5350539/coin_4_4_dd.log)
[coin_4_4_sparse.log](https://github.com/moves-rwth/storm/files/5350540/coin_4_4_sparse.log)
[coin_4_12_dd.log](https://github.com/moves-rwth/storm/files/5350542/coin_4_12_dd.log)
[coin_4_12_sparse.log](https://github.com/moves-rwth/storm/files/5350544/coin_4_12_sparse.log)
[wlan_3_200_dd.log](https://github.com/moves-rwth/storm/files/5350545/wlan_3_200_dd.log)
[wlan_3_200_sparse.log](https://github.com/moves-rwth/storm/files/5350546/wlan_3_200_sparse.log)
[wlan_4_300_dd.log](https://github.com/moves-rwth/storm/files/5350548/wlan_4_300_dd.log)
[wlan_4_300_sparse.log](https://github.com/moves-rwth/storm/files/5350549/wlan_4_300_sparse.log)
[zeroconf_3_dd.log](https://github.com/moves-rwth/storm/files/5350550/zeroconf_3_dd.log)
[zeroconf_3_sparse.log](https://github.com/moves-rwth/storm/files/5350551/zeroconf_3_sparse.log)
[zeroconf_4_dd.log](https://github.com/moves-rwth/storm/files/5350552/zeroconf_4_dd.log)
[zeroconf_4_sparse.log](https://github.com/moves-rwth/storm/files/5350553/zeroconf_4_sparse.log)

volkm:
We can reproduce this issue in the current master branch.

A minimal example is the following:
##### Sparse
`storm --prism ~/prism-benchmarks/models/mdps/wlan/wlan1.nm  --prop "Pmax=? [F col=COL]" -const "COL=1" -bisim`
yields:
```
Model type: 	MDP (sparse)
States: 	2
Transitions: 	2
Choices: 	2
```


##### DD
`storm --prism ~/prism-benchmarks/models/mdps/wlan/wlan1.nm  --prop "Pmax=? [F col=COL]" -const "COL=1" -bisim -e dd-to-sparse`
yields:
[...]

PR #991, opened 2026-08-07
deduplicate same choices when building an explicit quotient.

sjunges:
Makes it easier to ensure bisim and dd-to-bisim coincide, makes the MDP a bit smaller, and doesnt cost much. 
 
 Relates to #91; this is only part of the solution, the main 'problem' lies on the sparse bisim side.

### DIFF
--- a/src/storm/storage/dd/bisimulation/QuotientExtractor.cpp
+++ b/src/storm/storage/dd/bisimulation/QuotientExtractor.cpp
@@ -321,9 +321,10 @@ class InternalSparseQuotientExtractorBase {
             STORM_LOG_ASSERT(!this->rowPermutation.empty(), "Expected proper row permutation.");
             std::vector<ExportValueType> valueVector = extractVectorInternal(vector, this->allSourceVariablesCube, this->nondeterminismOdd);
 
-            // Reorder the values according to the known row permutation.
-            std::vector<ExportValueType> reorderedValues(valueVector.size());
-            for (uint64_t pos = 0; pos < valueVector.size(); ++pos) {
+            // Reorder the values according to the known row permutation. Note that rowPermutation may be shorter
+            // than valueVector if redundant (duplicate) choices were removed while building the transition matrix.
+            std::vector<ExportValueType> reorderedValues(rowPermutation.size());
+            for (uint64_t pos = 0; pos < rowPermutation.size(); ++pos) {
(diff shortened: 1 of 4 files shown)


src/storm/storage/dd/bisimulation/QuotientExtractor.cpp
```diff
@@ -358,11 +359,27 @@ class InternalSparseQuotientExtractorBase {
...
+            // Remove choices that became exact duplicates of the (in the sorted order) preceding choice of the
+            // same state after quotienting, i.e. redundant nondeterminism that resulted from collapsing states
+            // into the same block. This mirrors what the sparse bisimulation engine does and keeps the two
+            // engines' quotients comparable.
+            std::vector<uint64_t> deduplicatedPermutation;
+            deduplicatedPermutation.reserve(rowPermutation.size());
+            for (uint64_t i = 0; i < rowPermutation.size(); ++i) {
+                uint64_t const rowIdx = rowPermutation[i];
+                bool const isDuplicate =
+                    i > 0 && rowToState[rowPermutation[i - 1]] == rowToState[rowIdx] && matrixEntries[rowPermutation[i - 1]] == matrixEntries[rowIdx];
```
tquatmann: Does this distinguish choices with different rewards but equal transition probabilities?

src/storm/storage/dd/bisimulation/QuotientExtractor.cpp
```diff
@@ -1009,6 +1075,19 @@ std::shared_ptr<storm::models::sparse::Model<ExportValueType>> QuotientExtractor
                      "Representatives do not cover all blocks.");
     InternalSparseQuotientExtractor<DdType, ValueType, ExportValueType> sparseExtractor(model, partitionAsBdd, partition.getBlockVariable(),
                                                                                         partition.getNumberOfBlocks(), representatives);
+
+    // Collect the state-action rewards of the preserved reward models upfront, so that the extraction of the
```
sjunges, line 1079: Is this reused later? Where were these rewards previously collected?

src/storm/storage/dd/bisimulation/QuotientExtractor.cpp
sjunges: it didnt, thanks

PR #997, opened 2026-08-09
fix for absorbing states in state elim

### LINKED ISSUE #86: Elimination with Policy Iteration potentially broken
Hi,

when trying to combine policy iteration with state elimination, I run into the following bug. Is this combination of methods somehow not a good idea, or does this force the elimination solver in a corner case?

Best,
Sebastian

```

Command line arguments: -drn unrolling.out --prop 'Pmax=? [F "_goal"]' '--minmax:method' pi --eqsolver elimination

Time for model construction: 0.004s.

Model type: 	MDP (sparse)
States: 	17
Transitions: 	36
Choices: 	20
Reward Models:  none
State Labels: 	2 labels
   * init -> 1 item(s)
   * _goal -> 1 item(s)
Choice Labels: 	none

Model checking property "1": Pmax=? [F "_goal"] ...
ERROR (EliminatorBase.cpp:51): The scaling mode 'divide-one-minus' requires a non-one value in the given column.
Assertion failed: (columnValue != storm::utility::one<ValueType>()), function eliminate, file storm/solver/stateelimination/EliminatorBase.cpp, line 51.
ERROR: The program received signal 6 and will be aborted in 3s.
[1]    85131 abort     storm -drn unrolling.out --prop "Pmax=? [F \"_goal\"]"  
```

sjunges:
Fixes underlying issue of #86 (which I cannot reproduce from cli, i would say it resolves #86)

Absorbing states should simply get a solution of 0, as we solve for the least fixed point in these problems. This is in line with what we do in other places.

### DIFF
--- a/src/storm/solver/EliminationLinearEquationSolver.cpp
+++ b/src/storm/solver/EliminationLinearEquationSolver.cpp
@@ -49,9 +49,10 @@ void EliminationLinearEquationSolver<ValueType>::setMatrix(storm::storage::Spars
 template<typename ValueType>
 bool EliminationLinearEquationSolver<ValueType>::internalSolveEquations(Environment const& env, std::vector<ValueType>& x,
                                                                         std::vector<ValueType> const& b) const {
-    // FIXME: This solver will not work for all input systems. More concretely, the current implementation will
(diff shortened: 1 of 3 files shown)


volkm (approved):
LGTM.

PR #1008, opened 2026-08-12
Fixed bugs revealed by clang-tidy

volkm:
Please take a second look to check whether they now behave as expected.

The check `stateFormula.isBooleanLiteralFormula` occurred twice.

### DIFF
--- a/src/storm/modelchecker/prctl/helper/HybridMdpPrctlHelper.cpp
+++ b/src/storm/modelchecker/prctl/helper/HybridMdpPrctlHelper.cpp
@@ -300,7 +300,7 @@ std::unique_ptr<CheckResult> HybridMdpPrctlHelper<DdType, ValueType>::computeGlo
     Environment const& env, OptimizationDirection dir, storm::models::symbolic::NondeterministicModel<DdType, ValueType> const& model,
     storm::dd::Add<DdType, ValueType> const& transitionMatrix, storm::dd::Bdd<DdType> const& psiStates, bool qualitative) {
     std::unique_ptr<CheckResult> result =
-        computeUntilProbabilities(env, dir == OptimizationDirection::Minimize ? OptimizationDirection::Maximize : OptimizationDirection::Maximize, model,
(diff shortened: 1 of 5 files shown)


src/storm/modelchecker/prctl/helper/HybridMdpPrctlHelper.cpp
```diff
@@ -300,7 +300,7 @@ std::unique_ptr<CheckResult> HybridMdpPrctlHelper<DdType, ValueType>::computeGlo
     Environment const& env, OptimizationDirection dir, storm::models::symbolic::NondeterministicModel<DdType, ValueType> const& model,
     storm::dd::Add<DdType, ValueType> const& transitionMatrix, storm::dd::Bdd<DdType> const& psiStates, bool qualitative) {
     std::unique_ptr<CheckResult> result =
-        computeUntilProbabilities(env, dir == OptimizationDirection::Minimize ? OptimizationDirection::Maximize : OptimizationDirection::Maximize, model,
+        computeUntilProbabilities(env, dir == OptimizationDirection::Minimize ? OptimizationDirection::Maximize : OptimizationDirection::Minimize, model,
```
sjunges, line 303: Is there potentially some negation here for the globally into a until? Do we have tests for globally operators? Should we?

volkm: I was unsure myself and looked at how the call is for [SymbolicMdpPrctlHelper](https://github.com/volkm/storm/blob/592f8a0637aff1b5b02e368c2eb797bc4e571b40/src/storm/modelchecker/prctl/helper/SymbolicMdpPrctlHelper.cpp#L152].
I only skimmed MdpPrctlModelCheckerTest.cpp but did not see a test.

sjunges: That convinces me. I will open an issue to test globally properties.

PR #1014, opened 2026-08-14
Fixed issues with use-after-move

volkm:
Revealed by clang-tidy with bugprone-use-after-move.

Also enabled some more warnings (which are already satisfied on the code base) and fixed clang-tidy scripts such that relative paths are handled correctly now.

### DIFF
--- a/src/storm-pars/modelchecker/region/monotonicity/MonotonicityResult.cpp
+++ b/src/storm-pars/modelchecker/region/monotonicity/MonotonicityResult.cpp
@@ -18,7 +18,7 @@ MonotonicityResult<VariableType>::MonotonicityResult() {
 
 template<typename VariableType>
 void MonotonicityResult<VariableType>::addMonotonicityResult(VariableType var, MonotonicityResult<VariableType>::Monotonicity mon) {
-    monotonicityResult.insert(std::pair<VariableType, MonotonicityResult<VariableType>::Monotonicity>(std::move(var), std::move(mon)));
+    monotonicityResult.insert(std::pair<VariableType, MonotonicityResult<VariableType>::Monotonicity>(std::move(var), mon));
 }
 
 template<typename VariableType>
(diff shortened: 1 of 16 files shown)


src/storm/modelchecker/helper/indefinitehorizon/visitingtimes/SparseDeterministicVisitingTimesHelper.cpp
```diff
@@ -471,6 +471,8 @@ std::vector<ValueType> SparseDeterministicVisitingTimesHelper<ValueType>::comput
     }
 
     // Get the solver object and satisfy requirements
+    // The solver consumes the matrix, so the acyclic check must be performed before moving it.
+    bool const hasCycles = storm::utility::graph::hasCycle(sccMatrix);
```
volkm: This check is now always performed. Is this an acceptable performance impact? Or do you have a better suggestion how to handle this part?

src/storm/models/sparse/Model.cpp
```diff
@@ -585,7 +585,7 @@ void Model<ValueType, RewardModelType>::writeJsonToStream(std::ostream& outStrea
                 }
             }
             if (!choiceRewardsJson.empty()) {
-                choiceRewardsJson["rew"] = std::move(choiceRewardsJson);
+                choiceJson["rew"] = std::move(choiceRewardsJson);
```
volkm, line 588: This was incorrect before, but I guess the function was seldom used.

src/storm-pars/modelchecker/region/monotonicity/MonotonicityResult.cpp
sjunges, line 1: Relates to #976 which overhauls this anyways.

src/storm/modelchecker/helper/indefinitehorizon/visitingtimes/SparseDeterministicVisitingTimesHelper.cpp
tquatmann: We can get the requirements without building the solver like this:
```suggestion
    auto req = linearEquationSolverFactory.getRequirements(env);
    if (req.acyclic().isCritical()) {
        // The solver consumes the matrix, so the acyclic check must be performed before moving it.
        STORM_LOG_THROW(!storm::utility::graph::hasCycle(sccMatrix), storm::exceptions::UnmetRequirementException,
                        "The solver requires an acyclic model, but the model is not acyclic.");
        req.clearAcyclic();
    }
```

tquatmann (approved):
Great fixes! Thanks! The visiting times overhead can be avoided, see my suggestion :)

src/storm/modelchecker/helper/indefinitehorizon/visitingtimes/SparseDeterministicVisitingTimesHelper.cpp
volkm: Thanks for the suggestion. I have integrated it.

PR #1016, opened 2026-08-14
merging of equivalent states as preprocessing

tquatmann:
* extends the GoalStateMerger so that it can now merge states based on an input formula (e.g. all prob0/prob1 states
* adds a CLI option --merge-equivalent-states to invoke that as a preprocessing
* Simplifies SparseParametricModelSimplifier by calling the new GoalStateMerger

Relates to https://github.com/stormchecker/storm/pull/996


src/storm-pars/transformer/SparseParametricModelSimplifier.cpp
```diff
@@ -117,14 +47,14 @@ std::shared_ptr<SparseModelType> SparseParametricModelSimplifier<SparseModelType
...
-            (!rewardModelName.is_initialized() || storm::utility::isConstant(actionRewards[sparseMatrix.getRowGroupIndices()[state]]))) {
+            (!rewardModelName.has_value() || storm::utility::isConstant(actionRewards[sparseMatrix.getRowGroupIndices()[state]]))) {
             for (auto const& entry : sparseMatrix.getRowGroup(state)) {
                 if (!storm::utility::isConstant(entry.getValue())) {
                     selectedStates.set(state, false);
                     break;
                 }
             }
-            if (state && this->preserveParametricTransitions) {
+            if (state && preserveParametricTransitions) {
```
tquatmann: This line (which was there before) confused me. Should it say `if (selectedStates.get(state) && preserveParametricTransitions)` ? @linusheck

linusheck: Thanks for catching this! It should just say `if (preserveParametricTransitions)`. I guess the bug is that if `state==0` this is broken :D

tquatmann:
Currently relying on #1028 being merged first.

src/storm/api/transformation.h
```diff
@@ -146,6 +150,28 @@ std::shared_ptr<storm::models::sparse::Model<ValueType>> transformToNondetermini
     }
 }
 
+/*!
+ * Identifies states of the given model and formula that are equivalent and merges them into a single state, yielding a (potentially smaller) model on which the
```
volkm:
```suggestion
 * Identifies states of the given model that are equivalent w.r.t. to the given formula and merges them into a single state, yielding a (potentially smaller) model on which the
```

src/storm/api/transformation.h
```diff
@@ -146,6 +150,28 @@ std::shared_ptr<storm::models::sparse::Model<ValueType>> transformToNondetermini
...
+ * given formula can still be checked.
+ * @note depending on the formula, this merges, e.g., states with probability 0/1 or reward 0/infty based on graph-based analysis.
+ * @return the resulting model, or nullptr if merging is not applicable (e.g. because the model type or the formula is not supported).
+ */
+template<typename ValueType>
+std::shared_ptr<storm::models::sparse::Model<ValueType>> mergeEquivalentStatesForFormula(std::shared_ptr<storm::models::sparse::Model<ValueType>> const& model,
+                                                                                         storm::logic::Formula const& formula) {
+    // storm::transformer::GoalStateMerger is only instantiated for double, rational number, and rational function value types.
```
volkm: If the instantiations are limited, maybe move the function to a cpp and explicitly instantiate it. This would avoid including `RationalFunctionForward` (which might be slightly better for build times?).

src/storm/api/transformation.h
```diff
@@ -146,6 +150,28 @@ std::shared_ptr<storm::models::sparse::Model<ValueType>> transformToNondetermini
...
+    if constexpr (std::is_same_v<ValueType, double> || std::is_same_v<ValueType, storm::RationalNumber> || std::is_same_v<ValueType, storm::RationalFunction>) {
+        if (model->isOfType(storm::models::ModelType::Dtmc) || model->isOfType(storm::models::ModelType::Ctmc) ||
+            model->isOfType(storm::models::ModelType::Mdp) || model->isOfType(storm::models::ModelType::MarkovAutomaton)) {
```
volkm: Why not throw a `NotSupportedException` instead of returning a nullptr?

src/storm-cli-utilities/model-handling.h
```diff
@@ -735,6 +735,21 @@ std::pair<std::shared_ptr<storm::models::ModelBase>, bool> preprocessModel(std::
...
     }
 
+    // Merging of states should be done before applying bisimulation as this order leads to the smallest quotient
+    if (transformationSettings.isMergeEquivalentStatesSet()) {
+        auto formulas = createFormulasToRespect(input.properties);
+        if (formulas.size() != 1) {
+            STORM_LOG_WARN("Skipping merging of equivalent states as it requires exactly one input property. " << formulas.size()
+                                                                                                               << " properties given instead.");
+        } else if (auto mergedModel = storm::api::mergeEquivalentStatesForFormula<ValueType>(result.first, *formulas.front())) {
+            STORM_LOG_INFO("Merged target/sink states relevant for the considered property '" << *formulas.front() << "'.");
```
volkm: I know this works, but I personally dislike assignments in conditions. Could we maybe explicitly check for the nullptr?

src/storm-pars/transformer/SparseParametricModelSimplifier.cpp
```diff
@@ -8,99 +8,29 @@
...
 
+/*!
+ * Eliminates all states that satisfy
+ * * there is only one enabled action (i.e., there is no nondeterministic choice at the state),
+ * * all outgoing transitions are constant,
+ * * there is no statelabel defined, and
+ * * (if rewardModelName is given) the reward collected at the state is constant.
+ *
+ * The resulting model will only have the rewardModel with the provided name (or no reward model at all if no name was given).
+ * Labelings of eliminated states will be lost
```
volkm, line 29: This contradictory when elimination only happens if no state label is defined. Do you mean choice labels here?

src/storm-pars/transformer/SparseParametricModelSimplifier.cpp
```diff
@@ -159,6 +89,215 @@ std::shared_ptr<SparseModelType> SparseParametricModelSimplifier<SparseModelType
...
+ * The resulting model will only have the rewardModel with the provided name (or no reward model at all if no name was given)
+ */
+template<typename SparseModelType>
+std::shared_ptr<SparseModelType> eliminateNeutralEndComponents(SparseModelType const& model, storm::storage::BitVector const& ignoredStates,
+                                                               std::optional<std::string> const& rewardModelName) {
+    // Get the actions that can be part of an EC
+    storm::storage::BitVector possibleECActions(model.getNumberOfChoices(), true);
+    for (auto state : ignoredStates) {
+        for (uint_fast64_t actionIndex = model.getTransitionMatrix().getRowGroupIndices()[state];
+             actionIndex < model.getTransitionMatrix().getRowGroupIndices()[state + 1]; ++actionIndex) {
```
volkm: Use `getRowGroupIndices(state)`?

src/storm-pars/transformer/SparseParametricModelSimplifier.cpp
```diff
@@ -159,6 +89,215 @@ std::shared_ptr<SparseModelType> SparseParametricModelSimplifier<SparseModelType
...
+            simplifiedFormula = std::make_shared<storm::logic::ProbabilityOperatorFormula const>(eventuallyFormula, operatorFormula.getOperatorInformation());
+            enableStateElimination = true;
+            enableEndComponentElimination = originalModel.isNondeterministicModel() && !minimizing;  // end components can only exist for Pmax queries.
+        } else if (operatorFormula.getSubformula().isReachabilityProbabilityFormula()) {
+            enableStateElimination = true;
+            enableEndComponentElimination = originalModel.isNondeterministicModel() && !minimizing;  // end components can only exist for Pmax queries.
+        } else if (operatorFormula.getSubformula().isBoundedUntilFormula()) {
+            // multidimensional or reward bounded until formulas are not considered in this simplifier.
+            if (auto const& boundedUntilFormula = operatorFormula.getSubformula().asBoundedUntilFormula();
+                !boundedUntilFormula.isMultiDimensional() && !boundedUntilFormula.getTimeBoundReference().isRewardBound()) {
```
volkm: I find this not so nice to read. Maybe move the assignment outside the if? Should not have an important performance impact.

src/storm-pars/transformer/SparseParametricModelSimplifier.cpp
```diff
@@ -159,6 +89,215 @@ std::shared_ptr<SparseModelType> SparseParametricModelSimplifier<SparseModelType
...
+                // Since we have a discrete-time model, we may assume a step bound.
```
volkm, line 201: Can we maybe assert the discrete-time model somewhere? Would be more robust in case we want to look at continuous-time models at some point.

src/test/storm/transformer/GoalStateMergerTest.cpp
```diff
@@ -0,0 +1,121 @@
+#include <cmath>
+
+#include "storm-config.h"
+#include "test/storm_gtest.h"
+
+#include "storm-parsers/api/storm-parsers.h"
+#include "storm-parsers/parser/PrismParser.h"
+#include "storm/api/storm.h"
```
volkm: If possible, try to include the dedicated api headers in `storm/api`.

src/test/storm/transformer/GoalStateMergerTest.cpp
```diff
@@ -0,0 +1,121 @@
...
+                                              << " (dropUnreachableFromInit=" << dropUnreachableFromInit << ").";
+        ASSERT_EQ(1ull, mergerResult->model->getInitialStates().getNumberOfSetBits())
+            << "For merged model of " << prismModelFile << " (dropUnreachableFromInit=" << dropUnreachableFromInit << ").";
+
+        // The original (unmodified) formula must still be valid on the merged model and induce the same value at the initial state.
+        auto const mergedVal = computeValue(mergerResult->model);
+        bool const ok = modelVal == 0.0 ? mergedVal == 0 : std::abs((mergedVal - modelVal) / modelVal) <= 1e-6;
+        EXPECT_TRUE(ok) << "Relative difference between original model result (" << modelVal << ") and merged model result (" << mergedVal
+                        << ") is too high.\nFailed for model " << prismModelFile << " and formula " << *formulas.front()
+                        << " (dropUnreachableFromInit=" << dropUnreachableFromInit << ").";
```
volkm: You could also distinguish between modelVal == 0.0 and else, and then use `EXPECT_NEAR(mergedVal, modelVal, 1e-6)` (but this uses absolute precision)

src/storm/transformer/GoalStateMerger.cpp
```diff
@@ -2,68 +2,432 @@
...
+    if (!checker.canHandle(subformula)) {
+        return std::nullopt;
+    }
+    return std::move(checker.check(subformula)->template asExplicitQualitativeCheckResult<typename SparseModelType::ValueType>().getTruthValuesVector());
+}
+
+std::optional<uint64_t> representative(storm::storage::BitVector const& b) {
+    if (auto it = b.begin(); it != b.end()) {
+        return *it;
+    }
```
volkm, line 62: Maybe use `getNextSetIndex(0)`? Or check for `empty` and otherwise return `b.begin()`? The current versions is not immediately understandable.

src/storm/transformer/GoalStateMerger.cpp
```diff
@@ -2,68 +2,432 @@
...
+    } else {
+        STORM_LOG_ASSERT(originalModel.isOfType(Dtmc) || originalModel.isOfType(Mdp), "Unhandled model type: " << originalModel.getType());
+    }
 
-    result.first.model = buildOutputModel(maybeStates, result.first, std::move(transitionMatrix), std::move(labeling), std::move(rewardModels));
+    // Reapply state valuations, choice labeling, choice origins if available
+    if (originalModel.hasStateValuations()) {
+        std::vector<uint64_t> newToOldStateIndices(maybeStates.begin(), maybeStates.end());
+        if (result.first.targetState.has_value()) {
+            STORM_LOG_ASSERT(result.first.targetState.value() == newToOldStateIndices.size(), "unexpected position of target state.");
```
volkm:
```suggestion
            STORM_LOG_ASSERT(result.first.targetState.value() == newToOldStateIndices.size(), "Unexpected position of target state.");
```

src/storm/transformer/GoalStateMerger.cpp
```diff
@@ -2,68 +2,432 @@
...
+            newToOldStateIndices.push_back(representativeTargetState.value_or(*targetStates.begin()));
+        }
+        if (result.first.sinkState.has_value()) {
+            STORM_LOG_ASSERT(result.first.sinkState.value() == newToOldStateIndices.size(), "unexpected position of sink state.");
```
volkm:
```suggestion
            STORM_LOG_ASSERT(result.first.sinkState.value() == newToOldStateIndices.size(), "Unexpected position of sink state.");
```

src/storm/transformer/GoalStateMerger.cpp
```diff
@@ -2,68 +2,432 @@
...
+    if (!phiStates || !psiStates) {
+        return std::nullopt;
+    }
+
+    auto const& transitions = originalModel.getTransitionMatrix();
+    storm::storage::SparseMatrix<ValueType> const backwardTransitions = originalModel.getBackwardTransitions();
+    std::pair<storm::storage::BitVector, storm::storage::BitVector> statesWithProbability01;
+    if (!originalModel.isNondeterministicModel()) {
+        statesWithProbability01 = storm::utility::graph::performProb01(backwardTransitions, *phiStates, *psiStates);
+    } else if (std::optional<bool> const minimizing = isMinimizing(formula); !minimizing.has_value()) {
```
volkm, line 191: `!isMinimizing(formula).has_value()` would not work?

volkm (approved):
LGTM.

I only have minor mostly code-style comments. Feel free to ignore some of them.

src/storm-pars/transformer/SparseParametricModelSimplifier.cpp
tquatmann: The comment was outdated. `there is no statelabel defined` is replaced by `is in consideredStates`

src/storm-pars/transformer/SparseParametricModelSimplifier.cpp
tquatmann: Added a static_assert

src/storm/api/transformation.h
tquatmann: `GoalStateMerger::mergeForFormula` also returns a nullptr if merging is not applicable (e.g. because the formula is not handled).

The idea is that we can just try simplifying the model through this API function and if that is not applicable, the caller can just fall back to the original model. The `SparseParametricModelSimplifier` does something similar. Alternative solutions would be

-  a `canHandle`-like function that has to detect exactly those cases where merging is possible (additional code needed, possible inconsistencies) or
- throw and catch a `NotSupportedException` (we usually avoid exceptions for control flow).

I'd prefer the nullptr variant as is.

src/storm/transformer/GoalStateMerger.cpp
tquatmann: It would work, but it would require two calls to `isMinimizing`

```
if (isMinimizing(formula).has_value()) {
    ...
} else if (isMinimizing(formula).value()) {
    ...
} else {
    ...
]
```

I prefer the current way (but no hard feelings, if you insist :) )

src/test/storm/transformer/GoalStateMergerTest.cpp
tquatmann: I simplified this a bit.

We use relative precision by default, so I would like to keep the check for relative precision.

src/storm/transformer/GoalStateMerger.cpp
volkm: I did not see the implicit second call. Then it makes sense.

PR #1025, opened 2026-08-18
eliminate more settings

### LINKED ISSUE #968: StandardGameSolver: Exact solution based on value type?
The StandardGameSolver uses this line in its constructors. 

```
linearEquationSolverIsExact = storm::settings::getModule<storm::settings::modules::GeneralSettings>().isExactSet();
```

to determine whether it can solve linear equations exactly. I would like to replace this by simply looking at the value type? 

@tquatmann  Could you confirm that that is an improvement?

sjunges:
Exploration now has its own environment. A game solver is exact if it uses exact numbers.

### DIFF
--- a/src/storm-dft/builder/ExplicitDFTModelBuilder.cpp
+++ b/src/storm-dft/builder/ExplicitDFTModelBuilder.cpp
@@ -9,6 +9,7 @@
 #include "storm/models/sparse/Ctmc.h"
 #include "storm/models/sparse/MarkovAutomaton.h"
 #include "storm/settings/SettingsManager.h"
+#include "storm/settings/modules/GeneralSettings.h"
 #include "storm/transformer/NonMarkovianChainTransformer.h"
 #include "storm/utility/ProgressMeasurement.h"
 #include "storm/utility/SignalHandler.h"
@@ -366,7 +367,8 @@ template<typename ValueType, typename StateType>
 void ExplicitDFTModelBuilder<ValueType, StateType>::exploreStateSpace(double approximationThreshold) {
     size_t nrExpandedStates = 0;
     size_t nrSkippedStates = 0;
-    storm::utility::ProgressMeasurement progress("explored states");
+    storm::utility::ProgressMeasurement progress("explored states",
+                                                 storm::settings::getModule<storm::settings::modules::GeneralSettings>().getShowProgressDelay());
     progress.startNewMeasurement(0);
     // TODO: do not empty queue every time but break before
     while (!explorationQueue.empty()) {
--- a/src/storm-dft/builder/ExplicitDFTModelBuilder.cpp
+++ b/src/storm-dft/builder/ExplicitDFTModelBuilder.cpp
@@ -9,7 +9,6 @@
 #include "storm/models/sparse/Ctmc.h"
 #include "storm/models/sparse/MarkovAutomaton.h"
 #include "storm/settings/SettingsManager.h"
-#include "storm/settings/modules/GeneralSettings.h"
 #include "storm/transformer/NonMarkovianChainTransformer.h"
 #include "storm/utility/ProgressMeasurement.h"
 #include "storm/utility/SignalHandler.h"
@@ -367,8 +366,7 @@ template<typename ValueType, typename StateType>
 void ExplicitDFTModelBuilder<ValueType, StateType>::exploreStateSpace(double approximationThreshold) {
     size_t nrExpandedStates = 0;
     size_t nrSkippedStates = 0;
-    storm::utility::ProgressMeasurement progress("explored states",
-                                                 storm::settings::getModule<storm::settings::modules::GeneralSettings>().getShowProgressDelay());
+    storm::utility::ProgressMeasurement progress("explored states");
     progress.startNewMeasurement(0);
     // TODO: do not empty queue every time but break before
     while (!explorationQueue.empty()) {
--- a/src/storm/environment/exploration/ExplorationEnvironment.h
+++ b/src/storm/environment/exploration/ExplorationEnvironment.h
@@ -1,0 +1,40 @@
+#pragma once
+
+#include <boost/optional.hpp>
+#include <cstdint>
+
+#include "storm/modelchecker/exploration/ExplorationTypes.h"
+
+namespace storm {
+
+class ExplorationEnvironment {
+   public:
+    ExplorationEnvironment();
+    ~ExplorationEnvironment();
+    ExplorationEnvironment(ExplorationEnvironment const& other);
+    ExplorationEnvironment& operator=(ExplorationEnvironment const& other);
+
+    storm::modelchecker::exploration_detail::PrecomputationType getPrecomputationType() const;
+    void setPrecomputationType(storm::modelchecker::exploration_detail::PrecomputationType value);
+
+    uint64_t getStepsUntilPrecomputation() const;
+    void setStepsUntilPrecomputation(uint64_t value);
+
+    boost::optional<uint64_t> const& getSampledPathsUntilPrecomputation() const;
+    void setSampledPathsUntilPrecomputation(uint64_t value);
+    void unsetSampledPathsUntilPrecomputation();
+
+    storm::modelchecker::exploration_detail::NextStateHeuristic getNextStateHeuristic() const;
+    void setNextStateHeuristic(storm::modelchecker::exploration_detail::NextStateHeuristic value);
+
+    double getPrecision() const;
+    void setPrecision(double value);
+
+   private:
+    storm::modelchecker::exploration_detail::PrecomputationType precomputationType;
+    uint64_t stepsUntilPrecomputation;
+    boost::optional<uint64_t> sampledPathsUntilPrecomputation;
+    storm::modelchecker::exploration_detail::NextStateHeuristic nextStateHeuristic;
+    double precision;
+};
+}  // namespace storm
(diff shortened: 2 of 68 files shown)


src/storm-dft/builder/ExplicitDFTModelBuilder.cpp
```diff
@@ -366,7 +367,8 @@ template<typename ValueType, typename StateType>
 void ExplicitDFTModelBuilder<ValueType, StateType>::exploreStateSpace(double approximationThreshold) {
     size_t nrExpandedStates = 0;
     size_t nrSkippedStates = 0;
-    storm::utility::ProgressMeasurement progress("explored states");
+    storm::utility::ProgressMeasurement progress("explored states",
+                                                 storm::settings::getModule<storm::settings::modules::GeneralSettings>().getShowProgressDelay());
```
volkm: I see why you introduced the delay in ProgressMeasurement, but now I have again a dependency on GeneralSettings in storm-dft :).
How would I get rid of it? As part of the builder environment? Or should there be an output environment?
Or could we just set a magic number here? Does this really need to be configurable?

src/storm/environment/solver/SolverEnvironment.h
```diff
@@ -63,6 +63,10 @@ class SolverEnvironment {
     void setForceExact(bool value);
     bool isDebugSet() const;
     void setDebug(bool value);
+    bool isVerboseSet() const;
+    void setVerbose(bool value);
+    uint64_t getShowProgressDelay() const;
+    void setShowProgressDelay(uint64_t delay);
```
volkm, line 69: Should this really be part of the SolverEnvironment? I would imagine that modelcheckers would also use this. Maybe add a general OutputEnvironment?

src/storm/environment/modelchecker/ModelCheckerEnvironment.h
```diff
@@ -32,10 +34,21 @@ class ModelCheckerEnvironment {
     void setLtl2daTool(std::string const& value);
     void unsetLtl2daTool();
 
+    bool isFilterRewZeroSet() const;
+    void setFilterRewZero(bool value);
+
+    bool isExportCdfSet() const;
+    void setExportCdf(bool value);
+    std::string const& getExportCdfDirectory() const;
+    void setExportCdfDirectory(std::string const& value);
```
volkm, line 41: Should this be part of an OutputEnvironment maybe?

src/storm/environment/modelchecker/ModelCheckerEnvironment.h
```diff
@@ -17,6 +17,8 @@ class ModelCheckerEnvironment {
    public:
     ModelCheckerEnvironment();
     ~ModelCheckerEnvironment();
+    ModelCheckerEnvironment(ModelCheckerEnvironment const& other);
+    ModelCheckerEnvironment& operator=(ModelCheckerEnvironment const& other);
```
volkm: Environments are typically part of the SubEnvironment which has the copy-constructors?
Or differently, if we really need these operators here, they should probably be added to all other environments as well?

volkm: Maybe also related to https://github.com/stormchecker/storm/blob/master/src/storm/environment/Environment.h#L12

src/storm/environment/exploration/ExplorationEnvironment.h
```diff
@@ -0,0 +1,40 @@
...
+    ExplorationEnvironment(ExplorationEnvironment const& other);
+    ExplorationEnvironment& operator=(ExplorationEnvironment const& other);
+
+    storm::modelchecker::exploration_detail::PrecomputationType getPrecomputationType() const;
+    void setPrecomputationType(storm::modelchecker::exploration_detail::PrecomputationType value);
+
+    uint64_t getStepsUntilPrecomputation() const;
+    void setStepsUntilPrecomputation(uint64_t value);
+
+    boost::optional<uint64_t> const& getSampledPathsUntilPrecomputation() const;
```
volkm: std::optional?

src/storm/modelchecker/prctl/helper/rewardbounded/MultiDimensionalRewardUnfolding.cpp
```diff
@@ -410,10 +408,8 @@ EpochModel<ValueType, SingleObjectiveMode>& MultiDimensionalRewardUnfolding<Valu
     if (!currentEpoch || !epochManager.compareEpochClass(epoch, currentEpoch.get())) {
         setCurrentEpochClass(epoch);
         epochModel.epochMatrixChanged = true;
-        if (storm::settings::getModule<storm::settings::modules::CoreSettings>().isShowStatisticsSet()) {
-            if (storm::utility::graph::hasCycle(epochModel.epochMatrix)) {
-                std::cout << "Epoch model for epoch " << epochManager.toString(epoch) << " is cyclic.\n";
-            }
+        if (storm::utility::graph::hasCycle(epochModel.epochMatrix)) {
+            STORM_LOG_STATISTICS("Epoch model for epoch " << epochManager.toString(epoch) << " is cyclic.");
```
volkm: Use STORM_LOG_STATISTICS_LAZY and move hasCycle inside to only perform the check if really needed?

src/storm/modelchecker/prctl/helper/SparseMdpPrctlHelper.cpp
```diff
@@ -1119,9 +1116,10 @@ QualitativeStateSetsReachabilityRewards getQualitativeStateSetsReachabilityRewar
 
 template<typename ValueType, typename SolutionType>
 QualitativeStateSetsReachabilityRewards computeQualitativeStateSetsReachabilityRewards(
-    storm::solver::SolveGoal<ValueType, SolutionType> const& goal, storm::storage::SparseMatrix<ValueType> const& transitionMatrix,
-    storm::storage::SparseMatrix<ValueType> const& backwardTransitions, storm::storage::BitVector const& targetStates,
-    std::function<storm::storage::BitVector()> const& zeroRewardStatesGetter, std::function<storm::storage::BitVector()> const& zeroRewardChoicesGetter) {
+    ModelCheckerEnvironment const& mcEnv, storm::solver::SolveGoal<ValueType, SolutionType> const& goal,
```
volkm: Why not use the normal Environment here as well?

src/storm/modelchecker/prctl/helper/SparseMdpPrctlHelper.cpp
```diff
@@ -1149,7 +1147,8 @@ QualitativeStateSetsReachabilityRewards computeQualitativeStateSetsReachabilityR
 }
 
 template<typename ValueType, typename SolutionType>
-QualitativeStateSetsReachabilityRewards getQualitativeStateSetsReachabilityRewards(storm::solver::SolveGoal<ValueType, SolutionType> const& goal,
+QualitativeStateSetsReachabilityRewards getQualitativeStateSetsReachabilityRewards(ModelCheckerEnvironment const& mcEnv,
```
volkm: Environment?

src/storm/storage/dd/bisimulation/BisimulationOptions.h
```diff
@@ -15,6 +17,8 @@ struct BisimulationOptions {
     InitialPartitionMode initialPartitionMode = InitialPartitionMode::Finer;
     bool useRepresentatives = false;
     bool useOriginalVariables = false;
+    bool verbose = false;
+    uint64_t showProgressDelay = 5;
```
volkm, line 21: The delay is also used in other places. Should we define this magic number somewhere centrally? Maybe directly in ProgressMeasurment? Or make it a CMake setting?

volkm (commented):
Thanks for the work.

I would favor moving the output settings (verbose, progress delay, exportX) outside of the existing environments and use a dedicated environment. These are also relevant outside the solver/modelchecker.
For some I could also see other approaches:
- fix progress delay centrally and not make it configurable
- verbose could be replaced by using the loggers with correct log level or STATISTICS_LAZY or introducing a new logger.

src/storm/storage/dd/bisimulation/BisimulationOptions.h
sjunges: I don't know; should progress delays be uniform over different parts of the code? Should we at compile time fix this default? 

Should there be a default. For me, these fields are not critical, so if a user/a programmer is not aware of these fields, I dont mind that much, so I rather keep the code a bit leaner and write the defaults where they are used, also without coupling lots of code places.

src/storm/environment/modelchecker/ModelCheckerEnvironment.h
sjunges: I dont know what would be governed by an output environment. I dont see what generating some kind of utlity  environment would resolve?

 I currently opted to do this the same way we handle pareto plots in multiobjective model checking. There could be arguments to make this part of a task, but these are also already heavily overloaded.

src/storm/environment/solver/SolverEnvironment.h
sjunges: I think there is an argument to be made that I want to have verbosity of solvers and model checkers independently, but I could also move it upwards to the general environments? These currently have no flags, but maybe that is better there. No strong opinion on this.

src/storm-dft/builder/ExplicitDFTModelBuilder.cpp
sjunges: yes, sorry! I think this depends a bit with whether we want to have the delay as part of an environment and how important we think it is that we use uniform values here everywhere. I could also imagine that you make this part of the DftModelBuilderOptions?

src/storm/environment/modelchecker/ModelCheckerEnvironment.h
volkm: After the discussion, I agree with keeping it here.

PR #1004, opened 2026-08-11
Revised POMDP belief exploration

AlexBork:
This PR contains the revised POMDP belief exploration implementation.
The revised implementation is more modular, making it easier to add extensions.
In addition, it is more performant than the old one.

This PR also adds the functionality to analyse reward-bounded reachability properties on POMDP, as described in https://proceedings.mlr.press/v286/bork25a.html 

The revised implementation is fully replacing the current one.
Wo do not port over all functionality from the old implementation.

Changes outside of Storm-POMDP are part of separate PR #1028

PR #1028, opened 2026-08-18
Changes to Storm for revised belief exploration

AlexBork:
This PR contains changes to Storm outside Storm-POMDP for the revised belief exploration implementation, see PR #1004.

- Transformer for transition-based to action-based rewards by inserting intermediate states
- Transformer for ValueTypes of sparse models
- Getter for optional time bound in BoundedUntilFormula


src/storm/logic/BoundedUntilFormula.cpp
```diff
@@ -251,6 +251,22 @@ storm::expressions::Expression const& BoundedUntilFormula::getUpperBound(unsigne
...
+        return lowerBound.at(i).get();
+    } else {
+        return std::nullopt;
+    }
+}
+
+std::optional<TimeBound> BoundedUntilFormula::getUpperBoundAsOptionalTimeBound(unsigned i) const {
+    if (hasUpperBound(i)) {
+        return upperBound.at(i).get();
+    } else {
```
sjunges: Why not just return upperBound.at(i) rather than the case split (we would need to move from boost to std then, but that would be good anyways.

src/storm/transformer/SparseModelValueTypeTransformer.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
```
sjunges: What happens here if InputType == OutputType? Why do we ever want those instantiations? 

What do we do if we go from doubles to rational numbers to ensure that probabilities still sum to one?

src/storm/transformer/TransitionToActionRewardTransformer.h
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "storm/models/sparse/Model.h"
```
sjunges, line 7: Some simple test cases for this would be good.

src/storm/logic/BoundedUntilFormula.cpp
AlexBork: PR #1030 addressed the change, so I will also fix this here :)

src/storm/transformer/SparseModelValueTypeTransformer.h
AlexBork: InputType == OutputType i somthing we should never need. I can add a static assertion that those are never generated.
I indeed did not consider the sum issue.
I cannot think of anything to address this properly. 
As background: I want to use this transformation to transform a rational to a double POMDP to apply the preprocessing (checking of underlying MDP there) without needing to re-parse the input model.

For now, this is also not essential, so I am considering just throwing it out. Any thoughts?

sjunges: - Input=Output: We are declaring them right now, right? That was what promted this in the first place. 
- I also understand the to-double use case, it is also something we sometimes do in the parametric setting. 
- We could renormalize afterwards for the sum to make sense.

Maybe rename to "toDouble" that supports going from exact rationals to doubles fixes parts of my concerns. :-)

src/storm/logic/BoundedUntilFormula.cpp
sjunges: Great, I appreciate it!

src/storm/transformer/SparseModelValueTypeTransformer.h
AlexBork: Wow, I somehow completely overlooked that I added the Input=Output declarations :D
Will remove those.

src/storm/transformer/SparseRationalModelToDoubleTransformer.cpp
```diff
@@ -0,0 +1,111 @@
...
+        if (rewardModel.hasTransitionRewards()) {
+            optionalTransitionRewardMatrix = rewardModel.getTransitionRewardMatrix().toValueType<double>();
+        }
+        convertedComponents.rewardModels.emplace(
+            rewardModelName, storm::models::sparse::StandardRewardModel<double>(
+                                 std::move(optionalStateRewardVector), std::move(optionalStateActionRewardVector), std::move(optionalTransitionRewardMatrix)));
+    }
+    switch (inputModel->getType()) {
+        case storm::models::ModelType::Dtmc:
+            return std::make_shared<storm::models::sparse::Dtmc<double>>(storm::models::sparse::Dtmc<double>(convertedComponents));
```
sjunges: this swtich case is extremely ugly. Is this the standard way to do this? @tquatmann ?

src/storm/transformer/TransitionToActionRewardTransformer.cpp
```diff
@@ -0,0 +1,232 @@
...
+    } else if (originalModel->isOfType(storm::models::ModelType::Ctmc)) {
+        components.rateTransitions = true;
+    } else {
+        STORM_LOG_THROW(originalModel->isOfType(storm::models::ModelType::Dtmc) || originalModel->isOfType(storm::models::ModelType::Mdp),
+                        storm::exceptions::UnexpectedException, "Unhandled model type.");
+    }
+    return {storm::utility::builder::buildModelFromComponents(originalModel->getType(), std::move(components)), std::move(originalToNewIndex)};
+}
+
+template struct TransitionToActionRewardTransformerReturnType<double>;
```
sjunges, line 234: Will this break on interval-based models? If not, lets add them as well.

src/storm/transformer/SparseRationalModelToDoubleTransformer.cpp
```diff
@@ -0,0 +1,111 @@
...
+namespace storm::transformer {
+std::shared_ptr<storm::models::sparse::Model<double>> sparseRationalModelToDouble(
+    std::shared_ptr<storm::models::sparse::Model<storm::RationalNumber>> const& inputModel, double precision) {
+    STORM_LOG_THROW(inputModel, storm::exceptions::IllegalArgumentTypeException, "Cannot transform a null model.");
+    storm::storage::sparse::ModelComponents<double> convertedComponents;
+    convertedComponents.transitionMatrix = inputModel->getTransitionMatrix().toValueType<double>();
+    if (inputModel->getType() != storm::models::ModelType::Ctmc) {
+        if (!convertedComponents.transitionMatrix.isProbabilistic(precision)) {
+            convertedComponents.transitionMatrix.divideRowsInPlace(convertedComponents.transitionMatrix.getRowSumVector());
+        }
```
tquatmann: imo, this should rather go through each row individually, compute the row sum and (if necessary) scale the matrix values.

Currently, this unnecessarily allocates memory for the entire rowSumVector and also touches matrix rows that are actually precise enough.

src/storm/transformer/SparseRationalModelToDoubleTransformer.cpp
```diff
@@ -0,0 +1,111 @@
...
+        std::optional<std::vector<double>> optionalStateRewardVector = std::nullopt;
+        std::optional<std::vector<double>> optionalStateActionRewardVector = std::nullopt;
+        std::optional<storm::storage::SparseMatrix<double>> optionalTransitionRewardMatrix = std::nullopt;
+        if (rewardModel.hasStateRewards()) {
+            std::vector<double> resultVector;
+            resultVector.reserve(rewardModel.getStateRewardVector().size());
+            for (auto const& oldValue : rewardModel.getStateRewardVector()) {
+                resultVector.push_back(storm::utility::convertNumber<double>(oldValue));
+            }
+            optionalStateRewardVector = resultVector;
```
tquatmann: Can be simplified to
```suggestion
            optionalStateRewardVector = storm::utility::vector::convertNumericVector<double>(rewardModel.getStateRewardVector());
```

(same below)

src/storm/transformer/SparseRationalModelToDoubleTransformer.cpp
tquatmann: There is `buildModelFromComponents` in `storm/utility/builder.h`. 

We still need some switch case like this for the model-specific components (exit rates for MAs, player data for SMG etc

src/storm/transformer/SparseRationalModelToDoubleTransformer.cpp
```diff
@@ -0,0 +1,111 @@
...
+        case storm::models::ModelType::Mdp:
+            return std::make_shared<storm::models::sparse::Mdp<double>>(storm::models::sparse::Mdp<double>(convertedComponents));
+        case storm::models::ModelType::Ctmc: {
+            auto ctmc = inputModel->as<storm::models::sparse::Ctmc<storm::RationalNumber>>();
+            std::vector<double> resultVector;
+            resultVector.reserve(ctmc->getExitRateVector().size());
+            for (auto const& oldValue : ctmc->getExitRateVector()) {
+                resultVector.push_back(storm::utility::convertNumber<double>(oldValue));
+            }
+            convertedComponents.exitRates = resultVector;
```
tquatmann:
```suggestion
```

We provide the rates in the transition matrix. A dedicated exit rate vector should not be set and will be ignored in this case.
We do need to set `rateTransitions = true` to indicate that the transitionMatrix contains rates, not probabilities.

src/storm/transformer/SparseRationalModelToDoubleTransformer.cpp
```diff
@@ -0,0 +1,111 @@
...
+            return std::make_shared<storm::models::sparse::Smg<double>>(models::sparse::Smg<double>(convertedComponents));
+        }
+        case storm::models::ModelType::S2pg: {
+            auto s2pg = inputModel->as<storm::models::sparse::StochasticTwoPlayerGame<storm::RationalNumber>>();
+            convertedComponents.player1Matrix = s2pg->getPlayer1Matrix();
+            return std::make_shared<storm::models::sparse::StochasticTwoPlayerGame<double>>(
+                models::sparse::StochasticTwoPlayerGame<double>(convertedComponents));
+        }
+        default:
+            STORM_LOG_THROW(false, storm::exceptions::IllegalArgumentTypeException,
```
tquatmann:
```suggestion
            STORM_LOG_THROW(false, storm::exceptions::NotSupportedException,
```

tquatmann: IllegalArgumentTypeException is more of a `storm::settings` thing.

src/storm/transformer/TransitionToActionRewardTransformer.cpp
```diff
@@ -0,0 +1,257 @@
...
+        for (uint64_t oldState = 0; oldState < originalModel->getNumberOfStates(); ++oldState) {
+            uint64_t const oldStartRow = transitions.getRowGroupIndices()[oldState];
+            uint64_t const newState = originalToNewIndex[oldState];
+            uint64_t const newStartRow = useGroups ? components.transitionMatrix.getRowGroupIndices()[newState] : newState;
+            uint64_t const numRowsInGroup = useGroups ? transitions.getRowGroupSize(oldState) : 1ull;
+            for (uint64_t groupOffset = 0; groupOffset < numRowsInGroup; ++groupOffset) {
+                auto& rewValue = newActionRewardVector[newStartRow + groupOffset];
+                if (oldRewardModel.hasStateRewards()) {
+                    rewValue += oldRewardModel.getStateReward(oldState);
+                }
```
tquatmann: For continuous time models, state rewards are earned over time. Collapsing them together with the action rewards like this is usually incorrect. However, for discrete-time models, it's probably convenient to merge the rewards.

I suggest this change here:

```suggestion
                if (originalModel->isDiscreteTimeModel() && oldRewardModel.hasStateRewards()) {
                    rewValue += oldRewardModel.getStateReward(oldState);
                }
```

and below something like

```cpp
std::optional<std::vector<ValueType>> newStateRewards;
if (originalModel->isContinuousTimeModel() && oldRewardModel.hasStateRewards()) {
        newStateRewards.emplace(numStates, storm::utility::zero<ValueType>())
        for (uint64_t origState = 0; origState < originalModel->getNumberOfStates(); ++origState) {
            uint64_t const newState = originalToNewIndex[origState];
            newStateRewards.value()[newState] = oldRewardModel.getStateReward(oldState);
        }
}
RewardModelType newRewardModel(newStateRewards, std::move(newActionRewardVector));
```

src/storm/transformer/TransitionToActionRewardTransformer.cpp
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    // Make a pass to find the different unique rewards with which a state is entered.
+    // We do not use std::set here as the interval ordering is not a strict weak ordering for overlapping intervals.
+    // Equality-based deduplication preserves distinct overlapping rewards.
+    std::vector<std::vector<detail::MultiRewardVector<RewardValueType>>> incomingRewards(originalModel->getNumberOfStates());
+    auto const& transitions = originalModel->getTransitionMatrix();
+    for (uint64_t row = 0; row < transitions.getRowCount(); ++row) {
+        rewardTransitionIterator.forEachRowEntry(
+            row, true, [&incomingRewards](uint64_t column, RewardValueType, detail::MultiRewardVector<RewardValueType> const& rewards) {
```
tquatmann:
```suggestion
            row, true, [&incomingRewards](uint64_t column, ValueType, detail::MultiRewardVector<RewardValueType> const& rewards) {
```

tquatmann (commented):
Thanks! After looking into my comments, this PR should be good to go imo.

src/storm/transformer/TransitionToActionRewardTransformer.cpp
```diff
@@ -185,15 +185,23 @@ TransitionToActionRewardTransformerReturnType<ValueType, RewardModelType> transf
...
-        RewardModelType newRewardModel(std::nullopt, std::move(newActionRewardVector));
+        std::optional<std::vector<typename RewardModelType::ValueType>> newStateRewards;
+        if (!originalModel->isDiscreteTimeModel() && oldRewardModel.hasStateRewards()) {
+            newStateRewards.emplace(numStates, storm::utility::zero<ValueType>());
+            for (uint64_t origState = 0; origState < originalModel->getNumberOfStates(); ++origState) {
+                uint64_t const newState = originalToNewIndex[origState];
+                newStateRewards.value()[newState] = oldRewardModel.getStateReward(origState);
+            }
+        }
+        RewardModelType newRewardModel(newStateRewards, std::move(newActionRewardVector));
```
tquatmann: `std::move(newStateRewards)`

tquatmann (approved):
LGTM! Just a very small nitpick.

PR #1030, opened 2026-08-20
Change boost::optional to std::optional in BoundedUntilFormula

AlexBork:


src/storm-parsers/parser/FormulaParserGrammar.cpp
volkm, line 473: Will changing this `boost::optional` be a separate step?

src/storm-parsers/parser/FormulaParserGrammar.cpp
```diff
@@ -473,11 +474,21 @@ std::shared_ptr<storm::logic::Formula const> FormulaParserGrammar::createEventua
                                            std::shared_ptr<storm::logic::TimeBoundReference>>>> const& timeBounds,
     storm::logic::FormulaContext context, std::shared_ptr<storm::logic::Formula const> const& subformula) const {
     if (timeBounds && !timeBounds.get().empty()) {
-        std::vector<boost::optional<storm::logic::TimeBound>> lowerBounds, upperBounds;
+        std::vector<std::optional<storm::logic::TimeBound>> lowerBounds, upperBounds;
         std::vector<storm::logic::TimeBoundReference> timeBoundReferences;
         for (auto const& timeBound : timeBounds.get()) {
-            lowerBounds.push_back(std::get<0>(timeBound));
-            upperBounds.push_back(std::get<1>(timeBound));
+            auto const& lowerBound = std::get<0>(timeBound);
```
volkm, line 482: If I understand correctly, this more complicated part is now only needed because we need to convert from `boost::optional` to `std::optional`?
Can you add a comment as explanation? Then we are aware to change it back if the input becomes `std::optional`.

src/storm-parsers/parser/FormulaParserGrammar.cpp
AlexBork: Yes, I think this requires further changes in the parser which I tried to avoid for now.

src/storm-parsers/parser/FormulaParserGrammar.cpp
AlexBork: I added comments to the two relevant places

volkm (approved):
LGTM. Thanks for the changes.

PR #1031, opened 2026-08-20
Add upper and lowerbounds to check results.

lukovdm:
This PR adds upper and lower bounds to `ExplicitQuantitativeCheckResults` and implements this for OVI and II. Further algorithms are implemented it #1048. 

Upper and lower bounds are stored in a class `SolutionBounds`. It has an optional vector of lower bounds and an optional vector of upper bounds. No bound mean the algorithm did not try to give any bounds, infinity as a value in the bound means the algorithm tried and that is the best it came up with.

The `ExplicitQuantitativeCheckResults` and `ExplicitQualitativeCheckResults` have had there map option removed and instead get an optional bitvector `states` holding the states for which the values and bounds hold. This was done such that the vectors in the SolutionBounds can be used easily in conjuction with any values stored in the check results. `ExplicitQualitativeCheckResults` was update to be consistent with `ExplicitQuantitativeCheckResults`.

Solution bounds are passed around from helper to PRCTLmodel checker using `(MDP/DTMC)SparseModelCheckingHelperReturnType`, the dtmc variant is new.

Solution bounds in this PR have been implemented for OVI and II.

Printing of bounds has been implemented where no bound is reported with a `-` if no bound is present. This was done to differentiate it from infinities.

So far I have the convention that any non-maybe states get bounds that exist from the algorithm. So if only lower bounds are produced, the maybe states get 1 [1, -] where - is no bound. This could be changed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,4 @@ nbproject/
 .idea
 .vscode
 *.out
+.claude/
--- a/src/storm/modelchecker/csl/SparseCtmcCslModelChecker.cpp
+++ b/src/storm/modelchecker/csl/SparseCtmcCslModelChecker.cpp
@@ -94,9 +94,11 @@ std::unique_ptr<CheckResult> SparseCtmcCslModelChecker<SparseCtmcModelType>::com
     std::unique_ptr<CheckResult> subResultPointer = this->check(env, pathFormula.getSubformula());
     ExplicitQualitativeCheckResult<ValueType> const& subResult = subResultPointer->template asExplicitQualitativeCheckResult<ValueType>();
     auto probabilisticTransitions = this->getModel().computeProbabilityMatrix();
-    std::vector<ValueType> numericResult = storm::modelchecker::helper::SparseDtmcPrctlHelper<ValueType>::computeGloballyProbabilities(
-        env, storm::solver::SolveGoal<ValueType>(this->getModel(), checkTask), probabilisticTransitions, probabilisticTransitions.transpose(),
-        subResult.getTruthValuesVector(), checkTask.isQualitativeSet());
+    // CTMC results do not carry bounds yet.
+    std::vector<ValueType> numericResult = std::move(storm::modelchecker::helper::SparseDtmcPrctlHelper<ValueType>::computeGloballyProbabilities(
+                                                         env, storm::solver::SolveGoal<ValueType>(this->getModel(), checkTask), probabilisticTransitions,
+                                                         probabilisticTransitions.transpose(), subResult.getTruthValuesVector(), checkTask.isQualitativeSet())
+                                                         .values);
     return std::unique_ptr<CheckResult>(new ExplicitQuantitativeCheckResult<ValueType>(std::move(numericResult)));
 }
 
--- a/src/storm/modelchecker/csl/SparseCtmcCslModelChecker.cpp
+++ b/src/storm/modelchecker/csl/SparseCtmcCslModelChecker.cpp
@@ -94,7 +94,6 @@ std::unique_ptr<CheckResult> SparseCtmcCslModelChecker<SparseCtmcModelType>::com
     std::unique_ptr<CheckResult> subResultPointer = this->check(env, pathFormula.getSubformula());
     ExplicitQualitativeCheckResult<ValueType> const& subResult = subResultPointer->template asExplicitQualitativeCheckResult<ValueType>();
     auto probabilisticTransitions = this->getModel().computeProbabilityMatrix();
-    // CTMC results do not carry bounds yet.
     std::vector<ValueType> numericResult = std::move(storm::modelchecker::helper::SparseDtmcPrctlHelper<ValueType>::computeGloballyProbabilities(
                                                          env, storm::solver::SolveGoal<ValueType>(this->getModel(), checkTask), probabilisticTransitions,
                                                          probabilisticTransitions.transpose(), subResult.getTruthValuesVector(), checkTask.isQualitativeSet())
--- a/src/storm/modelchecker/prctl/SparseDtmcPrctlModelChecker.cpp
+++ b/src/storm/modelchecker/prctl/SparseDtmcPrctlModelChecker.cpp
@@ -151,12 +151,17 @@ std::unique_ptr<CheckResult> SparseDtmcPrctlModelChecker<SparseDtmcModelType>::c
     std::unique_ptr<CheckResult> rightResultPointer = this->check(env, pathFormula.getRightSubformula());
     ExplicitQualitativeCheckResult<SolutionType> const& leftResult = leftResultPointer->template asExplicitQualitativeCheckResult<SolutionType>();
     ExplicitQualitativeCheckResult<SolutionType> const& rightResult = rightResultPointer->template asExplicitQualitativeCheckResult<SolutionType>();
+    storm::solver::SolutionBounds<SolutionType> solutionBounds;
     std::vector<SolutionType> numericResult =
         storm::modelchecker::helper::SparseDtmcPrctlHelper<ValueType, RewardModelType, SolutionType>::computeUntilProbabilities(
             env, storm::solver::SolveGoal<ValueType, SolutionType>(this->getModel(), checkTask), this->getModel().getTransitionMatrix(),
             this->getModel().getBackwardTransitions(), leftResult.getTruthValuesVector(), rightResult.getTruthValuesVector(), checkTask.isQualitativeSet(),
-            checkTask.getHint());
-    return std::unique_ptr<CheckResult>(new ExplicitQuantitativeCheckResult<SolutionType>(std::move(numericResult)));
+            checkTask.getHint(), &solutionBounds);
+    auto result = std::make_unique<ExplicitQuantitativeCheckResult<SolutionType>>(std::move(numericResult));
+    if (solutionBounds) {
+        result->setBounds(std::move(solutionBounds->first), std::move(solutionBounds->second));
+    }
+    return result;
 }
 
 template<typename SparseDtmcModelType>
@@ -168,11 +173,16 @@ std::unique_ptr<CheckResult> SparseDtmcPrctlModelChecker<SparseDtmcModelType>::c
         storm::logic::GloballyFormula const& pathFormula = checkTask.getFormula();
         std::unique_ptr<CheckResult> subResultPointer = this->check(env, pathFormula.getSubformula());
         ExplicitQualitativeCheckResult<SolutionType> const& subResult = subResultPointer->template asExplicitQualitativeCheckResult<SolutionType>();
+        storm::solver::SolutionBounds<SolutionType> solutionBounds;
         std::vector<SolutionType> numericResult =
             storm::modelchecker::helper::SparseDtmcPrctlHelper<ValueType, RewardModelType, SolutionType>::computeGloballyProbabilities(
                 env, storm::solver::SolveGoal<ValueType, SolutionType>(this->getModel(), checkTask), this->getModel().getTransitionMatrix(),
-                this->getModel().getBackwardTransitions(), subResult.getTruthValuesVector(), checkTask.isQualitativeSet());
-        return std::unique_ptr<CheckResult>(new ExplicitQuantitativeCheckResult<SolutionType>(std::move(numericResult)));
+                this->getModel().getBackwardTransitions(), subResult.getTruthValuesVector(), checkTask.isQualitativeSet(), &solutionBounds);
+        auto result = std::make_unique<ExplicitQuantitativeCheckResult<SolutionType>>(std::move(numericResult));
+        if (solutionBounds) {
+            result->setBounds(std::move(solutionBounds->first), std::move(solutionBounds->second));
+        }
+        return result;
     }
 }
 
--- a/src/storm/modelchecker/prctl/SparseDtmcPrctlModelChecker.cpp
+++ b/src/storm/modelchecker/prctl/SparseDtmcPrctlModelChecker.cpp
@@ -98,7 +98,7 @@ std::unique_ptr<CheckResult> SparseDtmcPrctlModelChecker<SparseDtmcModelType>::c
         auto formula = std::make_shared<storm::logic::ProbabilityOperatorFormula>(checkTask.getFormula().asSharedPointer(), opInfo);
         auto numericResult = storm::modelchecker::helper::SparseDtmcPrctlHelper<ValueType, RewardModelType, SolutionType>::computeRewardBoundedValues(
             env, this->getModel(), formula);
-        return std::unique_ptr<CheckResult>(new ExplicitQuantitativeCheckResult<SolutionType>(std::move(numericResult)));
+        return std::unique_ptr<CheckResult>(new ExplicitQuantitativeCheckResult<SolutionType>(this->getModel().getInitialStates(), std::move(numericResult)));
     } else {
         STORM_LOG_THROW(pathFormula.hasUpperBound(), storm::exceptions::InvalidPropertyException, "Formula needs to have (a single) upper step bound.");
         STORM_LOG_THROW(pathFormula.hasIntegerLowerBound(), storm::exceptions::InvalidPropertyException, "Formula lower step bound must be discrete/integral.");
@@ -156,7 +156,7 @@ std::unique_ptr<CheckResult> SparseDtmcPrctlModelChecker<SparseDtmcModelType>::c
         this->getModel().getBackwardTransitions(), leftResult.getTruthValuesVector(), rightResult.getTruthValuesVector(), checkTask.isQualitativeSet(),
         checkTask.getHint());
     auto result = std::make_unique<ExplicitQuantitativeCheckResult<SolutionType>>(std::move(ret.values));
-    setBounds(*result, ret.solutionBounds);
+    result->setBounds(std::move(ret.solutionBounds));
     return result;
 }
 
@@ -173,7 +173,7 @@ std::unique_ptr<CheckResult> SparseDtmcPrctlModelChecker<SparseDtmcModelType>::c
             env, storm::solver::SolveGoal<ValueType, SolutionType>(this->getModel(), checkTask), this->getModel().getTransitionMatrix(),
             this->getModel().getBackwardTransitions(), subResult.getTruthValuesVector(), checkTask.isQualitativeSet());
         auto result = std::make_unique<ExplicitQuantitativeCheckResult<SolutionType>>(std::move(ret.values));
-        setBounds(*result, ret.solutionBounds);
+        result->setBounds(std::move(ret.solutionBounds));
         return result;
     }
 }
@@ -238,7 +238,7 @@ std::unique_ptr<CheckResult> SparseDtmcPrctlModelChecker<SparseDtmcModelType>::c
             auto formula = std::make_shared<storm::logic::RewardOperatorFormula>(checkTask.getFormula().asSharedPointer(), checkTask.getRewardModel(), opInfo);
             auto numericResult = storm::modelchecker::helper::SparseDtmcPrctlHelper<ValueType, RewardModelType, SolutionType>::computeRewardBoundedValues(
                 env, this->getModel(), formula);
-            return std::unique_ptr<CheckResult>(new ExplicitQuantitativeCheckResult<ValueType>(std::move(numericResult)));
+            return std::unique_ptr<CheckResult>(new ExplicitQuantitativeCheckResult<ValueType>(this->getModel().getInitialStates(), std::move(numericResult)));
         } else {
             STORM_LOG_THROW(rewardPathFormula.hasIntegerBound(), storm::exceptions::InvalidPropertyException, "Formula needs to have a discrete time bound.");
             auto rewardModel = storm::utility::createFilteredRewardModel(this->getModel(), checkTask);
--- a/src/storm/modelchecker/prctl/SparseDtmcPrctlModelChecker.cpp
+++ b/src/storm/modelchecker/prctl/SparseDtmcPrctlModelChecker.cpp
@@ -151,16 +151,12 @@ std::unique_ptr<CheckResult> SparseDtmcPrctlModelChecker<SparseDtmcModelType>::c
     std::unique_ptr<CheckResult> rightResultPointer = this->check(env, pathFormula.getRightSubformula());
     ExplicitQualitativeCheckResult<SolutionType> const& leftResult = leftResultPointer->template asExplicitQualitativeCheckResult<SolutionType>();
     ExplicitQualitativeCheckResult<SolutionType> const& rightResult = rightResultPointer->template asExplicitQualitativeCheckResult<SolutionType>();
-    storm::solver::SolutionBounds<SolutionType> solutionBounds;
-    std::vector<SolutionType> numericResult =
-        storm::modelchecker::helper::SparseDtmcPrctlHelper<ValueType, RewardModelType, SolutionType>::computeUntilProbabilities(
-            env, storm::solver::SolveGoal<ValueType, SolutionType>(this->getModel(), checkTask), this->getModel().getTransitionMatrix(),
-            this->getModel().getBackwardTransitions(), leftResult.getTruthValuesVector(), rightResult.getTruthValuesVector(), checkTask.isQualitativeSet(),
-            checkTask.getHint(), &solutionBounds);
-    auto result = std::make_unique<ExplicitQuantitativeCheckResult<SolutionType>>(std::move(numericResult));
-    if (solutionBounds) {
-        result->setBounds(std::move(solutionBounds->first), std::move(solutionBounds->second));
-    }
+    auto ret = storm::modelchecker::helper::SparseDtmcPrctlHelper<ValueType, RewardModelType, SolutionType>::computeUntilProbabilities(
+        env, storm::solver::SolveGoal<ValueType, SolutionType>(this->getModel(), checkTask), this->getModel().getTransitionMatrix(),
+        this->getModel().getBackwardTransitions(), leftResult.getTruthValuesVector(), rightResult.getTruthValuesVector(), checkTask.isQualitativeSet(),
+        checkTask.getHint());
+    auto result = std::make_unique<ExplicitQuantitativeCheckResult<SolutionType>>(std::move(ret.values));
+    setBounds(*result, ret.solutionBounds);
     return result;
 }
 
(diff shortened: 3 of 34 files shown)


src/storm/modelchecker/prctl/helper/SparseDtmcPrctlHelper.cpp
```diff
@@ -172,7 +172,7 @@ template<typename ValueType, typename RewardModelType, typename SolutionType>
 std::vector<SolutionType> SparseDtmcPrctlHelper<ValueType, RewardModelType, SolutionType>::computeUntilProbabilities(
     Environment const& env, storm::solver::SolveGoal<ValueType, SolutionType>&& goal, storm::storage::SparseMatrix<ValueType> const& transitionMatrix,
     storm::storage::SparseMatrix<ValueType> const& backwardTransitions, storm::storage::BitVector const& phiStates, storm::storage::BitVector const& psiStates,
-    bool qualitative, ModelCheckerHint const& hint) {
+    bool qualitative, ModelCheckerHint const& hint, storm::solver::SolutionBounds<SolutionType>* solutionBounds) {
```
volkm: We generally try to avoid pointers. Maybe use a reference?

src/storm/modelchecker/results/ExplicitQuantitativeCheckResult.h
```diff
@@ -85,9 +115,26 @@ class ExplicitQuantitativeCheckResult : public QuantitativeCheckResult<ValueType
...
+    void assertBoundsShape(boost::variant<vector_type, map_type> const& bounds) const;
+
+    /*!
+     * Writes the value of the given state, followed by its bounds if any are known.
+     */
+    void printValue(std::ostream& out, storm::storage::sparse::state_type state) const;
+
+    // The values of the quantitative check result. These are estimates of the actual values, which lie within
+    // the bounds below but carry no further guarantee.
     boost::variant<vector_type, map_type> values;
```
volkm: I would have thought that the values itself would now incorporate the type `SolutionBounds` to already include the lower and upper bounds.
Why are separate lower/upper bounds the better approach?

src/storm/solver/SolutionBounds.h
```diff
@@ -0,0 +1,17 @@
...
+
+namespace storm::solver {
+
+/*!
+ * Sound lower and upper bounds on a solution vector, as computed by e.g. interval iteration.
+ * The first component bounds the solution from below, the second one from above. Both have the same size as
+ * the solution. An unset value means that the algorithm in question did not provide any bounds.
+ */
+template<typename ValueType>
+using SolutionBounds = std::optional<std::pair<std::vector<ValueType>, std::vector<ValueType>>>;
```
volkm: Maybe use a struct to be more flexible for future changes?

src/storm/modelchecker/results/ExplicitQuantitativeCheckResult.h
lukovdm: Several of the methods producing bounds currently do so in the form of a vector of lower bounds and a vector of upper bounds. This is also the format they are usually consumed in. I first tried incorporating the bounds in the value, but that led to a lot of rebuilding of vectors from the boundvalues. I thought this was ineligent and it looked slow.

I would like a better way to bundle soundess results of modelchecking that is more universal, but every CheckResult has a slightly different (justified) way of handling its values and thus I did not see a good way to do this.

src/storm/modelchecker/results/ExplicitQuantitativeCheckResult.h
```diff
@@ -85,11 +116,42 @@ class ExplicitQuantitativeCheckResult : public QuantitativeCheckResult<ValueType
...
 
+    // Sound lower bounds on the actual values, if an algorithm provided them.
+    std::optional<boost::variant<vector_type, map_type>> lowerBounds;
+
+    // Sound upper bounds on the actual values, if an algorithm provided them.
+    std::optional<boost::variant<vector_type, map_type>> upperBounds;
```
sjunges: I dont know whehter this should be part of this PR, but this could be a good moment to replace boost::variant here with std::variant?

src/storm/modelchecker/results/ExplicitQuantitativeCheckResult.h
volkm: I see that the bare-bone methods return vectors. I would have thought that at the CheckResult stage, the result is somewhat stable and will not change anymore. So this could be good point to go to the `SolutionBounds`?

I think the general discussion is at what level we want to introduce the SolutionBounds. For me, I would for example already find them useful in some of the helper methods such as Fox-Glynn.

sjunges: Are you suggesting to have vectors of pairs instead of a pair of vectors? I am not sure I would agree with that, from an algorithm perspective.

volkm: I would keep the pair of vectors. I agree that this is more performant.
In the current version `using SolutionBounds = std::optional<std::pair<std::vector<ValueType>, std::vector<ValueType>>>;` so it would should already fit. I would just use these new type to make sure we have a consistent handling of the bounds.

lukovdm: The `ExplicitQualitativeCheckResult`  also supports maps of values instead of vectors of values. Which is used in filtered results and methods which only produce a value for the initial state. But I did not add this as a variant to the `SolutionBounds`, as that would require supporting maps everywhere.

src/storm/modelchecker/csl/SparseCtmcCslModelChecker.cpp
```diff
@@ -94,9 +94,10 @@ std::unique_ptr<CheckResult> SparseCtmcCslModelChecker<SparseCtmcModelType>::com
...
     auto probabilisticTransitions = this->getModel().computeProbabilityMatrix();
-    std::vector<ValueType> numericResult = storm::modelchecker::helper::SparseDtmcPrctlHelper<ValueType>::computeGloballyProbabilities(
-        env, storm::solver::SolveGoal<ValueType>(this->getModel(), checkTask), probabilisticTransitions, probabilisticTransitions.transpose(),
-        subResult.getTruthValuesVector(), checkTask.isQualitativeSet());
+    std::vector<ValueType> numericResult = std::move(storm::modelchecker::helper::SparseDtmcPrctlHelper<ValueType>::computeGloballyProbabilities(
+                                                         env, storm::solver::SolveGoal<ValueType>(this->getModel(), checkTask), probabilisticTransitions,
+                                                         probabilisticTransitions.transpose(), subResult.getTruthValuesVector(), checkTask.isQualitativeSet())
+                                                         .values);
     return std::unique_ptr<CheckResult>(new ExplicitQuantitativeCheckResult<ValueType>(std::move(numericResult)));
```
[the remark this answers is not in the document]
lukovdm: Additional algoritms are implemented in #1048.

src/storm/modelchecker/results/ExplicitQuantitativeCheckResult.h
lukovdm: This has been solved by using vectors everywhere and giving an optional bitvector selecting states.

src/storm/modelchecker/results/ExplicitQualitativeCheckResult.h
```diff
@@ -52,8 +68,22 @@ class ExplicitQualitativeCheckResult : public QualitativeCheckResult {
...
     virtual void complement() override;
 
+    /*!
+     * Retrieves whether the result holds a truth value for the given state.
+     */
+    bool hasValueForState(storm::storage::sparse::state_type state) const;
+
+    /*!
+     * Retrieves the states this result holds truth values for.
+     * @pre This is not a result for all states.
```
volkm: Could it not work for all states as well and then just return the BitVector where everything is true?

src/storm/modelchecker/results/ExplicitQualitativeCheckResult.h
```diff
@@ -52,8 +68,22 @@ class ExplicitQualitativeCheckResult : public QualitativeCheckResult {
...
+     */
+    storm::storage::BitVector const& getStates() const;
+
+    /*!
+     * Retrieves the truth values, one per state this result is for. If this is not a result for all states, the
+     * i-th truth value belongs to the i-th state selected by getStates().
```
volkm:
```suggestion
     * Retrieves the truth values, one per state this result is for.
     * The i-th truth value belongs to the i-th state selected by getStates().
```
I think it should also work for all states.

src/storm/modelchecker/results/ExplicitQualitativeCheckResult.h
```diff
@@ -79,8 +109,36 @@ class ExplicitQualitativeCheckResult : public QualitativeCheckResult {
...
+                ++offset;
+            }
+        } else {
+            for (uint64_t state = 0; state < truthValues.size(); ++state) {
+                f(state, state);
+            }
+        }
+    }
+
+    // The states this result holds truth values for, or nothing at all if it is a result for all states.
```
volkm:
```suggestion
    // The states this result holds truth values for, or std::nullopt if it is a result for all states.
```
To make the difference to an empty bitvector clear.

src/storm/modelchecker/results/ExplicitQualitativeCheckResult.h
```diff
@@ -79,8 +109,36 @@ class ExplicitQualitativeCheckResult : public QualitativeCheckResult {
...
+            for (auto const& state : *states) {
+                f(state, offset);
+                ++offset;
+            }
+        } else {
+            for (uint64_t state = 0; state < truthValues.size(); ++state) {
+                f(state, state);
+            }
+        }
+    }
```
volkm, line 127: I imagine function `f` will always access the truth value at `truthValues[offset]` or `state` respective? Maybe already include it here and just apply f on state and corresponding truth value?

Same holds for the ExplicitQuantitativeCheckResult then.

src/storm/modelchecker/results/ExplicitQualitativeCheckResult.cpp
```diff
@@ -12,87 +14,49 @@ namespace storm {
...
-template<typename ValueType>
-ExplicitQualitativeCheckResult<ValueType>::ExplicitQualitativeCheckResult(storm::storage::sparse::state_type state, bool value) : truthValues(map_type()) {
-    boost::get<map_type>(truthValues)[state] = value;
-}
-
-template<typename ValueType>
-ExplicitQualitativeCheckResult<ValueType>::ExplicitQualitativeCheckResult(storm::storage::BitVector const& truthValues) : truthValues(truthValues) {
-    // Intentionally left empty.
+ExplicitQualitativeCheckResult<ValueType>::ExplicitQualitativeCheckResult(storm::storage::sparse::state_type state, bool value)
+    : states(storm::storage::BitVector(state + 1)), truthValues(1, value) {
```
volkm, line 18: Why is it index 1 and not 0?

src/storm/modelchecker/results/ExplicitQualitativeCheckResult.cpp
```diff
@@ -110,78 +74,53 @@ QualitativeCheckResult& ExplicitQualitativeCheckResult<ValueType>::operator|=(Qu
...
-        return !boost::get<vector_type>(truthValues).empty();
-    } else {
-        for (auto& element : boost::get<map_type>(truthValues)) {
-            if (element.second) {
-                return true;
-            }
-        }
-        return false;
-    }
+    return !truthValues.empty();
```
volkm, line 77: Great to see that the new approach simplifies these functions quite a bit :)

src/storm/modelchecker/results/ExplicitQualitativeCheckResult.cpp
```diff
@@ -110,78 +74,53 @@ QualitativeCheckResult& ExplicitQualitativeCheckResult<ValueType>::operator|=(Qu
...
+bool ExplicitQualitativeCheckResult<ValueType>::hasValueForState(storm::storage::sparse::state_type state) const {
+    if (states) {
+        return state < states->size() && states->get(state);
     }
+    return state < truthValues.size();
+}
+
+template<typename ValueType>
+uint64_t ExplicitQualitativeCheckResult<ValueType>::getOffset(storm::storage::sparse::state_type state) const {
+    STORM_LOG_THROW(this->hasValueForState(state), storm::exceptions::InvalidOperationException, "Unknown key '" << state << "'.");
```
volkm:
```suggestion
    STORM_LOG_ASSERT(this->hasValueForState(state), "State '" << state << "' has no value.");
```
Function is private, so I would use assert here for performance reasons.

src/storm/modelchecker/results/ExplicitQualitativeCheckResult.cpp
```diff
@@ -110,78 +74,53 @@ QualitativeCheckResult& ExplicitQualitativeCheckResult<ValueType>::operator|=(Qu
...
-    }
+    return truthValues.get(this->getOffset(state));
 }
 
 template<typename ValueType>
-typename ExplicitQualitativeCheckResult<ValueType>::vector_type const& ExplicitQualitativeCheckResult<ValueType>::getTruthValuesVector() const {
-    return boost::get<vector_type>(truthValues);
+storm::storage::BitVector const& ExplicitQualitativeCheckResult<ValueType>::getStates() const {
+    STORM_LOG_THROW(!this->isResultForAllStates(), storm::exceptions::InvalidOperationException,
+                    "Unable to retrieve the states of a result that is for all states.");
```
volkm: Why not return a full BitVector?

src/storm/modelchecker/results/ExplicitQualitativeCheckResult.cpp
```diff
@@ -201,44 +140,16 @@ bool ExplicitQualitativeCheckResult<ValueType>::isExplicitQualitativeCheckResult
...
-            out << "{true}";
-        } else if (allFalse) {
-            out << "{false}";
-        } else {
-            out << "{true, false}";
-        }
+    if (!this->isResultForAllStates() && truthValues.size() == 1) {
+        std::ios::fmtflags oldflags(out.flags());
+        out << std::boolalpha << truthValues.get(0);
+        out.flags(oldflags);
```
volkm:
```suggestion
        out << (truthValues.get(0) ? "true" : "false");
```

src/storm/modelchecker/results/ExplicitQuantitativeCheckResult.cpp
```diff
@@ -2,135 +2,207 @@
...
-ExplicitQuantitativeCheckResult<ValueType>::ExplicitQuantitativeCheckResult(std::vector<ValueType>&& values)
-    requires(!std::is_same_v<storm::utility::ExtendedValueType<ValueType>, ValueType>)
-    : values(storm::utility::fromSentinel(std::move(values))) {
-    // Intentionally left empty.
+ExplicitQuantitativeCheckResult<ValueType>::ExplicitQuantitativeCheckResult(ExplicitQualitativeCheckResult<ValueType> const& other) {
+    auto const toValue = [](bool truthValue) { return truthValue ? storm::utility::one<ExtendedValueType>() : storm::utility::zero<ExtendedValueType>(); };
+
+    storm::storage::BitVector const& truthValues = other.getTruthValuesVector();
+    values.reserve(truthValues.size());
+    for (std::size_t i = 0, n = truthValues.size(); i < n; i++) {
```
volkm:
```suggestion
    for (std::size_t i = 0; i < truthValues.size(); i++) {
```

src/storm/modelchecker/results/ExplicitQuantitativeCheckResult.cpp
```diff
@@ -2,135 +2,207 @@
...
-    return storm::utility::narrowFinite<ValueType>(boost::get<vector_type>(values));
+storm::solver::SolutionBounds<typename ExplicitQuantitativeCheckResult<ValueType>::ExtendedValueType> const&
+ExplicitQuantitativeCheckResult<ValueType>::getSolutionBounds() const {
+    return bounds;
+}
+
+template<typename ValueType>
+void ExplicitQuantitativeCheckResult<ValueType>::assertBoundsShape([[maybe_unused]] vector_type const& bounds) const {
+    STORM_LOG_ASSERT(bounds.size() == values.size(), "Bounds must have the same size as the values.");
+}
```
volkm: Currently, this is only used twice and could be inlined instead.

src/storm/modelchecker/results/ExplicitQuantitativeCheckResult.cpp
```diff
@@ -2,135 +2,207 @@
...
+
+template<typename ValueType>
+void ExplicitQuantitativeCheckResult<ValueType>::setBounds(storm::solver::SolutionBounds<ExtendedValueType> bounds) {
+    if (bounds.hasLower()) {
+        this->setLowerBounds(std::move(*bounds.lower));
+    }
+    if (bounds.hasUpper()) {
+        this->setUpperBounds(std::move(*bounds.upper));
+    }
```
volkm: Why not set `this->bounds` directly?

src/storm/modelchecker/results/ExplicitQuantitativeCheckResult.cpp
```diff
@@ -320,54 +340,62 @@ void printRange(std::ostream& out, ValueType const& min, ValueType const& max) {
...
-                }
-                print(out, element);
+    if (values.size() >= 10 && minMaxSupported) {
+        std::pair<ExtendedValueType, ExtendedValueType> minmax = this->getMinMax();
+        printRange(out, minmax.first, minmax.second);
+        if (this->hasLowerBounds() || this->hasUpperBounds()) {
+            // The smallest lower and the largest upper bound enclose all values, with a dash for a side that is not known.
+            out << " [";
+            if (this->hasLowerBounds()) {
+                print(out, storm::utility::minmax(this->getLowerBoundVector()).first);
```
volkm: Why not use `minimum` here and `maximum` later on?

src/storm/solver/SolutionBounds.h
volkm, line 1: Maybe also add a helper function here (or in ExplicitQuantitativeCheckResults) which asserts that lower $\leq$ values $\leq$ upper. This could be a helpful sanity check in some places.

src/test/storm/modelchecker/bounds/SolutionBoundsTest.cpp
```diff
@@ -0,0 +1,190 @@
...
+        GTEST_SKIP() << "Z3 not available.";
+#endif
+    }
+
+    storm::Environment const& env() const {
+        return _environment;
+    }
+
+    ExtendedValueType parseNumber(std::string const& input) const {
+        return storm::utility::convertNumber<ExtendedValueType>(storm::utility::convertNumber<ValueType>(input));
```
volkm: Is there no way to directly go to `ExtendedValueType`?

src/test/storm/modelchecker/bounds/SolutionBoundsTest.cpp
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    /*!
+     * How far outside the reported enclosure the exact answer may still lie. A bound is only ever as sound as the
+     * arithmetic it is computed in: a procedure working in doubles can land a bound an ulp or two on the wrong
+     * side of an answer that is not representable at all. That is the rounding we accept, not a gap in the
+     * reasoning, so the slack is far below the solver precision these configurations ask for.
+     */
```
volkm: Try to make these "Claude mannerism" a bit more concise

src/test/storm/modelchecker/bounds/SolutionBoundsTest.cpp
```diff
@@ -0,0 +1,190 @@
...
+        program = program.preprocess();
+        auto formulas = storm::api::extractFormulasFromProperties(storm::api::parsePropertiesForPrismProgram(propertyString, program));
+        auto model = storm::api::buildSparseModel<ValueType>(program, formulas)->template as<ModelType>();
+        storm::modelchecker::CheckTask<storm::logic::Formula, ValueType> task(*formulas.front(), true);
+        if constexpr (std::is_same_v<ModelType, storm::models::sparse::Dtmc<ValueType>>) {
+            return storm::modelchecker::SparseDtmcPrctlModelChecker<ModelType>(*model).check(this->env(), task);
+        } else {
+            return storm::modelchecker::SparseMdpPrctlModelChecker<ModelType>(*model).check(this->env(), task);
+        }
+    }
```
volkm: Should these be merged?

src/test/storm/modelchecker/bounds/SolutionBoundsTest.cpp
volkm, line 1: As far as I see, the lower/upper bounds should become part of the regular model checking. I suggest to extend the existing tests to incorporate and check these bounds there (e.g. in `test/storm/modelchecker/prctl`).

src/storm/modelchecker/prctl/helper/SparseDtmcPrctlHelper.cpp
```diff
@@ -350,19 +367,30 @@ std::vector<SolutionType> SparseDtmcPrctlHelper<ValueType, RewardModelType, Solu
...
+        auto& bounds = returnValue.solutionBounds;
+        for (auto* bound : {&bounds.lower, &bounds.upper}) {
+            if (*bound) {
+                for (auto& entry : **bound) {
+                    entry = storm::utility::one<SolutionType>() - entry;
+                }
+            }
+        }
+        std::swap(bounds.lower, bounds.upper);
+        return returnValue;
```
volkm: There is now also `ExplicitQuantitativeCheckResult<ValueType>::oneMinus()` but I guess this cannot be easily used?

src/storm/modelchecker/prctl/SparseDtmcPrctlModelChecker.cpp
```diff
@@ -98,7 +98,7 @@ std::unique_ptr<CheckResult> SparseDtmcPrctlModelChecker<SparseDtmcModelType>::c
         auto formula = std::make_shared<storm::logic::ProbabilityOperatorFormula>(checkTask.getFormula().asSharedPointer(), opInfo);
         auto numericResult = storm::modelchecker::helper::SparseDtmcPrctlHelper<ValueType, RewardModelType, SolutionType>::computeRewardBoundedValues(
             env, this->getModel(), formula);
-        return std::unique_ptr<CheckResult>(new ExplicitQuantitativeCheckResult<SolutionType>(std::move(numericResult)));
+        return std::unique_ptr<CheckResult>(new ExplicitQuantitativeCheckResult<SolutionType>(this->getModel().getInitialStates(), std::move(numericResult)));
```
volkm, line 102: I see that `computeRewardBoundedValues` only returns values for the initial states. We should (at some point) maybe make this explicit, similar to the SparseMdpPrctlHelper

volkm (commented):
Overall, the changes look good to me. Just a number of (mostly smaller) comments.

I think the main decision is whether we need to distinguish between results for all states and for some states, or whether we just always use `states` to capture this. As `states` is a BitVector, the overhead should not be too much.

src/test/storm/modelchecker/prctl/dtmc/DtmcPrctlModelCheckerTest.cpp
```diff
@@ -640,6 +641,27 @@ class DtmcPrctlModelCheckerTest : public ::testing::Test {
...
+        }
+        // The result is filtered to the initial states at this point, so its bounds are those of the initial states.
+        auto const& explicitResult = result->template asExplicitQuantitativeCheckResult<ValueType>();
+        ValueType const tolerance = TestType::isExact ? this->parseNumber("0") : this->parseNumber("1e-12");
+        if (explicitResult.hasLowerBounds()) {
+            EXPECT_LE(storm::utility::minimum(explicitResult.getLowerBoundVector()), expected + tolerance) << "The lower bound exceeds the expected value.";
+        }
+        if (explicitResult.hasUpperBounds()) {
+            EXPECT_LE(expected, storm::utility::maximum(explicitResult.getUpperBoundVector()) + tolerance) << "The upper bound falls below the expected value.";
+        }
```
volkm, line 662: Is the result computed for a single state or for all states? In the case for all states, these bounds are not tight enough: it should consider the bounds of exactly the initial state (and not the extreme values over all states).

lukovdm: After getQuantitativeResultAtInitialState, the check result is filtered to only have the initial state values. This also applies to the bounds. Thus the minimum and maximum are both applied only to the initial states.

volkm: Maybe use `explicitResult.getLowerBoundVector().front()` to make this clear?

.gitignore
```diff
@@ -47,3 +47,4 @@ nbproject/
 .idea
 .vscode
 *.out
+.claude/
```
tquatmann: This went in here by accident I assume (I don't mind this change, though).

src/storm/solver/AbstractEquationSolver.h
```diff
@@ -263,6 +290,9 @@ class AbstractEquationSolver {
     boost::optional<std::vector<ValueType>> upperBounds;
 
    private:
+    // Sound bounds on the solution, if the last call to this solver produced any.
+    mutable SolutionBounds<ValueType> solutionBounds;
```
tquatmann, line 301: Is it intentional that these are not declared using a separate SolutionType template parameter? Like e.g. MinMaxLinearEquationSolver uses? Same for the pre-existing lowerBound, upperBound.

src/storm/solver/IterativeMinMaxLinearEquationSolver.cpp
tquatmann, line 573: Also set solution bounds here (lower=upper=0)

src/storm/modelchecker/prctl/helper/SparseMdpPrctlHelper.cpp
```diff
@@ -819,6 +856,16 @@ MDPSparseModelCheckingHelperReturnType<SolutionType> SparseMdpPrctlHelper<ValueT
...
+        // Inverse the bounds, lb = 1 - ub and ub = 1 - lb.
+        auto& bounds = result.solutionBounds;
+        for (auto* bound : {&bounds.lower, &bounds.upper}) {
+            if (*bound) {
+                for (auto& entry : **bound) {
+                    entry = storm::utility::one<SolutionType>() - entry;
+                }
+            }
+        }
+        std::swap(bounds.lower, bounds.upper);
```
tquatmann: Maybe add `invertProbabilityBounds()` or something to the solutionBounds class and use it here. This also appears in the DTMC case.
(I personally prefer doing this with std::ranges::transform or something, `**bound` looks a bit scary to me 👻 )

src/storm/modelchecker/results/ExplicitQualitativeCheckResult.cpp
tquatmann: truthValues is a vector of size `1`, initialized with `value`, so this is correct (but slightly weird to read, I agree)

src/storm/modelchecker/results/ExplicitQuantitativeCheckResult.cpp
```diff
@@ -320,54 +326,62 @@ void printRange(std::ostream& out, ValueType const& min, ValueType const& max) {
...
 }
 
+template<typename ValueType>
+void ExplicitQuantitativeCheckResult<ValueType>::printValue(std::ostream& out, uint64_t offset) const {
+    print(out, values[offset]);
+    if (!this->hasLowerBounds() && !this->hasUpperBounds()) {
+        return;
+    }
+    out << " [";
+    // A side that is not known is written as a dash, so that it cannot be read as an infinite bound.
```
tquatmann, line 351: Could we rather use `?` ? I think this makes the "unknown" semantics more clear  (compare `[0.5,-]` vs `[0.5,?]`)

It also makes parsing these bounds easier: The dash in `[-,0]` must be distinguished from the minus in  `[-1,0]`.

src/storm/modelchecker/results/ExplicitQuantitativeCheckResult.cpp
```diff
@@ -242,23 +259,12 @@ typename ExplicitQuantitativeCheckResult<ValueType>::ExtendedValueType ExplicitQ
...
-        count = boost::get<vector_type>(values).size();
-    } else {
-        for (auto const& element : boost::get<map_type>(values)) {
-            STORM_LOG_THROW(!storm::utility::isInfinity(element.second), storm::exceptions::InvalidOperationException,
-                            "Cannot compute the average of values containing infinity.");
-            sum += element.second;
-        }
-        count = boost::get<map_type>(values).size();
+    for (auto const& element : values) {
+        STORM_LOG_THROW(!storm::utility::isInfinity(element), storm::exceptions::InvalidOperationException,
```
tquatmann: imo, the average is well defined as long as we don't have both, plus and minus infinity in there?
This also does not check for -infinity, right?

src/storm/modelchecker/results/ExplicitQualitativeCheckResult.cpp
volkm: True, I misread it.

src/storm/solver/AbstractEquationSolver.h
lukovdm: I am now also confused by these, but the `MinMaxLinearEquationSolver` and any other equation solver that has `SolutionTypes` all inherit from `AbstractEquationSolver<SolutionType>` and not `ValueType`. Thus these should all be seen as solutionTypes.

tquatmann: Ahh, I see, that makes sense.
Maybe the template parameter in AbstractEquationSolver should then be called `SolutionType` and not `ValueType`, right?

lukovdm: That makes sense yes.

PR #1034, opened 2026-08-23
Split storm-dft.h into analysis, io and transformation

volkm:
- Split `storm-dft.h` into `analysis.h`, `io.h` and `transformation.h`
- Added forward declarations in DFT parsers
- cleaned up includes in DFT tests


src/storm-dft/api/analysis.cpp
```diff
@@ -1,26 +1,46 @@
-#include "storm-dft/api/storm-dft.h"
+#include "analysis.h"
```
sjunges, line 1: wrong naming scheme? Or is this the new naming scheme?

src/storm-dft/api/transformation.h
```diff
@@ -0,0 +1,72 @@
...
+template<typename ValueType>
+std::shared_ptr<storm::dft::storage::DFT<ValueType>> prepareForMarkovAnalysis(storm::dft::storage::DFT<ValueType> const& dft);
+
+/*!
+ * Transform DFT to GSPN.
+ *
+ * @param dft DFT.
+ * @return Pair of GSPN and id of failed place corresponding to the top level element.
+ */
+template<typename ValueType>
```
sjunges: The transform to GSPN/JANI feel a bit like the odd-ones-out. They are more like conversion compare do the other questions?

sjunges: (and they drag in a lot of headers)

src/test/storm-dft/api/DftModelBuildingTest.cpp
```diff
@@ -1,9 +1,13 @@
...
 
-#include "storm-dft/api/storm-dft.h"
+#include "storm-dft/api/io.h"
+#include "storm-dft/api/transformation.h"
 #include "storm-dft/builder/ExplicitDFTModelBuilder.h"
-#include "storm-parsers/api/storm-parsers.h"
+#include "storm-dft/utility/RelevantEvents.h"
+#include "storm-parsers/api/properties.h"
+#include "storm/api/properties.h"
+#include "storm/storage/jani/Property.h"
```
sjunges, line 10: It is a bit ugly that you need to invoke three proper-related headers :-)

src/storm-dft/api/transformation.h
volkm: I could split it into translation and transformation. Maybe call it gspn_translation to avoid confusion.

src/test/storm-dft/api/DftModelBuildingTest.cpp
volkm: Yes, but this is a separate issue in my view. In particular the inclusion of jani/Property is very ugly because it is only forward declared otherwise.

src/storm-dft/api/analysis.cpp
volkm: To be honest, I thought this was the new naming scheme. But I am not sure whether we agreed on it.

PR #1035, opened 2026-08-25
Remove uses of infinity that do not need a number

lukovdm:
Several places used `storm::utility::infinity<T>()`, which is the literal 100000000000 for the exact value types, as a stand-in for something structural rather than for a value. These locations can use combinations of optionals or finite upper bounds on the value.

This is part of a several PRs that aim to solve the rational infinity problem. The next PR will introduce a wrapper around `ValueTypes` that adds infinity and which can be used when solutions like in this PR do not work.

### DIFF
--- a/src/storm-dft/builder/DftExplorationHeuristic.h
+++ b/src/storm-dft/builder/DftExplorationHeuristic.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <optional>
+
 #include <memory>
 
 #include "storm/exceptions/NotImplementedException.h"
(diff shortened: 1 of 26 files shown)


src/storm-dft/builder/DftExplorationHeuristic.h
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <optional>
+
 #include <memory>
```
volkm:
```suggestion
#include <memory>
#include <optional>

```

PR #1037, opened 2026-08-25
Replaced settings by environments in storm-dft

volkm:
Builds on https://github.com/stormchecker/storm/pull/1036 and https://github.com/stormchecker/storm/pull/1034 which should be merged first.

Use DFT environment in storm-dft now instead of settings. The settings are completely moved into storm-dft-cli now, making it truly independent of the settings.
We still need to initialize the settings in the `storm-dft` library, because the Storm core might still need them. This could be moved in the future as well.
Note that due to adding the DFT settings modules later, `storm-dft --help` now shows all DFT related settings last.

- Also used a callback for export options of the Markov model (getting rid of the IO settings)
- used `STORM_LOG_STATISTICS_LAZY` for statistics.


src/storm-dft-cli/CMakeLists.txt
```diff
@@ -1,5 +1,10 @@
 # Create storm-dft.
-add_executable(storm-dft-cli ${PROJECT_SOURCE_DIR}/src/storm-dft-cli/storm-dft.cpp)
+add_executable(storm-dft-cli
+    ${PROJECT_SOURCE_DIR}/src/storm-dft-cli/storm-dft.cpp
```
sjunges, line 3: This breaks our common glob-pattern.

volkm: As far as I can see, we only use the glob pattern in the libs, not in the cli. For instance, storm-pars has [the same pattern](https://github.com/stormchecker/storm/blob/master/src/storm-pars-cli/CMakeLists.txt).

PR #1042, opened 2026-09-03
Reduced direct carl calls

volkm:
The goal is to reduce the direct calls to `carl` in the code and instead go through either an adapter or `constants.h`.

Some notable changes:
- Instantiate via `storm::RationalNumber` and `storm::RationalFunctionCoefficient` instead of `Cln/GmpRationalNumber`, because the type of value (exact or from rational function) is more relevant than the underlying library. Introduced define `STORM_RATIONAL_NUMBER_DIFFERS_FROM_COEFFICIENT` to disable instantiations if both number types coincide.
- Replaced `carl:hash_add` by `boost::hash_combine`. Carl still uses an [older version](https://github.com/stormchecker/carl-storm/blob/master/src/carl/util/hash.h) of the hash which has [issues](https://stackoverflow.com/a/50978188).
- Added `gcd` and `lcm` for Integers and revised `vector::toIntegralVector`. The function now performs the operations on integers instead of rational representations of integers.

### DIFF
--- a/resources/3rdparty/sylvan/src/storm_wrapper.cpp
+++ b/resources/3rdparty/sylvan/src/storm_wrapper.cpp
@@ -212,7 +212,7 @@ storm_rational_number_ptr storm_rational_number_pow(storm_rational_number_ptr a,
     storm::RationalNumber const& srn_a = *static_cast<storm::RationalNumber const*>(a);
     storm::RationalNumber const& srn_b = *static_cast<storm::RationalNumber const*>(b);
 
-    int_fast64_t exponentAsInteger = storm::utility::convertNumber<int_fast64_t>(srn_b);
+    int64_t exponentAsInteger = storm::utility::convertNumber<int64_t>(srn_b);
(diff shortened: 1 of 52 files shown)


resources/3rdparty/sylvan/src/storm_wrapper.cpp
```diff
@@ -212,7 +212,7 @@ storm_rational_number_ptr storm_rational_number_pow(storm_rational_number_ptr a,
     storm::RationalNumber const& srn_a = *static_cast<storm::RationalNumber const*>(a);
     storm::RationalNumber const& srn_b = *static_cast<storm::RationalNumber const*>(b);
 
-    carl::sint exponentAsInteger = carl::toInt<carl::sint>(srn_b);
+    int_fast64_t exponentAsInteger = storm::utility::convertNumber<int_fast64_t>(srn_b);
```
sjunges: I dont think we use the fast64_t anymore? :-)

volkm: I was not sure what our policy is. But I found [this issue](https://github.com/stormchecker/storm/issues/259) that we indeed do not use fast64. Will change it.

Issue #1033, opened 2026-08-21
Restructure storage directory


volkm:
The `storage` directory contains a large number of classes and should be restructured.
- We should introduce a new folder `languages` which contains language-specific data structures such as `storage/jani` and `storage/prism`. Folder `storage/umb` might also be a candidate.
- `storage/bisimulation` should be moved to `transformer`
- `Qvbs` could maybe be moved to `io`?

PR #1049, opened 2026-09-11
New Sparse Bisimulation

### LINKED ISSUE #833: Differences in CTMC bisimulation quotient
Applying bisimulation minimization on the CTMC `resources/examples/testfiles/ctmc/embedded2.sm` yields differently sized quotients, see below.
I can understand that floating point bisimulation can yield different quotient due to precision issues, but I would have expected that using exact arithmetic does not make a difference.

- Encountered when adding a CTMC test case in https://github.com/moves-rwth/storm/pull/826
- Could be related to https://github.com/moves-rwth/storm/issues/91

## Original model
```
Model type: 	CTMC (sparse)
States: 		3478
Transitions: 	14639
```

## Exact without labels
`storm --prism embedded2.sm --bisimulation:type strong -bisim -pc -e sparse --exact`

On Linux Docker image (stable and ci):
```
Model type: 	CTMC (sparse)
States: 		180
Transitions: 	1195
```

On Mac:
```
Model type: 	CTMC (sparse)
States: 		136
Transitions: 	793
```

## Exact with labels
`storm --prism embedded2.sm --bisimulation:type strong -bisim -pc -e sparse --exact  --build-all-labels`

Same results on both Linux Docker image and Mac:
```
Model type: 	CTMC (sparse)
States: 		1127
Transitions: 	5730
```

## Double without labels
`storm --prism embedded2.sm --bisimulation:type strong -bisim -pc -e sparse`

On Linux Docker image (stable and ci):
```
Model type: 	CTMC (sparse)
States: 	 	1734
Transitions: 	9355
```

On Mac:
```
Model type: 	CTMC (sparse)
States: 	 	1318
Transitions: 	7276
```

## Double with labels
`storm --prism embedded2.sm --bisimulation:type strong -bisim -pc -e sparse --build-all-labels`

On Linux Docker image (stable and ci):
```
Model type: 	CTMC (sparse)
States: 		1720
Transitions: 	8623
```

On Mac:
```
Model type: 	CTMC (sparse)
States: 		1731
Transitions: 	8619
```

## DD
The Dd engine yields an assertion error:
```
ERROR (QuotientExtractor.cpp:1178): Illegal non-probabilistic matrix.
```
which is related to https://github.com/moves-rwth/storm/discussions/822

tquatmann:
Completely revised sparse bisimulation implementation.

## What's new

- Usually faster compared to the current implementation. In particular for MDPs.
- Now supports Markov automata, multiple reward models, preservation of choice labels.
- Returns mappings between quotient and original model states/choices. Useful, e.g. to translate a scheduler for the quotient model to a scheduler for the original model.
- More explicit and robust handling of the approximate case (`tolerance > 0`), in particular for MDPs where choices might be "equivalent up to a tolerance".
- (hopefully) no more undefined behaviour with positive tolerances: previously `std::sort` was called on a non-total order induced by the ConstantsComparator. There is still [an issue](https://github.com/stormchecker/carl-storm/issues/147) with RationalFunction which imo should be handled within carl.
- Moved to transformer (see also #1033).
- fix #833.

## Where to start

`performBisimulationMinimization` in `Bisimulation.cpp` serves as an entry point and goes through the three min steps of bisimulation minimization:

1. Compute an initial partition based on the properties to preserve and on provided options.
2. Apply partition refinement (Splitter-based for DTMC/CTMC or Signature-based for DTMC/CTMC/MDP/MA)
3. Build the quotient from the partition.

## Caveats

I deliberately did not port the "measure driven initial partitions", which lump together states with equal value (e.g. prob0/prob1 states for a `P=? [ F "target" ]` query). While that could be useful in certain cases, I found that it made the entire procedure significantly more complicated—conceptually (qualitative model checking would be a subtask of bisimulation, which is a preprocessing for model checking...) and code-wise. 
The functionality has been moved into [mergeEquivalentStatesForFormula](https://github.com/stormchecker/storm/pull/1016) which can be combined with bisimulation. Hence the combination`--merge-equivalent-states --bisimulation` should be equivalent to the old `--bisimulation`, except that the new one has to create an intermediate model.  #91 is relevant.


src/storm/api/bisimulation.h
```diff
@@ -3,116 +3,36 @@
...
-    } else if (model->isOfType(storm::models::ModelType::Ctmc)) {
-        return performDeterministicSparseBisimulationMinimization<storm::models::sparse::Ctmc<ValueType>>(
-            model->template as<storm::models::sparse::Ctmc<ValueType>>(), formulas, type, graphPreserving, tolerance);
-    } else {
-        return performNondeterministicSparseBisimulationMinimization<storm::models::sparse::Mdp<ValueType>>(
-            model->template as<storm::models::sparse::Mdp<ValueType>>(), formulas, type, graphPreserving, tolerance);
-    }
+    std::shared_ptr<storm::models::sparse::Model<ValueType>> const& model, std::vector<std::shared_ptr<storm::logic::Formula const>> const& formulas = {},
+    storm::bisimulation::Options const& options = {}) {
+    return storm::bisimulation::performBisimulationMinimization<ValueType>(*model, formulas, options).quotient;
```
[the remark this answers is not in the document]
tquatmann: Such a reduction does not preserve all formulas. Rejecting models with transition rewards is intentional.

src/storm/logic/TraverseFormulaVisitor.h
```diff
@@ -0,0 +1,56 @@
+#pragma once
+
+#include <functional>
+
+#include "storm/logic/FormulaVisitor.h"
+
+namespace storm::logic {
+
+/*!
+ * Visits a formula and its subformulas in pre-order, i.e., each formula is visited before its subformulas.
```
sjunges, line 10: Here it says preorder

src/storm/logic/Formula.h
```diff
@@ -111,6 +111,12 @@ class Formula : public std::enable_shared_from_this<Formula> {
     boost::any accept(FormulaVisitor const& visitor) const;
     virtual boost::any accept(FormulaVisitor const& visitor, boost::any const& data) const = 0;
 
+    /*!
+     * Invokes the given callback on this formula and on its subformulas, where a formula is visited before its subformulas.
```
sjunges: ah, here it is also preorder. Read this wrong on the first occurence. Maybe write preoder here as well.

src/storm/settings/modules/BisimulationSettings.cpp
```diff
@@ -91,6 +93,22 @@ BisimulationSettings::BisimulationSettings() : ModuleSettings(moduleName) {
                                          .setDefaultValueString("full")
                                          .build())
                         .build());
+
+    this->addOption(storm::settings::OptionBuilder(moduleName, toleranceOptionName, true,
```
sjunges, line 97: is this now also used for dd-based bisim?

src/storm/storage/sparse/ChoiceOrigins.cpp
```diff
@@ -49,6 +49,10 @@ uint_fast64_t ChoiceOrigins::getIdentifier(uint_fast64_t choiceIndex) const {
     return indexToIdentifier[choiceIndex];
 }
 
+std::vector<uint_fast64_t> const& ChoiceOrigins::getIdentifiers() const {
```
sjunges, line 52: dont use fast64_t

sjunges: i see how it is consistent :-)

src/storm/logic/TraverseFormulaVisitor.h
```diff
@@ -0,0 +1,56 @@
+#pragma once
+
+#include <functional>
+
+#include "storm/logic/FormulaVisitor.h"
+
+namespace storm::logic {
```
sjunges, line 7: Should we open an issue to transform other usages of this pattern to a traverser? (I can see that that wouldnt need to be part of this PR, but it is something we should consider doing to make sure that we actually use this method where it fits)...

src/storm-cli-utilities/model-handling.h
```diff
@@ -705,14 +709,19 @@ template<typename ValueType>
...
     }
-    std::optional<double> tolerance = storm::settings::getModule<storm::settings::modules::GeneralSettings>().getPrecision();
-
-    STORM_LOG_INFO("Performing bisimulation minimization...");
-    return storm::api::performBisimulationMinimization<ValueType>(model, createFormulasToRespect(input.properties), bisimType, graphPreserving, tolerance);
+    options.actionSensitive = bisimulationSettings.isActionSensitiveSet();
+    STORM_LOG_INFO("Performing bisimulation minimization (type: "
+                   << (options.bisimulationType == storm::bisimulation::BisimulationType::Weak ? "weak" : "strong") << ", tolerance: " << options.tolerance
+                   << (options.actionSensitive ? ", action-sensitive" : "") << ")...");
+    return storm::api::performBisimulationMinimization<ValueType>(model, createFormulasToRespect(input.properties), options);
```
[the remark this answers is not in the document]
tquatmann: Given that the measureDriven initial partitions are gone, the graphPreserving setting is no longer relevant. I will drop it from the CLI code.

src/storm/storage/sparse/ChoiceOrigins.cpp
tquatmann: Yes, this was intentional because we take that vector by reference.
I'd rather keep it this way and globally replace all remaining `fast64_t` at a convenient time.

src/storm/settings/modules/BisimulationSettings.cpp
tquatmann: The tolerance is currently sparse only. The help message already mentions this. Should I make this more clear somehow?

dd-bisim does not use any tolerance-based comparison right now.
This example yields some evidence: sparse-bisim correctly yields a quotient with 4 states while dd-bisim yields a quotient with 5 states.

```
dtmc
module M
  s : [0..5] init 0;
  [] s=0 -> 0.5: (s'=1) + 0.5: (s'=2);
  [] s=1 -> 0.1: (s'=3) + 0.2: (s'=4) + 0.7: (s'=5);
  [] s=2 -> 0.3: (s'=3) + 0.7: (s'=5);
  [] s>2 -> true;
endmodule
label "target" = s=3 | s=4;
```

Note that (s=3 and s=4) and (s=1 and s=2) are bisimular, but s=1 and s=2 are split because `0.1+0.2==0.3` doesn't hold in double arithmethic. sparse-bisim with `--bisimulation:tolerance 0` has the same problem.

src/storm/logic/TraverseFormulaVisitor.h
tquatmann: Yes! #1052

src/storm-cli-utilities/model-handling.h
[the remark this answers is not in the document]
tquatmann: Given that the measureDriven initial partitions are gone, the graphPreserving setting is no longer relevant. I will drop it from the CLI code.

src/storm/transformer/bisimulation/Signatures.cpp
```diff
@@ -0,0 +1,461 @@
...
+        ++it2;
+    }
+    return false;
+}
+
+template<typename ValueType, SignatureMode Mode>
+auto Signatures<ValueType, Mode>::getApproximateSplitCondition() const -> SplitCondition
+    requires(Mode == SignatureMode::Approximative)
+{
+    return SplitCondition(stateSignatureCache, halfTolerance);
```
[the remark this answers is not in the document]
tquatmann: `halfTolerance` is used in two places: canonicalizing the choices within one state, and the cross-state `SplitCondition` (here). Those two half-steps chain, because `splitBlockByOrder` compares every element against the anchor of the current sub-block — the first element, which is also the representative — not against its predecessor. So an original choice value is at most `halfTolerance + halfTolerance = tolerance` away from the value that ends up in the quotient, which is exactly what the `halfTolerance + halfTolerance` in `extendQuotientData` relies on. That is sound as it stands.

I did reword the documentation of the `tolerance` option in bisimulation::Options. It now stresses that every original value deviates at most the tolerance from its quotient counterpart.

src/storm/transformer/bisimulation/Quotient.cpp
```diff
@@ -0,0 +1,217 @@
...
+            for (; quotientChoice < quotientChoiceEnd; ++quotientChoice) {
+                for (auto const& [column, value] : getQuotientRow(quotientChoice)) {
+                    builder.addNextValue(quotientChoice, column, value);
+                }
+            }
+        }
+        components.transitionMatrix = builder.build();
+    }
+
+    // build state labeling
```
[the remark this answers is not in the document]
tquatmann: I added state valuations below (line 142, after processing the stateLabeling)

PR #1055, opened 2026-09-14
Unblock WeakUntil/Release in JaniParser

vivoeiva:
It is mostly mechanical addition to `JaniParser` to pass W / R operators and build `WeakUntilFormula` / `ReleaseFormula` without adding any logic behind it. These operators are also disabled by default and should not change STORM behavior.

I understand that this PR has little value for STORM itself, I have added these to cleanly use `JaniParser` in a STORM extension. Happy to hear your feedback on that.


src/storm/logic/ToPrefixStringVisitor.cpp
```diff
@@ -183,6 +183,18 @@ boost::any ToPrefixStringVisitor::visit(UntilFormula const& f, boost::any const&
...
+boost::any ToPrefixStringVisitor::visit(WeakUntilFormula const& f, boost::any const& data) const {
+    std::string left = boost::any_cast<std::string>(f.getLeftSubformula().accept(*this, data));
+    std::string right = boost::any_cast<std::string>(f.getRightSubformula().accept(*this, data));
+    return std::string("W ") + left + " " + right;
+}
+
+boost::any ToPrefixStringVisitor::visit(ReleaseFormula const& f, boost::any const& data) const {
+    std::string left = boost::any_cast<std::string>(f.getLeftSubformula().accept(*this, data));
+    std::string right = boost::any_cast<std::string>(f.getRightSubformula().accept(*this, data));
+    return std::string("R ") + left + " " + right;
```
tquatmann:
```suggestion
    // Write as weak release (V), see https://spot.lre.epita.fr/ioltl.html
    return std::string("V ") + left + " " + right;
```

tquatmann (commented):
Thanks! This is looking quite good already.

Before we merge, please also specialise the `ExtractMaximalStateFormulasVisitor` which inherits from CloneVisitor.

tquatmann (approved):
LGTM!

PR #1056, opened 2026-09-14
Fixed compiler warning on Debian 12

volkm:
Debug mode on Debian 12 throws compiler warnings of the form `error: control reaches end of non-void function [-Werror=return-type]` because it is still excepting a return after `STORM_LOG_THROW(false, ...);`.
This PR silences the warning by using `STORM_LOG_THROW_UNCONDITIONALLY(exception, msg)`.


vivoeiva:
I am sorry to mix in, but maybe it is a use case for `__builtin_unreachable()` to silence the warning? Not sure if it is supported by all targeted compilers though.

volkm:
Thanks for the suggestion. `__builtin_unreachable()` could also be a good option and makes it explicit that this part should not matter.

tquatmann:
Maybe a macro `STORM_LOG_UNREACHABLE(msg)` or `STORM_LOG_THROW_UNCONDITIONALLY(exception, msg)` that unconditionally throws (which might silence the warning already and if not, the macro can set `__buildin_unreachable()` so that we don't have builtin functions all over the place).
There is also [std::unreachable()](https://en.cppreference.com/cpp/utility/unreachable) but that's c++23.

volkm:
CI still runs through for [Debian 12 in Debug mode](https://github.com/volkm/storm/actions/runs/35790486513/job/106957507637).

sjunges:
This solution looks good to me.

volkm:
Thanks, I will merge it.

By the way, there are >1330 other places with `STORM_LOG_THROW(false, ..)` but I (Claude) only changed the ones which required it.

PR #1060, opened 2026-09-15
Workaround for libatomic in Spot

volkm:
The stormpy wheels have recently started failing, because importing stormpy cannot find libatomic, see e.g. [this run](https://github.com/stormchecker/stormpy-wheels/actions/runs/34939660199/job/104304225430).
This library is excluded by auditwheel when repairing the wheel, since `libatomic.so.1` is treated as part of the manylinux baseline system and therefore not bundled. However, not all systems ship with libatomic. The Debian/Ubuntu default Docker images for example do not have it and the tests therefore fail.

The issue appeared when starting to use Spot 2.16. While Spot does not have libatomic as a new dependency, they fixed [the inclusion](https://gitlab.lre.epita.fr/spot/spot/-/commit/df768a378b883a576f1b5c40b687ca8b4182d706) such that it now is used if the default CMPXCHG16B is missing. As we are using `STORM_PORTABLE=ON`, this is missing and our Spot build falls back to libatomic.

Long story short: we now explicitly include libatomic as a static library and the stormpy wheels are successful again (tested on my fork).


volkm:
The picture has slightly changed. Since 2 days ago, Ubuntu and Debian Docker images now seem to ship with libatomic. Only the Fedora image is now failing for the wheel CI. Under these circumstances we could also argue that libatomic is assumed to be present (auditwheel at least has this assumption). Then we would not need this PR and could just manually install libatomic in the Fedora image.

sjunges:
I would indeed say that fedora should ship with libatomic and for now install it in our image. If this does not resolve on fedora etc as well over the next weeks, we can reconsider.

volkm:
Thanks, I think this indeed the simplest solution. I will update the image and then close this PR for now.

PR #1062, opened 2026-09-17
Fix boolean over path ops in JaniParser

vivoeiva:
Currently JaniParser emits `BooleanStateFormula` whenever it reaches a boolean operator over some other operator(s). This change it to actual check if one of the op's underneath is a path operator and emits `BooleanPathFormula `/ `BooleanStateFormula` depending on that check.

Tests expanded to check for new behavior. The embedded `die.jani` also passes through the actual MC and emits plausible results.


volkm (approved):
The code looks good to me.

@tquatmann Are the any conceptual issues with this change I might have overlooked?

tquatmann:
LGTM!

PR #1048, opened 2026-09-10
Upper and lower bounds for more algorithms

lukovdm:
This is subsumed by #1066.

PR #1066, opened 2026-09-22
Sound Solution bounds for DTMC/MDP model checking methods

lukovdm:
Subsumes #1048

Builds on top of #1031.

This PR adds bounds for the following algorithms:

- Value iteration (power method) — solver/NativeLinearEquationSolver.cpp (solveEquationsPower), solver/helper/ValueIterationHelper.cpp. A sweep that moved every entry one way proves the operand is a post- or pre-fixpoint, giving that side of the enclosure. Only for in-place (Gauss-Seidel) multiplication.
- Value iteration (MinMax) — solver/IterativeMinMaxLinearEquationSolver.cpp (solveEquationsValueIteration). The same sweep certificate, claimed only when the caller has set hasUniqueSolution().
- Sound value iteration — solver/helper/SoundValueIterationHelper.cpp, plus solveEquationsSoundValueIteration in both solvers. The enclosure SVI maintains, read out before the final averaging step collapses it. Needs both scaling factors to be known.
- Guessing value iteration — solveEquationsGuessingValueIteration in both solvers. The two vectors GVI keeps, which hold verified guesses and so enclose the solution throughout.
- Rational search — solveEquationsRationalSearch in both solvers. The solution as both bounds, once sharpening has verified an exact fixpoint.
- Policy iteration — solver/IterativeMinMaxLinearEquationSolver.cpp (performPolicyIteration). Whatever the inner linear solver established. Optimality of the scheduler says nothing about the accuracy of the values.
- State elimination — solver/EliminationLinearEquationSolver.cpp. The solution as both bounds; the method is direct.
- Eigen SparseLU — solver/EigenLinearEquationSolver.cpp. The solution as both bounds. The iterative Eigen methods are left out.
- Acyclic solving — solver/AcyclicLinearEquationSolver.cpp, solver/AcyclicMinMaxLinearEquationSolver.cpp. The solution as both bounds; values are substituted in topological order and final when written.
- Topological solving — solver/TopologicalLinearEquationSolver.cpp, solver/TopologicalMinMaxLinearEquationSolver.cpp. An interval of ±precision around the result, under --sound only. The per-SCC deviations add to at most the configured precision along any chain.
- Robust value iteration (interval models) — modelchecker/prctl/helper/SparseDtmcPrctlHelper.cpp (computeRobustValuesForMaybeStates). Whatever the MinMax solver produced. Now that is the sweep certificate, so rewards report and probabilities do not.

This is the ones I have tackled so far, I am not entirely certain these bounds are correct, but they all seem logical to me and when possible I tried to read the papers to find if they claim a bound. Please suggest other algorithms I can report bounds for.


lukovdm:
Are there any algorithms that should be added. If not I will do another pass over the PR and then it should be ready for review.

tquatmann:
> Are there any algorithms that should be added. If not I will do another pass over the PR and then it should be ready for review.

I think this is a good list for now. My thoughts:

- Topological: if all SCCs are singleton, we can set the exact bound (same as acyclic)
- Step-bounded until, cumulative rewards, and Nest probabilities can get exact bounds
- If some preprocessing bounds are provided to the solver, we may take those
- If value iteration is started at a lower bound (like zero), it will always give a lower bound for the solution. Not sure if we can make use of this easily. The "A sweep that moved every entry one way"-approach sounds like it will affect runtime?

There might be some more algorithms, but we can always add them on demand.

PR #1071, opened 2026-09-23
Hide Eigen and exprtk template instantiations inside libstorm

sjunges:
Related to #1070, but now targets included 3rd party libraries.

Ideally, we would flip this around and highlight what is part of the API rather than saying that some things are not, but that is a major change that touches basically everything. For now, this should have a significant positive effect.


volkm (approved):
LGTM.

(I generally dislike Claude's phrasing of comments but it is good to explain the motivation.)

sjunges:
Lol. No claude involved (it was a different agent). Will reformulate those.

volkm:
At least I am glad I did not insult your writing ;)

PR #1073, opened 2026-09-23
Removed settings from storm-gspn

volkm:
Also split up GSPN export into dedicated functions for each export type.

The settings currently still need to be part of storm-gspn because storm-dft-cli depends on them.

These changes also fix an issue with the jani export of a GSPN translated from a DFT. Previously, there were two exports handled (one through storm-gspn and one directly in storm-dft) leading to variables such as `loc` defined twice.


sjunges:
LGTM

